PR #6, opened 2022-02-02
Fix nested parameters in setup_env.sh

sanchda:
Unlike zsh, bash doesn't allow nested parameters.  When `CC` was
undefined, this was causing the vendor directory to be called something
like  `vendor_,,`, which by itself isn't horrible (I built many
instances of ddprof this way on various VMs)--_but_ the build scripts
for some dependencies we _may_ use, like libunwind, glob in a weird way,
which breaks them when a pathname includes commas.

I don't think this will break our CI caches.

# What does this PR do?

Simple change to one of the environment setup scripts.

# Motivation

* Keep the vendor directory name informative
* Prevent hard-to-debug issues with new dependencies

# Additional Notes

If those broke C caches it would be a problem, but I think we're fine.

# How to test the change?

shrug.jpeg


r1viollet (approved):
Thanks for fixing this !
You are correct, build caches will be fine (they don't use this script)

PR #7, opened 2022-02-03
Remove usage of libdwfl internals

r1viollet:
# What does this PR do?

- Remove hacky usage of dwarf module internals. Instead I use the public API to retrieve the address values.
- Remove usage of dwarf frame to retrieve register values.
Although it was handy to keep track of registers, we need to avoid using the internal structures of the libdwfl library.

# Motivation

Avoid exposing the internals of an external library.

# Additional Notes

The work around the dwarf modules is somewhat incomplete. 
The behaviours in whole host make it very likely that we would need to free part of the dwarf modules. I am not seeing an API that would enable me to do so (unless I just free the dwfl object associated to a PID). 

# How to test the change?

This change impacts two behaviours
- Retrieval of dwarf modules
This is best tested in whole host mode. 
- Update of registers 
We should make sure that when we stop unwinding the registers are correctly set and that the debug frames are relevant.


cmake/Findelfutils.cmake
```diff
@@ -17,7 +17,7 @@ set(ELFUTILS_LIBS
     "${LIBDW_PATH}"
     "${LIBELF_PATH}")
 
-list(APPEND ELFUTILS_INCLUDE_LIST ${ELFUTILS_PATH} ${ELFUTILS_PATH}/libdwfl ${ELFUTILS_PATH}/libdw ${ELFUTILS_PATH}/libebl ${ELFUTILS_PATH}/libelf)
+list(APPEND ELFUTILS_INCLUDE_LIST ${ELFUTILS_PATH} ${ELFUTILS_PATH}/libdwfl ${ELFUTILS_PATH}/libdw ${ELFUTILS_PATH}/libebl ${ELFUTILS_PATH}/libelf ${ELFUTILS_PATH}/lib ${ELFUTILS_PATH}/libdwelf)
```
r1viollet, line 20: Irrelevant change here to be removed

include/unwind_state.hpp
```diff
@@ -55,7 +55,7 @@ typedef struct UnwindState {
   size_t stack_sz;
 
   UnwindRegisters initial_regs;
-  UnwindRegisters current_regs;
+  uint64_t current_eip;
```
r1viollet, line 58: This type should be a process address.

src/dwfl_module.cc
```diff
@@ -10,62 +10,75 @@ extern "C" {
...
-    if (vm_addr != dso._start - dso._pgoff) {
+  DDProfMod ddprof_mod;
+
+  ddprof_mod._mod = dwfl_addrmodule(dwfl, pc);
+  Dwarf_Addr low_addr = 0;
+  Dwarf_Addr high_addr = 0;
+
+  if (ddprof_mod._mod) {
+    const char *main_name = nullptr;
+    dwfl_module_info(ddprof_mod._mod, 0, &ddprof_mod._low_addr,
```
sanchda, line 33: Wow, good update.  I wonder what implications there are for populating these fields more fully.  Any thoughts?

src/dwfl_module.cc
```diff
@@ -10,62 +10,75 @@ extern "C" {
...
+        // Although the check is slightly fragile
+        // we are not comparing the full path as we use the proc maps
+        // to access them. So the root path could change while
+        // this would still be the same file.
         const char *dso_name = strrchr(filepath.c_str(), '/') + 1;
-        if (mod && strcmp(dso_name, mod->name) != 0) {
-          LG_NTC("Incoherent DSO (%s) %lx != %lx dwfl_module (%s)", dso_name,
-                 dso._start - dso._pgoff, vm_addr, mod->name);
-          mod = nullptr;
+        const char *mod_name = strrchr(main_name, '/') + 1;
```
sanchda, line 46: Are we assured that `main_name` will always contain a slash?  I'm reasonably confident that `dso_name` will, since both perf_event_open and /proc report absolute paths.

src/dwfl_thread_callbacks.cc
```diff
@@ -19,9 +19,9 @@ bool set_initial_registers(Dwfl_Thread *thread, void *arg) {
 
   // Use the current instead of initial registers to ensure we can call several
```
sanchda, line 20: Might be good to update the comment preceding these changes, since the comment mentions "current" versus "initial" registers.

include/dwfl_module.hpp
```diff
@@ -8,9 +8,17 @@ extern "C" {
 #include "dso.hpp"
 
 namespace ddprof {
+
+struct DDProfMod {
```
sanchda, line 12: Ah, good--thank you.

sanchda (dismissed):
LGTM!  Thank you!

src/dwfl_module.cc
r1viollet: Very good question. I'm testing things out right now.

nsavoire (approved):

src/dwfl_module.cc
r1viollet: I testing things out and could not see it happen. However if we don't guard against it, we will dump as null+1 will be considered as a string by strcmp.

src/dwfl_thread_callbacks.cc
r1viollet: Fixed

nsavoire (approved):

src/dwfl_module.cc
r1viollet: I did not understand this question fully. Perhaps we can come back to this.

src/dwfl_module.cc
sanchda: Right--doesn't hurt to be cautious if we don't have a strong guarantee from whatever gets the path.  Thanks for looking at that!

src/dwfl_module.cc
```diff
@@ -10,6 +10,15 @@ extern "C" {
 }
 
 namespace ddprof {
+
+static const char *get_filename(const char *full_path) {
```
sanchda, line 14: Great!  This won't work for relative paths, but such paths are never returned by the interfaces we use (those interfaces have no concept of "current working directory" since they're both handled by the kernel), so this is good.

sanchda (commented):
Historically:  LGTM!

PR #8, opened 2022-02-08
Free inconsistent dwfl modules

r1viollet:
# What does this PR do?

- Free inconsistent Dwfl modules
We can come into situations where DSOs don't match the Dwfl objects.
In this situation we can free all of the PID objects to ensure we will reload all DSOs and all dwfl objects at the next sample.
- Add a unit test around the creation of dwfl modules
- Whole host mode bug : avoid loading the wrong file
When you have a local libc and a libc in the container, we were giving precedence to the local version of the file. This was a bad strategy that ended up loading inconsistent DSO <--> DWFL Modules

# Motivation

Looking at whole host profiling, we quickly encounter situations where we are not in sync between libdwfl objects and internal DSO objects.

# Additional Notes

There are still some limitations
- We should not compare dwfl and DSO objects using filenames. Perhaps storing file IDs would be more robust. I wanted to avoid storing pointers to the dwfl mod objects internally. Other ideas are welcome.
- When we encounter an error loading a mod, we attribute that to the file. This is not true in some cases. It should be attributed to the PID + file (and we could retry for a different PID). 

# How to test the change?

I added a unit test.
The best way to observe a change is to try out the whole host profiler in an environment profiling different languages.

### DIFF
--- a/include/dwfl_module.hpp
+++ b/include/dwfl_module.hpp
@@ -10,17 +10,21 @@ extern "C" {
 namespace ddprof {
 
 struct DDProfMod {
-  DDProfMod()
-      : _mod(nullptr), _low_addr(0), _high_addr(0), _inconsistent(false) {}
+  enum Status {
+    kUnknown,
+    kInconsistent,
+  };
 
-  explicit DDProfMod(bool inconsistent) : DDProfMod() {
-    _inconsistent = inconsistent;
+  DDProfMod() : _mod(nullptr), _low_addr(0), _high_addr(0), _status(kUnknown) {}
+
+  explicit DDProfMod(Status inconsistent) : DDProfMod() {
+    _status = inconsistent;
   }
 
   Dwfl_Module *_mod;
   ProcessAddress_t _low_addr;
   ProcessAddress_t _high_addr;
-  bool _inconsistent;
+  Status _status;
 };
 
 // From a dso object (and the matching file), attach the module to the dwfl
--- a/include/dwfl_module.hpp
+++ b/include/dwfl_module.hpp
@@ -17,9 +17,7 @@ struct DDProfMod {
 
   DDProfMod() : _mod(nullptr), _low_addr(0), _high_addr(0), _status(kUnknown) {}
 
-  explicit DDProfMod(Status inconsistent) : DDProfMod() {
-    _status = inconsistent;
-  }
+  explicit DDProfMod(Status status) : DDProfMod() { _status = status; }
 
   Dwfl_Module *_mod;
   ProcessAddress_t _low_addr;
--- a/include/dwfl_module.hpp
+++ b/include/dwfl_module.hpp
@@ -10,10 +10,17 @@ extern "C" {
 namespace ddprof {
 
 struct DDProfMod {
-  DDProfMod() : _mod(nullptr), _low_addr(0), _high_addr(0) {}
+  DDProfMod()
+      : _mod(nullptr), _low_addr(0), _high_addr(0), _inconsistent(false) {}
+
+  explicit DDProfMod(bool inconsistent) : DDProfMod() {
+    _inconsistent = inconsistent;
+  }
+
   Dwfl_Module *_mod;
   ProcessAddress_t _low_addr;
   ProcessAddress_t _high_addr;
+  bool _inconsistent;
 };
 
 // From a dso object (and the matching file), attach the module to the dwfl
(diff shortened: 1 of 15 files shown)


include/dwfl_module.hpp
```diff
@@ -10,10 +10,17 @@ extern "C" {
...
+      : _mod(nullptr), _low_addr(0), _high_addr(0), _inconsistent(false) {}
+
+  explicit DDProfMod(bool inconsistent) : DDProfMod() {
+    _inconsistent = inconsistent;
+  }
+
   Dwfl_Module *_mod;
   ProcessAddress_t _low_addr;
   ProcessAddress_t _high_addr;
+  bool _inconsistent;
```
r1viollet, line 23: An enum might be better here.

include/dwfl_module.hpp
```diff
@@ -10,10 +10,21 @@ extern "C" {
 namespace ddprof {
 
 struct DDProfMod {
-  DDProfMod() : _mod(nullptr), _low_addr(0), _high_addr(0) {}
+  enum Status {
```
sanchda, line 13: Haha, I really like this.  Usually a status would have a "Good" type, but since we _cannot_ make such an evaluation (if we could, then because it's all happening in a separate process, our good view of the world can immediately become bad), then the best we can say is we haven't discovered an inconsistency _yet_.

Am I reading the semantics correctly?

include/dwfl_module.hpp
```diff
@@ -10,10 +10,21 @@ extern "C" {
...
+    kUnknown,
+    kInconsistent,
+  };
+
+  DDProfMod() : _mod(nullptr), _low_addr(0), _high_addr(0), _status(kUnknown) {}
+
+  explicit DDProfMod(Status inconsistent) : DDProfMod() {
```
sanchda: Very slight nit.  Even if this method is only called to mark an object as inconsistent in code, there's no such limitation in the implementation itself.  Please feel free to ignore, but proposing
1.  Change the parameter local name to "value" or "status" or anything
2. Change the name of the function to reflect the intent that it only be used to mark objects as inconsistent (and possibly adjust the method logic accordingly)

Again, your discretion--just seemed jarring on first read.

src/dwfl_hdr.cc
```diff
@@ -66,9 +67,14 @@ DDRes DwflWrapper::register_mod(ProcessAddress_t pc, const Dso &dso,
...
   if (!mod_added) {
+    // first time we see this binary for this pid
     DDProfMod ddprof_mod = update_module(_dwfl, pc, dso, fileInfoValue);
     if (!ddprof_mod._mod) {
-      LG_DBG("Unable to register mod %s", dso.to_string().c_str());
+      LG_WRN("Unable to register mod %s - %d", dso.to_string().c_str(),
+             ddprof_mod._status);
+      // If it overlaps with a previous binary we loaded, flag this
+      // inconsistency
+      _inconsistent = ddprof_mod._status == DDProfMod::kInconsistent;
```
sanchda, line 77: Avoids a branch, nice.

sanchda (dismissed):
LGTM!

As you say, there's a performance penalty for doing this, but I agree that our priority is to ensure stability.  Once we fix the underlying issues, the performance will return.

include/dwfl_module.hpp
r1viollet: Exactly my thoughts :-)

include/dwfl_module.hpp
r1viollet: Local name should have been renamed, thank you

nsavoire (approved):

include/dwfl_symbol_lookup.hpp
```diff
@@ -77,8 +79,7 @@ class DwflSymbolLookup_V2 {
   DwflSymbolLookup_V2();
 
   // Get symbol from internal cache or fetch through dwarf
-  SymbolIdx_t get_or_insert(ddprof::DwflWrapper &dwfl_wrapper,
-                            SymbolTable &table,
+  SymbolIdx_t get_or_insert(DwflWrapper *dwfl_wrapper, SymbolTable &table,
```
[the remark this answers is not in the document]
r1viollet: 🤦  it came automatically to me when I did the forward declaration. The reference was better. Thanks for pointing that out.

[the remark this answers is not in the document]
r1viollet: Fixed ! 👍

PR #10, opened 2022-02-14
Send last results when we shutdown ddprof

r1viollet:
# What does this PR do?

# Motivation

# Additional Notes

# How to test the change?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -53,7 +53,7 @@ compile_commands.json
 *.dump
 .env
 .env_perso.yml
-test/configs/perso.yml
+.test_config.yml
 .editorconfig
 test/data/*.pprof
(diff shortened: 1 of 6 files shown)


sanchda (approved):
Not much for me to comment here.  Seems pretty reasonable--LGTM!

r1viollet:
> Does that matter? I don't think it does, _unless_ ddprof is at fault. Then we have compounding blame, which will be humiliating. Even then, I _still_ don't think it matters because it would still be humiliating to be in that situation even if the user had to enable final exports.

Thanks for reviewing this, @nsavoire took these changes and added them to his much larger PR. I have yet to come up with a proper way to test between build and development tooling.

r1viollet:
Closing as this was already incorporated and improved by @nsavoire

PR #11, opened 2022-02-14
Import demangler from llvm

nsavoire:
# What does this PR do?

Add llvm demangler in third_party/llvm directory.

# Motivation

* svn export from github is currently failing
* avoid depending from `main` version

# Additional Notes

# How to test the change?


third_party/llvm/include/llvm/Demangle/README.txt
```diff
@@ -0,0 +1,61 @@
+Itanium Name Demangler Library
```
sanchda, line 1: This Readme discusses the organization of the LLVM code, referencing certain features of that particular library.  It may be better to omit this so as to not be confusing, although removing it is also confusing because the implementation references this readme.  🤷

third_party/llvm/include/llvm/Demangle/StringView.h
```diff
@@ -0,0 +1,122 @@
+//===--- StringView.h ----------------*- mode:c++;eval:(read-only-mode) -*-===//
```
sanchda, line 1: It's certainly confusing that this file references the readme, since the readme references the organization of the LLVM codebase in a way that isn't reflected in our own code...

sanchda (approved):
Looks good to me.  I left an advisory comment on one of the files, but it's not necessary to do anything about it.

Thanks!

third_party/llvm/include/llvm/Demangle/StringView.h
nsavoire: I think it's best to not modify upstream files to avoid adding even more confusion :)

sanchda: Sorry, I definitely agree with that--I wasn't saying that we should change the file.  I was just commenting that it's a little confusing that there's no good way to make the  dependency comments unambiguous.

Upon reflection, I don't think there's anything we can do about that though.  🙃

PR #12, opened 2022-02-14
Add support for building with ninja

nsavoire:
Needed to change checks for output directory existence in fetch scripts since Ninja eagerly creates directories for generated files.

# What does this PR do?

Add support for building with [ninja](https://ninja-build.org/).

# Motivation

Ninja is lightweight and fast.

### DIFF
--- a/app/base-env/Dockerfile
+++ b/app/base-env/Dockerfile
@@ -34,6 +34,7 @@ RUN apt update \
   libjemalloc-dev \
   libcap-dev \
   netcat-openbsd \
+  unzip \
   wget
 
 # Only necessary for ubuntu 16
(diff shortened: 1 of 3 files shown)


app/base-env/Dockerfile
```diff
@@ -4,7 +4,7 @@ ARG UBUNTU_VERSION=20
 FROM ubuntu:${UBUNTU_VERSION}.04 as base
 
 ARG UBUNTU_VERSION
-ENV OS_IDENTIFIER="UB${UBUNTU_VERSION}"
+ENV OS_IDENTIFIERq="UB${UBUNTU_VERSION}"
```
sanchda, line 7: Typo?

tools/fetch_elfutils.sh
```diff
@@ -37,8 +37,7 @@ TAR_ELF="elfutils-${VER_ELF}.tar.bz2"
 URL_ELF="https://sourceware.org/elfutils/ftp/${VER_ELF}/${TAR_ELF}"
 
 if [ -e ${TARGET_EXTRACT} ]; then
-    echo "Error, clean the directory : ${TARGET_EXTRACT}"
-    exit 1
+    rm -rf ${TARGET_EXTRACT}
```
sanchda, line 40: Am I understanding this correctly?  Whenever cmake is run and the elfutils directory already exists, then:
1. Delete the directory
2. Download the tarball again
3. build the efultils dependency from scratch again

sanchda (commented):
Ultimately this looks good, but I think I don't quite understand some of the behavior around configuration-time dependency management.

r1viollet (approved):
LGTM
Thanks for now checking if directory is empty. I think it is better.

PR #13, opened 2022-02-16
Deliver native profiler as shared and static libraries

nsavoire:
# What does this PR do?

# Motivation

# Additional Notes

# How to test the change?


CMakeLists.txt
```diff
@@ -201,6 +201,29 @@ message(STATUS "Install destination " ${CMAKE_INSTALL_PREFIX})
...
+  # taken from https://dvdhrm.wordpress.com/2013/03/08/linking-binary-data/
+  COMMAND ld -r -o ${DDPROF_EXE_OBJECT} -z noexecstack --format=binary ddprof
+  COMMAND objcopy --rename-section .data=.rodata,alloc,load,readonly,data,contents ${DDPROF_EXE_OBJECT}
+  DEPENDS ddprof
+)
+
+add_library(ddprof_exe_object OBJECT IMPORTED GLOBAL)
+set_target_properties(ddprof_exe_object PROPERTIES IMPORTED_OBJECTS "${DDPROF_EXE_OBJECT}")
+
+add_library(ddprof_wrapper-static
```
r1viollet, line 216: Oh, that is cool, we can do both shared and static !

src/perf_mainloop.cc
```diff
@@ -5,72 +5,107 @@
...
 
 #define rmb() __asm__ volatile("lfence" ::: "memory")
 
-#define WORKER_SHUTDOWN()                                                      \
-  { return; }
+static pid_t g_child_pid = 0;
+static bool g_termination_requested = false;
+
+static void handle_signal(int signum) {
+  g_termination_requested = true;
```
r1viollet, line 39: there is indeed no other way to pass arguments to signal handlers

src/perf_mainloop.cc
```diff
@@ -85,47 +120,69 @@ static void pollfd_setup(const PEventHdr *pevent_hdr, struct pollfd *pfd,
...
     // If there was an issue, return and let the caller check errno
     if (-1 == n && errno == EINTR) {
       continue;
-    } else if (-1 == n) {
-      DDRES_CHECK_OR_SHUTDOWN(ddres_error(DD_WHAT_POLLERROR),
-                              attr->finish_fun(ctx));
+    }
+    DDRES_CHECK_ERRNO(n, DD_WHAT_POLLERROR, "poll failed");
+
+    if (pfd[signal_pos].revents & POLLIN) {
```
sanchda, line 167: I think this is fine, but I want to point out that this won't be revisited while any of the perf ringbuffers have events which can be drained.  Under normal operation, I never see that blocked for long periods of time.  However, I can't bound the time it would take for the processing loop to make its way back here.

src/perf_mainloop.cc
```diff
@@ -85,47 +120,69 @@ static void pollfd_setup(const PEventHdr *pevent_hdr, struct pollfd *pfd,
...
+      struct signalfd_siginfo siginfo;
+      if (read(pfd[signal_pos].fd, &siginfo, sizeof(siginfo)) !=
```
sanchda, line 169: So, the `signalfd()` was set on a mask comprised of SIGINT and SIGTERM.  Since the only thing we can do at this point is possibly fail to read the value (and exit anyway, except with an error unrelated to the original termination of the process), is it possible to just rely on the presence of `pfd[signal_pos].revents & POLLIN` as the canonical indicator that a shutdown signal was received?

src/perf_mainloop.cc
```diff
@@ -85,47 +120,69 @@ static void pollfd_setup(const PEventHdr *pevent_hdr, struct pollfd *pfd,
...
     // If one of the perf_event_open() feeds was closed by the kernel, shut down
     // profiling
-    for (int i = 0; i < pe_len; ++i) {
-      // This closes profiling when there still may be viable feeds, but we
-      // don't handle that case yet.
-      if (pfd[i].revents & POLLHUP) {
-        flags->errors = false; // This is a graceful shutdown!
-        DDRES_GRACEFUL_SHUTDOWN(attr->finish_fun(ctx));
-      }
+    if (std::any_of(pfd, pfd + pe_len,
```
sanchda, line 183: Sanity check:  I think you've changed the return semantics of this function enough to no longer require setting the error flag here--so this is good!

src/ddprof_worker.cc
```diff
@@ -188,7 +188,7 @@ DDRes ddprof_pr_sample(DDProfContext *ctx, perf_event_sample *sample, int pos) {
   return ddres_init();
 }
 
-static void ddprof_reset_worker_stats(DsoHdr *dso_hdr) {
```
sanchda, line 191: Ah, good change.

src/perf_mainloop.cc
```diff
@@ -144,29 +201,51 @@ static void worker(DDProfContext *ctx, const WorkerAttr *attr,
...
         }
       }
-    } while (events);
 
-    // If I didn't process any events, then hit the timeout
-    if (!processed_samples) {
-      if (IsDDResNotOK(ddprof_worker_timeout(&flags->can_run, ctx))) {
-        attr->finish_fun(ctx);
-        WORKER_SHUTDOWN();
+      DDRES_CHECK_FWD(ddprof_worker_maybe_export(restart_worker, ctx));
```
sanchda, line 218: The way you've consolidated the export logic is certainly much cleaner.  I think the changes in this area are fine.  In particular, checking for `!processed_samples` doesn't seem like a meaningful check in the current codebase.

At one point we had a sleep test.  As long as that passes, this should be good.

src/wrapper/ddprof_wrapper.cc
```diff
@@ -0,0 +1,159 @@
...
+      }
+      defer { close(fd); };
+
+      if (write(fd, _binary_ddprof_start,
+                _binary_ddprof_end - _binary_ddprof_start) == -1) {
+        return -1;
+      }
+
+      char path_buf[32];
+      snprintf(path_buf, sizeof(path_buf), "/proc/self/fd/%d", fd);
```
sanchda, line 95: Tricky!  The `memfd()` anonymous regions are backed by tmpfs, but referenced on the VFS only through procfs--I think this is probably the only way to get a path reference to once (which is weird, because the procfs fd entry should be a symlink, haha).

sanchda (approved):
Wow!  Some really comprehensive changes to the dirty muckity-muck of the mainloop code.  I tried to figure out how things might break, but it looks pretty solid to me so far.  Thanks!

src/perf_mainloop.cc
```diff
@@ -5,72 +5,107 @@
...
+  struct sigaction sa;
+  DDRES_CHECK_ERRNO(sigemptyset(&sigset), DD_WHAT_MAINLOOP_INIT,
+                    "sigemptyset failed");
+  sa.sa_handler = &handle_signal;
+  sa.sa_mask = sigset;
+  sa.sa_flags = SA_RESTART;
+  DDRES_CHECK_ERRNO(sigaction(SIGTERM, &sa, NULL), DD_WHAT_MAINLOOP_INIT,
+                    "sigaction failed");
+  sigaction(SIGINT, &sa, NULL);
+  DDRES_CHECK_ERRNO(sigaction(SIGTERM, &sa, NULL), DD_WHAT_MAINLOOP_INIT,
```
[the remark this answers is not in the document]
nsavoire: Oops, bad copy paste ;)
And registration for SIGINT is there but without error checking

src/perf_mainloop.cc
```diff
@@ -5,72 +5,107 @@
...
+  sigset_t sigset;
+  struct sigaction sa;
+  DDRES_CHECK_ERRNO(sigemptyset(&sigset), DD_WHAT_MAINLOOP_INIT,
+                    "sigemptyset failed");
+  sa.sa_handler = &handle_signal;
+  sa.sa_mask = sigset;
+  sa.sa_flags = SA_RESTART;
+  DDRES_CHECK_ERRNO(sigaction(SIGTERM, &sa, NULL), DD_WHAT_MAINLOOP_INIT,
+                    "sigaction failed");
+  sigaction(SIGINT, &sa, NULL);
```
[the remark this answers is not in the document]
nsavoire: My goal was to make sure that if ddprof receives either SIGINT/SIGTERM (and perhaps other that I missed, eg. SIGKILL), the signal is forwarded to the worker.
Indeed only SIGTERM is emitted by libddprof_wrapper, but I wanted to be more generic.

src/perf_mainloop.cc
nsavoire: I had the same interrogation ;)
I guess under heavy load of whole host profiling, we might be stuck in this loop...
We could remove this loop altogether and rely on poll, or limit the loop to a given number of iterations.
Another question, why not use epoll instead of poll ?

src/perf_mainloop.cc
nsavoire: Yes, that's the idea ! `worker` is now responsible for setting  `MLWorkerFlags`

src/perf_mainloop.cc
nsavoire: No, this is not good, you are right. I forgot to add back the export check in case of `poll` timeout.

src/wrapper/ddprof_wrapper.cc
nsavoire: Yes, it's not very clean, I could use `fexecve` but it does not offer the inheritance of environment as `execl`. But perhaps I could pass `environ` directly to `fexecve` ?

src/perf_mainloop.cc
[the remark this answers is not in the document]
sanchda: Ah, I misunderstood--I see now.  I originally thought that this was being called on the library side, before ddprof daemonizes.  I see now that it is called afterward, in order to propagate signals to the child (as you say).

Sorry about that--this is great!

src/perf_mainloop.cc
sanchda: Right, an epoll loop would be much better here, especially since we don't need to constantly re-register file descriptors.

src/perf_mainloop.cc
sanchda: Yup,that's right!

src/wrapper/ddprof_wrapper.cc
sanchda: And for the record, I personally think this is pretty close to as clean as it could be.  The interfaces are rather limited.  When I said it was tricky, I meant that the idea itself has several very subtle implications, like the fact that you're making a memfd and then executing it, but there's no conspicuous canonical path other than in `/proc`.  I think this is great!

src/perf_mainloop.cc
```diff
@@ -0,0 +1,290 @@
...
+        // We've processed the current event, so we can advance the ringbuffer
+        rb->region->data_tail += hdr->size;
+
+        // Check for processing error
+        if (IsDDResNotOK(res)) {
+          return res;
+        }
+      }
+
+      now_ns = now_nanos();
```
sanchda, line 217: `now_nanos()` is going to issue a call to `gettimeofday()`, which in many contemporary cloud configurations will skip vDSO, emitting a syscall.  I would like this to be a problem in the future, but it isn't a problem now.

For samples this is (unfortunately!!!!!!) not a problem because libdwfl already calls many aggressive syscalls all the time 🙃 .

For events which are not samples, this will generally change the sample from a pure in-process computation to a syscall.  This might have some disruptions to our event processing queue, but I think this is probably better for now.

sanchda (dismissed):
Good enhancements--good catch fixing the timeout issue.  I had a feeling, but honestly I didn't trace through the calls properly in my initial review.

src/perf_mainloop.cc
sanchda: Followup comment:  my presumption that we weren't already calling `gettimeofday()` in this loop was wrong, and my comments are wrong.  Keeping them for the sake of exposition, but they are wrong.

sanchda (dismissed):
Looks good!

src/wrapper/ddprof_wrapper.cc
```diff
@@ -0,0 +1,191 @@
...
+  pid_t middle_pid = fork();
+  if (middle_pid == -1) {
+    return -1;
+  }
+
+  if (middle_pid == 0) {
+    // executed in middle
+    close(pipefd[0]);
+
+    auto defer_pipeclose = make_defer([&pipefd] { close(pipefd[1]); });
```
r1viollet, line 92: This is lovely. Much better than creating handler structs for all the things.

r1viollet (approved):
Very thoughtful updates ! Especially considering you just got into the codebase 👏 .

src/wrapper/ddprof_wrapper.cc
```diff
@@ -101,13 +102,18 @@ int ddprof_start_profiling() {
       grandchild_pid = getpid();
       char ddprof_str[] = "ddprof";
 
+#if __GLIBC_PREREQ(2, 27)
```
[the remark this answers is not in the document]
nsavoire: No strong opinion about using the wrapper, I changed for syscall only.

r1viollet (approved):
Big one ! 🚢

PR #15, opened 2022-02-17
Add detection of architecture

r1viollet:
# What does this PR do?

This uses a trick that runs a C program to output the architecture.
For now we just exit if we do not encounter the expected architecture. This is a preparation for arm support.

# Motivation

Detect in build system the architecture we are compiling on.

# Additional Notes

I am not the author. All credit goes to the author and to the Qt folks.
 
# How to test the change?

This will exit the cmake script with an error if we are on a platform different from x86.


r1viollet:
Closing this as there are more simple ways of achieving this (CMAKE_HOST_SYSTEM_PROCESSOR).

PR #16, opened 2022-02-18
[WIP] Feature/arm

sanchda:
Will fill this out when this is closer to completion.

Outstanding unwinding issues:

Sometimes frames are emitted with 0 value for IP, but symbolized properly.  Lines below copied from raw pprof locations listing in global mode
```
18: 0x0 M=3  /usr/bin/tmux:0 s=0
19: 0xaaaac2aaee7c M=5  /usr/bin/tmux:0 s=0
```

Sometimes the heap is emitted as a viable location
```
16: 0x0 M=3  Heap:0 s=0
 ```
 
 ...and other random garbage too!
 ```
11: 0x0 M=3  /home/sanchda/dev/ddprof_arm/build/ddprof:0 s=0
12: 0x0 M=3 [unknown_dso] :0 s=0()
13: 0x0 M=3  /usr/bin/sha256sum:0 s=0
14: 0x0 M=3 [0x51dc:file] /usr/bin/sha256sum:0 s=0
 ```
 
 Sometimes 0 mappings are populated (other examples above)
 ```
76: 0x0 M=3 [0x4bf4:file] /usr/bin/sha256sum:0 s=0
77: 0x0 M=3 [0x418c:file] /usr/bin/sha256sum:0 s=0
 ... (then pasted from the mappings section)
3: 0x0/0x0/0x0
 ```

### CODE REVIEW

**Comment by r1viollet** [comment]
Thanks for opening this ! I'm sending commits directly over here. Feel free to revert anything you would not agree with ! 🙇

**Comment by sanchda** [comment]
This is ready for review, but it cannot be merged into it incorporates and generalizes certain fixes from PR #23 which have to do with processor specification.

File: src/dwfl_thread_callbacks.cc [hunk 1]
```diff
@@ -13,16 +13,80 @@ pid_t next_thread(Dwfl *dwfl, void *arg, void **thread_argp) {
...
+  regs[n++] = us->initial_regs.regs[REGNAME(EBP)];
+  regs[n++] = us->initial_regs.regs[REGNAME(ESP)];
+  regs[n++] = us->initial_regs.regs[REGNAME(R8)];
+  regs[n++] = us->initial_regs.regs[REGNAME(R9)];
+  regs[n++] = us->initial_regs.regs[REGNAME(R10)];
+  regs[n++] = us->initial_regs.regs[REGNAME(R11)];
+  regs[n++] = us->initial_regs.regs[REGNAME(R12)];
+  regs[n++] = us->initial_regs.regs[REGNAME(R13)];
+  regs[n++] = us->initial_regs.regs[REGNAME(R14)];
+  regs[n++] = us->initial_regs.regs[REGNAME(R15)];
```
r1viollet, line 42: I think x86 will not work. we are miss-aligned here. initial regs are not as big as the reg structure.
The naming is nice to document and debug things, though having a generic copy structure might be easier. I will let you decide what you prefer.

src/dwfl_thread_callbacks.cc
```diff
@@ -13,16 +13,80 @@ pid_t next_thread(Dwfl *dwfl, void *arg, void **thread_argp) {
   return us->pid;
 }
 
+// DWARF and the Linux kernel don't have a uniform view of the processor, so we
+// can't just follow the register mask and shove it into the output registers.
+// Instead, we crib off of libdwfl's ARM/x86 unwind code in elfutil's
```
nsavoire, line 18: I could not find unwind-libdw.c in elfutils, you mean from perf ?

src/perf_mainloop.cc
```diff
@@ -31,7 +31,13 @@ extern "C" {
 
 #include "defer.hpp"
 
-#define rmb() __asm__ volatile("lfence" ::: "memory")
```
nsavoire, line 34: I already replaced the memory barrier by atomic_load/acquire and atomic_store/release that should mimic what is provided by ring_buffer API (https://github.com/torvalds/linux/blob/v5.16/tools/include/linux/ring_buffer.h#L59).
But I forgot to remove rmb definition ;)

r1viollet (approved):
LGTM
That will be a super cool thing to give to our users ! 👏

src/perf_mainloop.cc
r1viollet: Removed in latest commit

src/dwfl_thread_callbacks.cc
sanchda: I don't see any changes here.  Is the repo missing a commit?

r1viollet: Changes are actually pushed to the unwind_state structure to align things.

src/dwfl_thread_callbacks.cc
sanchda: Yes, I do--will update.  Thanks!

PR #19, opened 2022-02-21
Fix test for breakpoint

nsavoire:
# What does this PR do?

PerfOption::type is not a bitmap, `==` should be used instead of `&`.
This had the effect of using PERF_COUNT_SW_CPU_CLOCK instead of PERF_COUNT_SW_TASK_CLOCK.

# Motivation

# Additional Notes

# How to test the change?

### DIFF
--- a/src/perf.c
+++ b/src/perf.c
@@ -69,7 +69,7 @@ int perfopen(pid_t pid, const PerfOption *opt, int cpu, bool extras) {
   attr.freq = opt->freq;
 
   // Breakpoint
-  if (opt->type & PERF_TYPE_BREAKPOINT) {
+  if (opt->type == PERF_TYPE_BREAKPOINT) {
     attr.config = 0; // as per perf_event_open() manpage
     attr.bp_type = opt->bp_type;
   }


src/perf.c
```diff
@@ -69,7 +69,7 @@ int perfopen(pid_t pid, const PerfOption *opt, int cpu, bool extras) {
   attr.freq = opt->freq;
 
   // Breakpoint
-  if (opt->type & PERF_TYPE_BREAKPOINT) {
+  if (opt->type == PERF_TYPE_BREAKPOINT) {
```
sanchda, line 72: Ah, good catch.  This was my mistake.  Many of the configurations for these interfaces are bitmasks, but this is not the case with `perf_event_open()`'s type field of its input attr.

For a similar mistake, check out https://github.com/DataDog/ddprof/blob/main/src/ddprof_worker.cc#L161.  There we correctly use `==`, but checked the enum against the `type` field rather than the `config` field.  The consequence of the failed check are rather limited there as well.

sanchda (approved):
LGTM--thanks!

This is unfortunate, since the breakpoint stuff was actually part of an R&D to explore different ways of implementing memory profiling.  It has no relevance in the current codebase (although it might very soon with the OOM stuff, so I'm reluctant to propose its removal just yet).

The distinction between the clock we intended and the clock we got is somewhat limited for pure-userspace applications--in other words, I don't think this mistake is more harmful than it is embarrassing (for me).  You can check out the update implementation between the two of them:

https://github.com/torvalds/linux/blob/c5d714aa6dce262482992cccc0ab01383294f7ac/kernel/events/core.c#L10847
https://github.com/torvalds/linux/blob/c5d714aa6dce262482992cccc0ab01383294f7ac/kernel/events/core.c#L10770

You can see that the CPU clock gets its own `now` time in its update handler, whereas the task clock receives the `now` time from the event context.  I don't know enough these code paths to speculate properly, but my assumption is that event generation is very close to this update for user code, so the two clocks tend to agree (I'd be curious to learn whether one is more appropriate for systems under extreme CPU load, though).

For profiling which may include kernel stacks, I believe the CPU one is preferred, but I'd need to double-check.

nsavoire:
> LGTM--thanks!
> 
> This is unfortunate, since the breakpoint stuff was actually part of an R&D to explore different ways of implementing memory profiling. It has no relevance in the current codebase (although it might very soon with the OOM stuff, so I'm reluctant to propose its removal just yet).
> 
> The distinction between the clock we intended and the clock we got is somewhat limited for pure-userspace applications--in other words, I don't think this mistake is more harmful than it is embarrassing (for me). You can check out the update implementation between the two of them:
> 
> https://github.com/torvalds/linux/blob/c5d714aa6dce262482992cccc0ab01383294f7ac/kernel/events/core.c#L10847 https://github.com/torvalds/linux/blob/c5d714aa6dce262482992cccc0ab01383294f7ac/kernel/events/core.c#L10770
> 
> You can see that the CPU clock gets its own `now` time in its update handler, whereas the task clock receives the `now` time from the event context. I don't know enough these code paths to speculate properly, but my assumption is that event generation is very close to this update for user code, so the two clocks tend to agree (I'd be curious to learn whether one is more appropriate for systems under extreme CPU load, though).
> 
> For profiling which may include kernel stacks, I believe the CPU one is preferred, but I'd need to double-check.

I agree, from my experiments, both were very close, I don't think this had any impact.

r1viollet (approved):
LGTM
Great catch !

PR #21, opened 2022-02-24
Activate C++20 and add clang-tidy checks

nsavoire:
# What does this PR do?

* Compile in C++20 mode
* Enable clang-tidy checks during compilation when`-DENABLE_CLANG_TIDY=ON` is passed to cmake
* Fix some clang-tidy warnings

# Motivation

* Add more static analysis tools
* Benefit from latest C++ version

# Additional Notes

# How to test the change?

### DIFF
--- a/cmake/ExtendBuildTypes.cmake
+++ b/cmake/ExtendBuildTypes.cmake
@@ -6,6 +6,9 @@ if(NOT CMAKE_BUILD_TYPE)
   set(CMAKE_BUILD_TYPE "Release" CACHE STRING "CMake build type" FORCE)
 endif()
 
+set(CMAKE_CXX_STANDARD 20)
+set(CMAKE_C_STANDARD 11)
+
 add_compile_options(-Wall)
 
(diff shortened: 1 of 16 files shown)


cmake/ExtendBuildTypes.cmake
```diff
@@ -6,6 +6,9 @@ if(NOT CMAKE_BUILD_TYPE)
   set(CMAKE_BUILD_TYPE "Release" CACHE STRING "CMake build type" FORCE)
 endif()
 
+set(CMAKE_CXX_STANDARD 20)
```
r1viollet, line 9: Global C++ version is good. I had trouble finding this 👍

nsavoire: I don't know if this is the right place to put it, if you have a better idea, I am all ears :)

r1viollet: I was thinking at the start of the general cmake file, but this is also fine.

r1viollet (approved):
LGTM
Yay string views are here !

PR #23, opened 2022-02-28
Deliver collatz version linked with libddprof_wrapper

nsavoire:
# What does this PR do?

* Deliver ddprof / collatz / selfuw to their own install dir -> first step at refactoring delivery to S3
* Deliver a version of collatz linked dynamically with libddprof_wrapper, this version will be deployed to rel-env

# Motivation

# Additional Notes

* ddprof-build needs to be updated to account for these changes

# How to test the change?


CMakeLists.txt
```diff
@@ -265,15 +265,10 @@ if (${BUILD_NATIVE_LIB})
    list(APPEND NATIVE_LIB_LIBRARY_LIST libcap)
    list(APPEND NATIVE_LIB_INCLUDE_LIST ${LIBCAP_INCLUDE_DIR})
 
-   ## Headers we publish
-   set(LIB_HEADERS include/ddprof.h include/ddprof_context_lib.h include/ddprof_input.h include/lib/ddprof_output.hpp include/symbol_table.hpp include/mapinfo_table.hpp include/ddres.h include/ddres_def.h include/ddres_list.h include/ddres_helpers.h include/ddres_exception.hpp)
```
r1viollet, line 269: I agree we don't need this as they are for internal testing capabilities.

include/dwfl_internals.h
```diff
@@ -5,5 +5,5 @@
 
 #pragma once
 
-#include "libdw.h"
-#include "libdwfl.h"
+#include "elfutils/libdw.h"
```
sanchda, line 8: Yeah, good call--we were lacking this.

tools/fetch_elfutils.sh
```diff
@@ -58,5 +60,5 @@ echo "Compile elfutils using ${C_COMPILER}"
 if [[ "$(basename "${C_COMPILER}")" == clang* ]]; then
   export CFLAGS="-Wno-xor-used-as-pow -Wno-gnu-variable-sized-type-not-at-end -Wno-unused-but-set-parameter"
 fi
-cd ${TARGET_EXTRACT} && ./configure CC=${C_COMPILER} --disable-debuginfod --disable-libdebuginfod --disable-symbol-versioning
-make -j4 -C ${TARGET_EXTRACT}
+./configure CC="${C_COMPILER}" --disable-debuginfod --disable-libdebuginfod --disable-symbol-versioning --prefix "${TARGET_EXTRACT}"
+make "-j$(nproc)" install
```
sanchda, line 64: Just thinking out loud here, but `nproc` uses `sched_getaffinity()` to get the process CPU mask.  In a lot of cloud environments, resource isolation is implemented via cycle counting instead of CPU affinity, meaning that `sched_getaffinity()` will overestimate the compute capacity available to the cgroup.

I don't think this matters at all.  In fact, our builds are small enough that `make -j` would probably be fine as well.

sanchda (approved):
LGTM

bench/collatz/CMakeLists.txt
```diff
@@ -18,4 +18,12 @@ add_exe(collatz
...
 
-install(TARGETS collatz)
+add_exe(collatz-ddprof_wrapper-shared
+        ${COLLATZ_SRC}
+        LIBRARIES Threads::Threads ddprof_wrapper-shared
+        DEFINITIONS ${COLLATZ_DEFINITION_LIST} USE_DDPROF_WRAPPER)
+target_include_directories(collatz-ddprof_wrapper-shared PRIVATE ${STATSD_INCLUDE_DIR} ${LIBDDPROF_INCLUDE_DIR})
+
+install(TARGETS collatz DESTINATION collatz)
+install(TARGETS collatz-ddprof_wrapper-shared DESTINATION collatz)
```
r1viollet, line 28: Proposal for naming: 
```
collatz-auto-prof
```
As in automatic profiling

tools/fetch_elfutils.sh
nsavoire: `make -j` will not limit the number of jobs, I think I still prefer `nproc` even if it is slightly wrong

cmake/Findelfutils.cmake
```diff
@@ -9,28 +9,20 @@
...
-set(LIBDW_PATH ${ELFUTILS_PATH}/libdw/libdw.a CACHE FILEPATH "location of dw lib")
-set(LIBELF_PATH ${ELFUTILS_PATH}/libelf/libelf.a CACHE FILEPATH "location of elf lib")
-set(ELFUTILS_LIBS 
-    "${LIBDW_PATH}"
-    "${LIBELF_PATH}")
+set(LIBDW_PATH ${ELFUTILS_PATH}/lib/libdw.a)
+set(LIBELF_PATH ${ELFUTILS_PATH}/lib/libelf.a)
 
-list(APPEND ELFUTILS_INCLUDE_LIST ${ELFUTILS_PATH} ${ELFUTILS_PATH}/libdwfl ${ELFUTILS_PATH}/libdw ${ELFUTILS_PATH}/libebl ${ELFUTILS_PATH}/libelf)
+list(APPEND ELFUTILS_INCLUDE_LIST ${ELFUTILS_PATH} ${ELFUTILS_PATH}/include)
```
r1viollet, line 17: Using only public APIs is also better. thanks

r1viollet (approved):
Significant changes in the build system ! Thanks
We might want to be sure about the name as this will ship to users.

r1viollet (dismissed):
LGTM

r1viollet (approved):
LGTM

PR #24, opened 2022-03-01
Sanchda/arm fixups

sanchda:
# What does this PR do?

Fixes a few issues with the ARM build.  Not sure why these weren't caught in CI

# Motivation

To not break main

# Additional Notes

I'm not sure why this happened, since it seems pretty obvious that this should break in CI.  I can investigate sometime, but in the meantime we should probably incorporate these changes to stabilize main.  Feel free to merge on my behalf if this is approved.

### DIFF
--- a/src/unwind.cc
+++ b/src/unwind.cc
@@ -36,7 +36,7 @@ void unwind_init_sample(UnwindState *us, uint64_t *sample_regs,
   uw_output_clear(&us->output);
   memcpy(&us->initial_regs.regs[0], sample_regs,
          K_NB_REGS_UNWIND * sizeof(uint64_t));
-  us->current_ip = us->initial_regs.regs[REGNAME(EIP)];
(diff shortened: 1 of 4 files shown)


nsavoire (approved):
LGTM !
I guess the issue was not detected because we do not have an ARM CI yet.

src/unwind.cc
```diff
@@ -36,7 +36,7 @@ void unwind_init_sample(UnwindState *us, uint64_t *sample_regs,
   uw_output_clear(&us->output);
   memcpy(&us->initial_regs.regs[0], sample_regs,
          K_NB_REGS_UNWIND * sizeof(uint64_t));
-  us->current_ip = us->initial_regs.regs[REGNAME(EIP)];
+  us->current_ip = us->initial_regs.regs[REGNAME(PC)];
```
r1viollet, line 39: my bad, thanks for fixing

PR #25, opened 2022-03-03
Arm compatibility - build image update

r1viollet:
# What does this PR do?

Update dockerfile to ensure it can build on arm platforms

# Motivation

Arm compatibility

# Additional Notes

N/A

# How to test the change?

CI builds on ARM

### DIFF
--- a/app/base-env/Dockerfile
+++ b/app/base-env/Dockerfile
@@ -27,7 +27,7 @@ FROM base-${UBUNTU_VERSION} AS base-clang
 ENV CC=clang-${CLANG_VERSION}
 ENV CXX=clang++-${CLANG_VERSION}
 
-FROM base-${COMPILER} AS dependencies
+FROM base-${COMPILER} AS final
 
 # Tell docker to use bash as the default
 SHELL ["/bin/bash", "-c"]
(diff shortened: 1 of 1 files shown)


app/base-env/Dockerfile
```diff
@@ -70,20 +70,30 @@ RUN /llvm.sh ${CLANG_VERSION} && /gcc.sh ${GCC_VERSION}
 ENV PATH="/usr/lib/llvm-${CLANG_VERSION}/bin/:$PATH"
 
 # Newer CMake
+FROM dependencies as cmake
```
nsavoire, line 73: What is the advantage of starting a new stage here ?

app/base-env/Dockerfile
```diff
@@ -70,20 +70,30 @@ RUN /llvm.sh ${CLANG_VERSION} && /gcc.sh ${GCC_VERSION}
...
+  && if [ "$MARCH" = aarch64 ]; then SHA256=$SHA256_ARM; else SHA256=$SHA256_X86; fi \
+  && TAR_NAME="cmake-${VERSION}-Linux-${MARCH}.tar.gz" \
+  && curl -LO https://github.com/Kitware/CMake/releases/download/v${VERSION}/${TAR_NAME} \
+  && (printf "${SHA256}  ${TAR_NAME}" | sha256sum --check --strict --status) \
+  && tar --no-same-owner -C /usr/local --strip-components=1 -xf ${TAR_NAME} \
+  && rm ${TAR_NAME}
 
+
+
+FROM cmake as final
```
nsavoire, line 87: Same question as above

app/base-env/Dockerfile
r1viollet: None for now. I was thinking about what we can parallelise with buildkit later on. The steps after cmake could all start together. But I did not implement it. I can remove it

nsavoire (approved):

PR #33, opened 2022-03-07
Avoid flaky unit test

r1viollet:
# What does this PR do?

Testing that the pprof was created and encoded in the same second, generates random failures.
The proposal here is to consider a 2 second difference (instead of within the same second).

# Motivation

Fixing random failures in a unit test

### DIFF
--- a/test/ddprof_pprof-ut.cc
+++ b/test/ddprof_pprof-ut.cc
@@ -45,7 +45,7 @@ void test_pprof(const DDProfPProf *pprofs) {
 
   // Check that encoded time is close to now
   time_t local_time = time(NULL);
-  EXPECT_TRUE(local_time - start.seconds < 1);
+  EXPECT_TRUE(local_time - start.seconds < 2);
 
   ddprof_ffi_Buffer profile_buffer = {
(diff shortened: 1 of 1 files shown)


test/ddprof_pprof-ut.cc
```diff
@@ -45,7 +45,7 @@ void test_pprof(const DDProfPProf *pprofs) {
 
   // Check that encoded time is close to now
   time_t local_time = time(NULL);
-  EXPECT_TRUE(local_time - start.seconds < 1);
+  EXPECT_TRUE(local_time - start.seconds < 2);
```
sanchda, line 48: Ah, right.  Since `time()` only returns second-level granularity, if the true time is, say, `x.999` seconds (truncated to `x`), then the subsequent time may be, say `(x+1).001` so even though the absolute times are quite close, the second-by-second epochs will not pass (`1` is not greater than `(x+1) - x`).

Since we record time in the pprof at a greater granularity than seconds, it may also be possible to use a different time interface.  I have no idea what an appropriate tolerance is; for the time being, I feel that this check is OK.

sanchda (approved):
LGTM, thanks for making tests less flaky!

PR #36, opened 2022-03-09
Prevent install of llvm if it is not necessary

r1viollet:
# What does this PR do?

Prevent installation of llvm if we do not need it.

# Motivation

ARM builds on ubuntu 16 are failing.

### DIFF
--- a/app/base-env/Dockerfile
+++ b/app/base-env/Dockerfile
@@ -64,7 +64,8 @@ RUN curl -L https://github.com/github/codeql-action/releases/download/codeql-bun
 ## LLVM/GCC SETUP ##
 ####################
 ADD ./app/base-env/llvm.sh ./app/base-env/gcc.sh /
-RUN /llvm.sh ${CLANG_VERSION} && /gcc.sh ${GCC_VERSION}
+RUN if [ "$CC" = clang* ]; then /llvm.sh ${CLANG_VERSION}; fi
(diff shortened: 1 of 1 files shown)


sanchda (dismissed):
Right...  Of course there'd be different support here.

Overall though, if we don't need it for the build, it probably shouldn't be in the base env--so this is a good patch even if ARM was passing in the first place.

sanchda (dismissed):
LGTM

app/base-env/Dockerfile
```diff
@@ -64,7 +64,9 @@ RUN curl -L https://github.com/github/codeql-action/releases/download/codeql-bun
 ## LLVM/GCC SETUP ##
 ####################
 ADD ./app/base-env/llvm.sh ./app/base-env/gcc.sh /
-RUN /llvm.sh ${CLANG_VERSION} && /gcc.sh ${GCC_VERSION}
+ARG FORCE_LLVM_INSTALL="OFF"
+RUN if [ "$CC" == clang* ] || [ ${FORCE_LLVM_INSTALL} == "ON" ]; then /llvm.sh ${CLANG_VERSION}; fi
```
nsavoire, line 68: I would rather have an opt-out rather than an opt-in here:
* it's nice to have clang-format/clang-tidy in all the images
* build:coverage depends on clang-format

nsavoire (approved):
LGTM

PR #37, opened 2022-03-09
Kill profiler if timeout is reached

nsavoire:
# What does this PR do?

# Motivation

# Additional Notes

# How to test the change?

### DIFF
--- a/src/lib/dd_profiling.cc
+++ b/src/lib/dd_profiling.cc
@@ -179,6 +179,8 @@ void ddprof_stop_profiling(int timeout_ms) {
     return;
   }
 
+  defer { g_state.started = false; };
+
   auto time_limit =
       std::chrono::steady_clock::now() + std::chrono::milliseconds(timeout_ms);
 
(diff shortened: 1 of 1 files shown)


sanchda (approved):
Nice--looks good to me!

My only comment is that the initial `kill()` (line 187) and the check (line 191) will occur very close to each other in most systems, which may force a third `kill()` to be issued.  I don't think this matters, since this guarantees that `stop(0)` will send the kill signal at least one time, which I feel is important.  I sense it is better for `stop(0)` to do nonblocking stop rather than not stop at all.

nsavoire:
> Nice--looks good to me!
> 
> My only comment is that the initial `kill()` (line 187) and the check (line 191) will occur very close to each other in most systems, which may force a third `kill()` to be issued. I don't think this matters, since this guarantees that `stop(0)` will send the kill signal at least one time, which I feel is important. I sense it is better for `stop(0)` to do nonblocking stop rather than not stop at all.

I moved the sleep before the check, does that sound good to you ?

sanchda (approved):
LGTM

sanchda:
> > Nice--looks good to me!
> > My only comment is that the initial `kill()` (line 187) and the check (line 191) will occur very close to each other in most systems, which may force a third `kill()` to be issued. I don't think this matters, since this guarantees that `stop(0)` will send the kill signal at least one time, which I feel is important. I sense it is better for `stop(0)` to do nonblocking stop rather than not stop at all.
> 
> I moved the sleep before the check, does that sound good to you ?

Yup--the other way looked good, but this way looks better.  I expect that in the typical case, only one `kill()` needs to be sent.  It doesn't _really_ matter, but it could prevent an unnecessary 10ms wait when the user's application is performing its own time-critical shutdown operation.

PR #39, opened 2022-03-14
Statically link libdd_profiling.so with libstdc++ and export only profiling API

nsavoire:
# What does this PR do?

Statically link libdd_profiling.so with libstdc++ and export only profiling API

# Motivation

Ensure that libdd_profiling.so does not depend on libstdc++.

# Additional Notes

All symbols from libdd_profiling.so are hidden (except start/stop profiling functions) to avoid collisions with client code (especially for liibstdc++ symbols).


cmake/dd_profiling.version
```diff
@@ -0,0 +1,4 @@
+LIBDD_PROFILING {
```
r1viollet, line 1: This is very nice ! Great idea

r1viollet (approved):
LGTM !

cmake/dd_profiling.version
sanchda: Yeah, this works pretty well.  Listing only the global exported functions from the .so in CI:
```
$ readelf -Ws --dyn-syms libdd_profiling.so | grep 'GLOBAL DEFAULT' | grep -v UND | grep FUNC
    60: 0000000000003700   736 FUNC    GLOBAL DEFAULT   13 ddprof_start_profiling@@LIBDD_PROFILING
    61: 00000000000039e0   257 FUNC    GLOBAL DEFAULT   13 ddprof_stop_profiling@@LIBDD_PROFILING
   415: 00000000000039e0   257 FUNC    GLOBAL DEFAULT   13 ddprof_stop_profiling
   428: 0000000000003700   736 FUNC    GLOBAL DEFAULT   13 ddprof_start_profiling
```

There are plenty of references to other symbols, but they are marked local.  My understanding is that such symbols are not used at link-time to satisfy unresolved dependencies, which prevents symbol collisions.

sanchda (dismissed):
Good fix!  In the previous version, I was thinking that symbol visibility didn't matter due to the fact that we are executing the actual profiler in a new process image, but obviously the wrapper will bring something (transitively, via its dependencies), so I was wrong.

Linking in this way also minimizes the dynamic dependencies of the wrapper:
```
$ ldd libdd_profiling.so
	linux-vdso.so.1 (0x00007ffcc93f9000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f34ddcb7000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f34de7e7000)
```

As far as I understand, this is an _ideal_ instrumentation wrapper in a lot of very important ways.  Really, really good stuff!

cmake/dd_profiling.version
nsavoire: Yes, this is how I understand it, local symbols won't be used to resolve undefined symbols outside of the library.
I don't think `LIBDD_PROFILING` version is useful, I will remove it.

sanchda (approved):
LGTM

PR #40, opened 2022-03-15
Fix small typo in Build.md

peterg17:
Command should be `RelCMake` instead of `RelCmake` from setup_env.sh


sanchda (approved):
```
$ grep -nri relcmake
setup_env.sh:18:RelCMake() {
docs/Build.md:68:RelCmake ../
```

Whoopsies!  Good find; please feel free to merge!

PR #41, opened 2022-03-18
Tracepoints

r1viollet:
# What does this PR do?

Resuming @sanchda's work around tracepoint capabilities.

# Motivation

Getting relevant information around context switches.

# Additional Notes

Several bugs remain in this PR.


src/perf_option.c
```diff
@@ -51,6 +52,8 @@ const PerfOption *perfoptions_preset(int idx) {
   return &perfoptions[idx];
 }
 
+int perfoptions_get_tracepoint_idx() { return perfoptions_sz - 1; }
```
r1viollet, line 55: todo: adapt. slightly likely to break this way

PR #44, opened 2022-03-22
Fix check of dependencies

nsavoire:
# What does this PR do?

Check for actual libraries instead of just their count.

# Motivation

In the current state, adding a new dependency would not be detected because list in `expected_deps...` is too long.


tools/check_dependencies.sh
```diff
@@ -15,14 +15,15 @@ if [ $# != 2 ] || [ $1 == "-h" ]; then
     exit 0
 fi
 
-# Poor man's test, I'll come back to it
```
r1viollet, line 18: haha, never came back to it
Thanks for improving this.

r1viollet (approved):
LGTM

r1viollet (approved):
LGTM

PR #46, opened 2022-03-24
Add utility function to save context

nsavoire:
# What does this PR do?

Add utility function to save context (registers + stack) (registers + stack) necessary to perform remote unwinding.
Only works for x86_64 for now, I will do the implementation for ARM later.

Getting a green build was a battle with cppcheck / address sanitizer:
https://gitlab.ddbuild.io/DataDog/ddprof/-/pipelines/7440633

I added a third-party implementation for std::span because gcc 9 does not provide it. We could switch to gcc 11 on 16.04 but this would require compiling our own gcc.

Also added support for google-benchmark.

### DIFF
--- a/LICENSE-3rdparty.csv
+++ b/LICENSE-3rdparty.csv
@@ -3,3 +3,4 @@ elfutils,"https://sourceware.org/elfutils/",LGPL-3.0-or-later,Free Software Foun
 libcap,"https://git.kernel.org/pub/scm/libs/libcap/libcap.git/",BSD-3-Clause,Free Software Foundation 
 liblzma,"https://github.com/kobolabs/liblzma",Public-Domain,NA
 llvm-demangle,"https://github.com/llvm/llvm-project",Apache-2.0,llvm-project
+tcbrindle/span,"https://github.com/tcbrindle/span",Boost Software License 1.0
--- a/src/savecontext.cc
+++ b/src/savecontext.cc
@@ -1,0 +1,108 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0. This product includes software
+// developed at Datadog (https://www.datadoghq.com/). Copyright 2021-Present
+// Datadog, Inc.
+
+#include "savecontext.hpp"
+
+#include <algorithm>
+#include <cstring>
+#include <pthread.h>
+
+#define XSTR(s) STR(s)
+#define STR(s) #s
+
+#ifdef __x86_64__
+// hardcode register index in mask
+// Did not find a way to reuse enum values from perf-archmap.h
+#  define RBX_IDX 1
+#  define RBP_IDX 6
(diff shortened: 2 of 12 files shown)


LICENSE-3rdparty.csv
```diff
@@ -3,3 +3,4 @@ elfutils,"https://sourceware.org/elfutils/",LGPL-3.0-or-later,Free Software Foun
 libcap,"https://git.kernel.org/pub/scm/libs/libcap/libcap.git/",BSD-3-Clause,Free Software Foundation 
 liblzma,"https://github.com/kobolabs/liblzma",Public-Domain,NA
 llvm-demangle,"https://github.com/llvm/llvm-project",Apache-2.0,llvm-project
+tcbrindle/span,"https://github.com/tcbrindle/span",Boost Software License 1.0
```
r1viollet, line 6: Ah yes, gcc 9 is too old so we pull in tcbrindle. 👍

test/savecontext-ut.cc
```diff
@@ -0,0 +1,54 @@
...
+
+void funcA() __attribute__((noinline));
+void funcB() __attribute__((noinline));
+
+std::byte stack[PERF_SAMPLE_STACK_SIZE];
+
+void funcB() {
+  UnwindState state;
+  uint64_t regs[K_NB_REGS_UNWIND];
+  size_t stack_size = save_context(regs, stack);
```
r1viollet, line 24: I thought span construction was explicit. TIL

src/savecontext.cc
```diff
@@ -0,0 +1,98 @@
...
+    asm (
+    // Retrieve caller save registers
+    // Callee save register are not needed since they could contain anything after function return
+    SAVE_REG(RBX)
+    SAVE_REG(RBP)    
+    SAVE_REG(R12)
+    SAVE_REG(R13)
+    SAVE_REG(R14)
+    SAVE_REG(R15)
+    // Bump the stack by 8 bytes to remove the return address,
```
r1viollet, line 45: thanks for the comments !

r1viollet (dismissed):
Impressive mechanism ! 👏 👏 👏  Can't wait to try it out.
My review does not cover the asm part.

LICENSE-3rdparty.csv
nsavoire: The other solution would be to update to gcc 11 on ubuntu 16.04.

test/savecontext-ut.cc
nsavoire: Constructors from C arrays / std::array are not explicit, which is nice !

src/savecontext.cc
nsavoire: Only works for x86_64 for now, I will do the implementation for ARM later.

r1viollet (approved):
LGTM

PR #47, opened 2022-03-25
Fix flaky test

nsavoire:
# What does this PR do?

Turn error into a warning, dso in /proc/%pid%/root/... disappear when process exists, therefore RegionHolder might fail to open files.

# Motivation

Fix `native_bench.sh ${RECORD_OPTION} -b ${BINDIR} -e sleep.sh` test in CI that does many short `sleep`.


r1viollet (approved):
Agreed ! Already discussed in another thread
Thanks

PR #49, opened 2022-03-25
Avoid duplicating register indices in mask

nsavoire:
Use immediate mode constraint + %c modifier to use enum values in inline asm.


r1viollet (approved):
Thanks for the additional comments ! 
LGTM

PR #50, opened 2022-03-28
Make cmake fail if execute_process fails

nsavoire:
# What does this PR do?

Ensure that cmake fails if execute_process fails.

# Motivation

An error in fetch_elfutils/fetch_libddprof is currently silently ignored.


r1viollet (approved):
LGTM !
I did not know it would not fail. thanks

PR #52, opened 2022-04-04
Fix profiler activation logic

nsavoire:
# What does this PR do?

If generic profiling is disabled (`-d no`) and native profiling is not explicitly set, native profiler should probably be disabled.


r1viollet (approved):
LGTM !
This makes the behaviour more straightforward, thanks

PR #53, opened 2022-04-05
Ignore sample if ABI is set to PERF_SAMPLE_REGS_ABI_NONE

r1viollet:
# What does this PR do?

When investigating issues with tracepoints, we noticed that some samples did not have valid registers.
This change adds a check on the sample's ABI.

# Motivation

Errors noticed in tracepoint mode.

# Additional Notes

N/A

# How to test the change?

The usual test suite will cover this change as it goes through the standard sample function.
I do not expect any impact.

### DIFF
--- a/src/ddprof_worker.cc
+++ b/src/ddprof_worker.cc
@@ -451,7 +451,9 @@ DDRes ddprof_worker_process_event(struct perf_event_header *hdr, int pos,
     case PERF_RECORD_SAMPLE:
       if (wpid->pid) {
         perf_event_sample *sample = hdr2samp(hdr, DEFAULT_SAMPLE_TYPE);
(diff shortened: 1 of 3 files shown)


nsavoire (approved):
LGTM

src/ddprof_worker.cc
```diff
@@ -451,7 +451,9 @@ DDRes ddprof_worker_process_event(struct perf_event_header *hdr, int pos,
     case PERF_RECORD_SAMPLE:
       if (wpid->pid) {
         perf_event_sample *sample = hdr2samp(hdr, DEFAULT_SAMPLE_TYPE);
-        DDRES_CHECK_FWD(ddprof_pr_sample(ctx, sample, pos));
+        if (sample) {
```
sanchda, line 454: Ah, good call.

sanchda (commented):
Good idea overall, thank you!

PR #54, opened 2022-04-06
Fix integer overflow in collatz by using a bigger int type

nsavoire:
# What does this PR do?

Use `int64_t` instead of int as computation type for collatz.

# Motivation

Running `collatz` in SanitizedDebug mode resulted in the following errors:
```
runtime error: signed integer overflow: 827370449 * 3 cannot be represented in type 'int'
```

### DIFF
--- a/bench/collatz/collatz.c
+++ b/bench/collatz/collatz.c
@@ -50,7 +50,7 @@ unsigned long *my_counter = &(unsigned long){0};
 // This is the function body for every expanded function in the X-table
 #define FUNBOD                                                                 \
   {                                                                            \
-    int n = x & 1 ? x * 3 + 1 : x / 2;                                         \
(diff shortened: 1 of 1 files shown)


bench/collatz/collatz.c
```diff
@@ -50,7 +50,7 @@ unsigned long *my_counter = &(unsigned long){0};
 // This is the function body for every expanded function in the X-table
 #define FUNBOD                                                                 \
   {                                                                            \
-    int n = x & 1 ? x * 3 + 1 : x / 2;                                         \
+    int64_t n = x & 1 ? x * 3 + 1 : x / 2;                                         \
```
sanchda, line 53: Ah, good point.

sanchda (approved):
LGTM.  I never anticipated that we'd hit such high collatz numbers, but I guess we do in some of our test harnesses.  Thanks!

PR #57, opened 2022-04-07
Tracepoints

sanchda:
This is still extremely in-progress.  Review at your own peril, there are many things I still need to fix.

# What does this PR do?

Notably, this adds the ability for users to specify arbitrary tracepoints (uprobes, kprobes, ftrace) at the commandline.  It also allows users to harvest values from those tracepoints using RAW events.

It also implements a few internal improvements
 * Decouple the pprof sample type from the watcher, allowing multiple events to contribute to the same sample (alloc profiles)
 * Generalizes the concept of a "count type" to be a dependent profiling type which is updated only when the main type is defined to require it.  This harmonizes with the general semantics of "count" and the requirement that types are well-defined in the Datadog backend.
 * Without providing a user interface for it, allows watchers to send PID/TID as labels.  There is currently no backend support for this, though.
 
# Motivation

In its current form, this PR was inspired by a few different observations
 * the `perf_event_open()` interface that we know and love is a frontend for a far richer collection of events than the initial interface first allowed.  While debugging performance issues in the native profiler, it was clear that a rapid prototyping interface would be useful
 * The general strategy around `perfoptions` was in need of a slight facelift, especially in light of several in-progress development tasks
# Additional Notes

None

# How to test the change?

What's testing?


include/perf_watcher.h
```diff
@@ -0,0 +1,109 @@
...
+} PerfWatcher;
+
+// The Datadog backend only understands pre-configured event types.  Those
+// types are defined here, and then referenced in the watcher
+// The last column is a dependent type which is always aggregated as a count
+// whenever the main type is aggregated.
+#define PROFILE_TYPE_TABLE(X)                                                  \
+  X(NOCOUNT, nocount, nocount, NOCOUNT)                                        \
+  X(TRACEPOINT, tracepoint, events, NOCOUNT)                                   \
+  X(CPU_NANOS, cpu - time, nanoseconds, CPU_SAMPLE)                            \
```
nsavoire, line 48: Why change the name for `cpu-time`, won't this impact the backend ?

src/pprof/ddprof_pprof.cc
```diff
@@ -136,17 +123,16 @@ static void write_line(const ddprof::Symbol &symbol,
...
 
-  int64_t values[MAX_TYPE_WATCHER + 1] = {0};
-  // Counted a single time
-  values[0] = 1;
-  // Add a value to the watcher we are sampling, leave others zeroed
-  values[watcher_idx + 1] = value;
+  int64_t values[DDPROF_PWT_LENGTH] = {0};
+  values[watcher->profile_id] = value;
+  if (is_countable_type(watcher->profile_id))
+    values[watcher->count_id] = 1;
```
nsavoire, line 135: `count_id` is not initialized

src/pprof/ddprof_pprof.cc
```diff
@@ -136,17 +123,16 @@ static void write_line(const ddprof::Symbol &symbol,
...
   const ddprof::MapInfoTable &mapinfo_table = symbol_hdr->_mapinfo_table;
   ddprof_ffi_Profile *profile = pprof->_profile;
 
-  int64_t values[MAX_TYPE_WATCHER + 1] = {0};
-  // Counted a single time
-  values[0] = 1;
-  // Add a value to the watcher we are sampling, leave others zeroed
-  values[watcher_idx + 1] = value;
+  int64_t values[DDPROF_PWT_LENGTH] = {0};
+  values[watcher->profile_id] = value;
```
nsavoire, line 133: There is a off by one error here since `profile_id` starts from 1 (0 is `DDPROF_PWT_NOCOUNT`)

src/pprof/ddprof_pprof.cc
```diff
@@ -34,54 +32,43 @@ std_string_2_slice_c_char(const std::string &str) {
...
-                                         .len = strlen(options->unit)};
-
-    perf_value_type[value_index].type_ = perf_value_slice;
-    perf_value_type[value_index].unit = perf_unit;
+DDRes pprof_create_profile(DDProfPProf *pprof, unsigned type_default,
+    int64_t periodfreq_default) {
+  // Add all the defined profile types (perf_watcher.h), skipping the first
+  struct ddprof_ffi_ValueType perf_value_type[DDPROF_PWT_LENGTH];
+  unsigned num_profiles = 0;
+  for (unsigned i = 0; i < DDPROF_PWT_LENGTH; ++i) {
```
nsavoire, line 46: It would be nice to only create sample types corresponding to enabled watchers

include/perf_watcher.h
```diff
@@ -0,0 +1,109 @@
...
+typedef struct PerfWatcher {
+  const char *desc;
+  uint64_t sample_type;
+  int type;
+  unsigned long config;
+  union {
+    uint64_t sample_period;
+    uint64_t sample_frequency;
+  };
+  int profile_id;
```
nsavoire, line 26: Perhaps `sample_type_id` would be a better name for `profile_id`

include/perf_watcher.h
```diff
@@ -0,0 +1,109 @@
...
+  // perf_event_open configs
+  struct PerfWatcherOptions options;
+  int count_id;
```
nsavoire, line 29: `sample_count_type_id` ?

include/perf_watcher.h
```diff
@@ -0,0 +1,109 @@
...
+  bool send_tid;
+} PerfWatcher;
+
+// The Datadog backend only understands pre-configured event types.  Those
+// types are defined here, and then referenced in the watcher
+// The last column is a dependent type which is always aggregated as a count
+// whenever the main type is aggregated.
+#define PROFILE_TYPE_TABLE(X)                                                  \
+  X(NOCOUNT, nocount, nocount, NOCOUNT)                                        \
+  X(TRACEPOINT, tracepoint, events, NOCOUNT)                                   \
```
nsavoire, line 47: Why use `events` as unit instead of `count` ?

src/ddprof_context_lib.c
```diff
@@ -20,19 +20,16 @@ DDRes ddprof_context_set(DDProfInput *input, DDProfContext *ctx) {
...
-        *(perfoptions_preset(input->watchers[idx_watcher]));
-    if (input->sampling_value[idx_watcher]) // override preset
-      ctx->watchers[ctx->num_watchers].sample_period =
-          input->sampling_value[idx_watcher];
-  }
-  ctx->num_watchers = input->num_watchers;
+  // Shallow copy of the watchers from the input object into the context.  This
+  // enables us to easily manage the lifetime of
+  ctx->num_watchers = 0;
+  for (int *i = &ctx->num_watchers; *i < input->num_watchers; ++*i)
```
nsavoire, line 26: I find this loop hard to read ! Perhaps just using an integer counter and setting `cts->num_watchers = input->num_watchers` would be clearer.

src/pprof/ddprof_pprof.cc
r1viollet, line 46: So for now we add all of the types as columns inside a pprof.
At aggregation we look up what is the matching index considering the index of the watcher.
In a future PR I would also consider the position of the watcher in the ddprof_context.
The index to the value inside the pprof should be more relevant to the values we store inside a pprof ?

src/pprof/ddprof_pprof.cc
```diff
@@ -160,10 +146,47 @@ DDRes pprof_aggregate(const UnwindOutput *uw_output,
...
+
+  // These labels are hardcoded on the Datadog backend
+  static char pid_key[] = "process_id";
+  static char tid_key[] = "thread_id";
+  static char tracepoint_key[] = "tracepoint_type";
+
+  // Add any configured labels.  Note that TID alone has the same cardinalit as
+  // (TID;PID) tuples, so except for symbol table overhead it doesn't matter
+  // much if TID implies PID for clarity.
+  if (watcher->send_pid || watcher->send_tid) {
```
r1viollet, line 168: Thanks for these new labels !

src/pprof/ddprof_pprof.cc
```diff
@@ -34,54 +32,43 @@ std_string_2_slice_c_char(const std::string &str) {
...
-  if (options->freq) {
-    period.type_ = s_cpu_time;
-    period.value = (s_nanos_in_one_sec) / options->sample_frequency;
-  } else {
-    period.type_ = s_cpu_sample;
-    period.value = options->sample_period;
-  }
+                                                     .len = pprof->_nb_values};
+  struct ddprof_ffi_Period period = {
+    .type_ = perf_value_type[type_default],
```
nsavoire, line 69: Probably a off-by-one error here, `perf_value_type` is 0-based, but `type_default` is 1-based.

src/ddprof_worker.cc
```diff
@@ -390,10 +396,14 @@ DDRes ddprof_worker_init(DDProfContext *ctx) {
         ddprof_exporter_new(ctx->worker_ctx.user_tags, ctx->worker_ctx.exp[0]));
     DDRES_CHECK_FWD(
         ddprof_exporter_new(ctx->worker_ctx.user_tags, ctx->worker_ctx.exp[1]));
+
+    // The export profiles non-optionally have the default SampleType of
+    // cpu-time. This should maybe change when we have more than one
+    // customer-facing type.
     DDRES_CHECK_FWD(pprof_create_profile(ctx->worker_ctx.pprof[0],
-                                         ctx->watchers, ctx->num_watchers));
+                                         DDPROF_PWT_CPU_NANOS, 999));
```
nsavoire, line 404: Why `999` ?

nsavoire:
Nice work, I like the mapping watcher -> sample type !

src/ddprof_cmdline.c
```diff
@@ -36,44 +44,208 @@ bool arg_yesno(const char *str, int mode) {
...
+  static const char event_groupname[] = "custom_events";
+  if (watcher->profile_id == DDPROF_PWT_TRACEPOINT) {
+    watcher->tracepoint_name = watcher->desc;
+    watcher->tracepoint_group = event_groupname;
+  }
+  return true;
+}
+
+#ifdef __x86_64__
+int arg2reg[] = {-1, 5, 4, 3, 2, 16, 17}; // 1-indexed, so pad
```
r1viollet, line 83: Are these actual register numbers or the one that we capture through perf events ? I did a commit that might have been wrong. Could you check the following proposal ?
https://github.com/DataDog/ddprof/commit/a3c780d83b262a4880d6ee43c6bd054f886d4df6

sanchda: Well, the -1 is a placeholder to make things fail.  The next registers are numbered positionally in the lookup (they appear to correspond to the linked commit).  The last two, however, are wrong; I counted them incorrectly--good catch!

src/ddprof_worker.cc
sanchda: Just to avoid round numbers to prevent sampling at a multiple of a periodic activity.

src/ddprof_cmdline.c
nsavoire: Why not use enum values from `perf_archmap.h` ?

include/perf_watcher.h
sanchda: Sloppy mistake--I try to impose little structure on the X-table, in terms of strings, and the linter had no way of knowing that the `-` wasn't an operator.  I'll just write string-type columns as string literals.

src/pprof/ddprof_pprof.cc
sanchda: Indeed, I meant to factor it away entirely.  Good catch--thanks!

src/pprof/ddprof_pprof.cc
sanchda: I was originally going to hold off on this, but I'll add it in this PR.  It'll make adding more events frictionless.

src/pprof/ddprof_pprof.cc
sanchda: Good find.  I sloppily added that change without thorough testing and assumed it was a backend issue.  I think this is fixed now that I've changed it so the sample types are dynamically populated and referred to through a permutation vector between the static IDs and the used IDs

include/perf_watcher.h
sanchda: Agree, changed

src/ddprof_worker.cc
sanchda: That's a fair point.  I redid it in the current API, but I'm still working out whether it makes sense for tests.  🤔

include/perf_watcher.h
sanchda: Fixed in latest commit

src/pprof/ddprof_pprof.cc
sanchda: Fixed in latest commit

src/pprof/ddprof_pprof.cc
sanchda: Fixed in latest commit

include/perf_watcher.h
sanchda: Changed to `count_sample_type_id` since it is only used in one place, but that place needs to be informative :)

src/ddprof_context_lib.c
sanchda: Not sure what I was thinking there?  Fixed.

src/pprof/ddprof_pprof.cc
sanchda: Changed this,
1. During setup, all of the watchers are extracted from a DDProfContext and their underlying sample types are checked
2. Only sample types which are used by at least one watcher are populated in the pprof
3. During that process, a permutation vector mapping (static) sample type IDs to (dynamic) pprof IDs is created
4. that permutation vector is used during pprof_aggregate to correctly map both the main sample_type_id as well as the count sample_type_id (if one is defined for that sample type)

There's another way of doing it, which is to add more state to the watcher.  I originally wanted to do that, but opted for the permutation strategy since to my eye indexing one list into a smaller list is simpler.  I guess this is also useful if we ever dynamically add profile types during the lifetime of a single context (then just populate the corresponding entry in the permutation vector), but I find that very unlikely in the next ~6 months...

src/pprof/ddprof_pprof.cc
sanchda: Yeah, I changed this so that the first watcher which is defined with the lowest-numbered sample_type_id is used as the default.  Indices are more transparent now.

src/ddprof_cmdline.c
sanchda: Agree, fixed

include/perf.h
```diff
@@ -116,10 +111,11 @@ typedef struct perf_samplestacku {
...
 
 int perf_event_open(struct perf_event_attr *, pid_t, int, int, unsigned long);
-int perfopen(pid_t pid, const PerfOption *opt, int cpu, bool extras);
+int perfopen(pid_t pid, const PerfWatcher *opt, int cpu, bool extras);
 size_t perf_mmap_size(int buf_size_shift);
 void *perfown_sz(int fd, size_t size_of_buffer);
 void *perfown(int fd, size_t *size);
 int perfdisown(void *region, size_t size);
 long get_page_size(void);
+uint64_t get_sample_type(const PerfWatcher *watcher);
```
nsavoire, line 120: [Minor] I could not find a corresponding definition for this declaration ?

include/pprof/ddprof_pprof.h
```diff
@@ -31,7 +31,7 @@ DDRes pprof_create_profile(DDProfPProf *pprof, const PerfOption *options,
  */
 DDRes pprof_aggregate(const UnwindOutput *uw_output,
                       const SymbolHdr *symbol_hdr, uint64_t value,
-                      int watcher_idx, DDProfPProf *pprof);
+                      const PerfWatcher *watcher, const int*, DDProfPProf *pprof);
```
nsavoire, line 34: [Minor] Give a name to `const int*` argument

src/pprof/ddprof_pprof.cc
nsavoire: I would have added state to PerfWatcher by adding the index in pprof sample array, this would have avoided passing additional arguments to aggregation, but it's a matter of personal taste ;)

src/pprof/ddprof_pprof.cc
```diff
@@ -34,54 +32,77 @@ std_string_2_slice_c_char(const std::string &str) {
...
+  bool active_ids[DDPROF_PWT_LENGTH] = {};
+  struct ddprof_ffi_ValueType perf_value_type[DDPROF_PWT_LENGTH];
+  for (unsigned i = 0; i < num_watchers; ++i) {
+    int this_id = watchers[i].sample_type_id;
+    if (this_id < 0 || this_id == DDPROF_PWT_NOCOUNT || this_id >= DDPROF_PWT_LENGTH) {
+      LG_WRN("Watcher \"%s\" (%d) has invalid sample_type_id %d, ignoring",
+          watchers[i].desc, i, this_id);
+      continue;
+    }
+    active_ids[watchers[i].sample_type_id] = true;
```
nsavoire, line 54: I think you may also need to mark the dependent count sample type as active.

src/pprof/ddprof_pprof.cc
```diff
@@ -34,54 +32,77 @@ std_string_2_slice_c_char(const std::string &str) {
...
+  }
+
+  // Now that we have all of the sample ids, we populate the sample_type_id 
+  // permutation vector. This allows the exporter to take the absolute sample ID 
+  // and convert it into the index of the pprof.
+  memset(ctx->sample_type_pv, 0, sizeof(*ctx->sample_type_pv)*DDPROF_PWT_LENGTH);
```
nsavoire, line 60: `sample_type_pv` should probably be initialized to something different from 0, since 0 is a valid value.

src/pprof/ddprof_pprof.cc
```diff
@@ -34,54 +32,77 @@ std_string_2_slice_c_char(const std::string &str) {
...
-    period.type_ = s_cpu_sample;
-    period.value = options->sample_period;
+                                                     .len = pprof->_nb_values};
+
+  // To populate the defaults, we take the period/frequency of the first watcher
+  // with the first valid sample_type_id.  The checks above ensure we have one
+  // such watcher.
+  int64_t periodfreq_default = 0;
+  for (unsigned i = 0; i < num_watchers; ++i) {
+    if (ctx->sample_type_pv[0] == watchers[i].sample_type_id) {
```
nsavoire, line 97: I think the right test here is `ctx->sample_type_pv[watchers[i].sample_type_id] >= 0` (if `sample_type_id` is initialized to -1).

src/pprof/ddprof_pprof.cc
```diff
@@ -34,54 +32,77 @@ std_string_2_slice_c_char(const std::string &str) {
...
+      periodfreq_default = watchers[i].sample_period;
```
nsavoire, line 98: `watchers[i].sample_period` might be a freq here and would need to be converted tom period.

src/pprof/ddprof_pprof.cc
```diff
@@ -34,54 +32,77 @@ std_string_2_slice_c_char(const std::string &str) {
...
-    perf_value_type[value_index].unit = perf_unit;
+DDRes pprof_create_profile(DDProfPProf *pprof, DDProfContext *ctx) {
+  // Figure out which sample_type_ids are used by active watchers
+  const PerfWatcher *watchers = ctx->watchers;
+  size_t num_watchers = ctx->num_watchers;
+  bool active_ids[DDPROF_PWT_LENGTH] = {};
+  struct ddprof_ffi_ValueType perf_value_type[DDPROF_PWT_LENGTH];
+  for (unsigned i = 0; i < num_watchers; ++i) {
+    int this_id = watchers[i].sample_type_id;
+    if (this_id < 0 || this_id == DDPROF_PWT_NOCOUNT || this_id >= DDPROF_PWT_LENGTH) {
```
nsavoire, line 49: Could this happen?
If this is the case, then you probably need to check also on aggregation side that `sample_id` and `sample_type_pv[sample_id]` are both valid.

src/ddprof_worker.cc
```diff
@@ -390,10 +397,12 @@ DDRes ddprof_worker_init(DDProfContext *ctx) {
         ddprof_exporter_new(ctx->worker_ctx.user_tags, ctx->worker_ctx.exp[0]));
     DDRES_CHECK_FWD(
         ddprof_exporter_new(ctx->worker_ctx.user_tags, ctx->worker_ctx.exp[1]));
-    DDRES_CHECK_FWD(pprof_create_profile(ctx->worker_ctx.pprof[0],
-                                         ctx->watchers, ctx->num_watchers));
-    DDRES_CHECK_FWD(pprof_create_profile(ctx->worker_ctx.pprof[1],
-                                         ctx->watchers, ctx->num_watchers));
+
+    // The export profiles non-optionally have the default SampleType of
```
nsavoire, line 401: This comment might not match the actual implementation ;)

src/ddprof_cmdline.c
```diff
@@ -36,44 +45,210 @@ bool arg_yesno(const char *str, int mode) {
...
+  if (watcher->sample_type_id == DDPROF_PWT_TRACEPOINT) {
+    watcher->tracepoint_name = watcher->desc;
+    watcher->tracepoint_group = event_groupname;
+  }
+  return true;
+}
+
+#define R(x) REGNAME(x)
+#ifdef __x86_64__
+int arg2reg[] = {-1, R(RDI), R(RSI), R(RDX), R(RCX), R(R8), R(R9)};
```
r1viollet, line 85: thanks for adjusting these !

src/pprof/ddprof_pprof.cc
sanchda: I agree!

src/ddprof_worker.cc
sanchda: Oopsies :)

src/pprof/ddprof_pprof.cc
sanchda: I don't _think_ it could happen in the current code, but the boundary check is useful for preventing segfaults and DDPROF_PWT_NOCOUNT (which _should_ be the 0 value always, but there's no way to enforce that in the caller of the component).  Generally speaking, I feel like this is actually overly defensive, but at the same time this check is a boundary check on the pprof rather than the samples.

Not sure.  What do you think?  Overly defensive?  In the meantime, I agree that if we're being defensive, we might as well defend both of our castles and add a compatible check in the aggregation function--good idea.

src/pprof/ddprof_pprof.cc
sanchda: I noticed that too when I wrote this, but decided to take the value literally. Thinking it through more carefully, the current obvious profile type (cpu time) is a frequency, so it'd be somewhat silly not to convert it.  :) (that said, not sure whether that value is really used...)

Anyway, the new code assumes that the target period is in nanoseconds and if a frequency is given, it's in hertz.

src/pprof/ddprof_pprof.cc
sanchda: Yes, this test was totally broken. This is somewhat irrelevant in the new code though.

src/pprof/ddprof_pprof.cc
sanchda: It's a valid enum value, but it it should be guarded against later.  Somewhat of a flavor question.  One preference is to do absolute boundary checks, the other is to check against a sentinel value.  Still not sure, but in the new code the NOCOUNT profile type is checked against explicitly and the scope of the `pv` is limited to this function alone.

src/pprof/ddprof_pprof.cc
sanchda: Yeah, I very much disliked passing the pv around as an argument.  Although to be fair, as you observe, if you're going to pass *arrays* rather than *objects* you necessarily have to pass two arrays (one for the array, one for the index/permutation vector). Anyway, the flexibility you get from using composable objects is totally lost on this interface, which should have very little flexibility in practice. I've backed off from that technique here.

Accordingly, I've changed the PerfWatcher to contain two indices into the pprof ValueType field (one for the value, one for its count), which are populated on pprof creation.  This enables the PerfWatcher to record its original Profiling information as well as index into the pprof correctly.  Whether or not the count is used is still inferred from the original ID, even though the count is correctly populated in the pprof now if it is specified (as other than NOCOUNT).

r1viollet (approved):
Great PR ! 
Thanks !

include/perf_watcher.h
```diff
@@ -0,0 +1,112 @@
...
+#include <stdint.h>
+
+struct PerfWatcherOptions {
+  bool is_kernel;
+  bool is_freq;
+};
+
+typedef struct PerfWatcher {
+  const char *desc;
+  uint64_t sample_type;
```
r1viollet, line 19: [NIT] We could give a name like "perf_sample_type"

include/perf_watcher.h
```diff
@@ -0,0 +1,112 @@
...
+  // tracepoint configuration
+  uint8_t reg;
+  uint8_t trace_off;
+  uint8_t trace_sz;
+  const char *tracepoint_name;
+  const char *tracepoint_group;
+  // Other configs
+  bool send_pid;
+  bool send_tid;
+  int pprof_sample_idx;       // index into the SampleType in the pprof
```
r1viollet, line 38: The comments are helpful thanks !
As mentioned by Nicolas, if we had c++ or an init function, we would benefit from having these init at -1

src/pprof/ddprof_pprof.cc
nsavoire: [Minor] I don't have a definite opinion about this (asserting pre-conditions vs defensively checking them). When modules are internal, I tend to prefer assertions.
I think what matters is to be consistent. For example here if you check for `sample_type_id` validity, then the check should also be done line 97. Or perhaps the function should just return an error, on the first check failure.

nsavoire:
Thanks for the changes

nsavoire (approved):
Great work, it will greatly simplify integration of memory profiling :)
LGTM !

sanchda:
Some good (minor) feedback left in this PR, but I think this will change again soon with the metrics work, so deferring until the interface changes (again!).  Thanks everyone!

PR #58, opened 2022-04-08
Relenv downstream branch parameter

r1viollet:
# What does this PR do?

Add a parameter to define which reliability environment branch should be considered for downstream tests


sanchda (approved):
Looks good!

PR #59, opened 2022-04-11
Fix hyper certification requirements

r1viollet:
# What does this PR do?

Adapt to API changes.
This does not compile as it would require a new publication of libddprof.

libddprof matching PR is here :
https://github.com/DataDog/libddprof/pull/47

# Motivation

Today we can not publish a new version of libddprof

### DIFF
--- a/src/exporter/ddprof_exporter.cc
+++ b/src/exporter/ddprof_exporter.cc
@@ -154,24 +154,30 @@ static DDRes add_single_tag(ddprof_ffi_Vec_tag &tags_exporter,
 static DDRes fill_tags(const UserTags *user_tags,
                        const DDProfExporter *exporter,
                        ddprof_ffi_Vec_tag &tags_exporter) {
+
+  // language is guaranteed to be filled
   DDRES_CHECK_FWD(
       add_single_tag(tags_exporter, "language",
                      std::string_view(exporter->_input.language.ptr,
                                       exporter->_input.language.len)));
 
-  DDRES_CHECK_FWD(
-      add_single_tag(tags_exporter, "language", exporter->_input.environment));
+  if (exporter->_input.environment)
+    DDRES_CHECK_FWD(
+        add_single_tag(tags_exporter, "env", exporter->_input.environment));
 
-  DDRES_CHECK_FWD(add_single_tag(tags_exporter, "version",
-                                 exporter->_input.service_version));
+  if (exporter->_input.service_version)
+    DDRES_CHECK_FWD(add_single_tag(tags_exporter, "version",
+                                   exporter->_input.service_version));
 
-  DDRES_CHECK_FWD(
-      add_single_tag(tags_exporter, "service", exporter->_input.service));
+  if (exporter->_input.service)
+    DDRES_CHECK_FWD(
+        add_single_tag(tags_exporter, "service", exporter->_input.service));
 
-  DDRES_CHECK_FWD(
-      add_single_tag(tags_exporter, "profiler_version",
-                     std::string_view(exporter->_input.profiler_version.ptr,
-                                      exporter->_input.profiler_version.len)));
+  if (exporter->_input.profiler_version.len)
+    DDRES_CHECK_FWD(add_single_tag(
+        tags_exporter, "profiler_version",
+        std::string_view(exporter->_input.profiler_version.ptr,
+                         exporter->_input.profiler_version.len)));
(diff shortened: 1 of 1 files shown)


r1viollet:
Pending a new release from libddprof

sanchda (approved):
I'll read the libddprof PR to familiarize myself with the new interface, but as far as ddprof is concerned this looks good to me.  Thanks!

nsavoire (dismissed):
Thanks for the fixes !

src/exporter/ddprof_exporter.cc
```diff
@@ -28,8 +30,9 @@ extern "C" {
 static const int k_timeout_ms = 10000;
 static const int k_size_api_key = 32;
 
-static ddprof_ffi_CharSlice cpp_string_to_CharSlice(const std::string &str) {
-  return (ddprof_ffi_CharSlice){.ptr = (char *)str.c_str(), .len = str.size()};
+static ddprof_ffi_CharSlice
+cpp_string_to_CharSlice(const std::string_view &str) {
+  return (ddprof_ffi_CharSlice){.ptr = (char *)str.data(), .len = str.size()};
```
nsavoire, line 35: C style cast are not necessary here.

src/exporter/ddprof_exporter.cc
```diff
@@ -28,8 +30,9 @@ extern "C" {
 static const int k_timeout_ms = 10000;
 static const int k_size_api_key = 32;
 
-static ddprof_ffi_CharSlice cpp_string_to_CharSlice(const std::string &str) {
-  return (ddprof_ffi_CharSlice){.ptr = (char *)str.c_str(), .len = str.size()};
+static ddprof_ffi_CharSlice
+cpp_string_to_CharSlice(const std::string_view &str) {
```
nsavoire, line 34: Why not use function overloading instead of using a new function name ?

src/exporter/ddprof_exporter.cc
```diff
@@ -140,49 +143,66 @@ DDRes ddprof_exporter_init(const ExporterInput *exporter_input,
...
-                      std::vector<ddprof_ffi_Tag> &tags_exporter) {
-  tags_exporter.push_back(ddprof_ffi_Tag{
-      .name = char_star_to_CharSlice("language"),
-      .value = string_view_to_CharSlice(exporter->_input.language)});
-
-  if (exporter->_input.environment) {
-    tags_exporter.push_back(ddprof_ffi_Tag{
-        .name = char_star_to_CharSlice("env"),
-        .value = char_star_to_CharSlice(exporter->_input.environment)});
+static DDRes add_single_tag_c(ddprof_ffi_Vec_tag &tags_exporter,
```
nsavoire, line 146: Is this function really necessary since `string_view` has an implicit constructor from`const char*` ?

src/exporter/ddprof_exporter.cc
```diff
@@ -140,49 +143,66 @@ DDRes ddprof_exporter_init(const ExporterInput *exporter_input,
...
+                       const DDProfExporter *exporter,
+                       ddprof_ffi_Vec_tag &tags_exporter) {
 
-  if (exporter->_input.profiler_version.len) {
-    tags_exporter.push_back(ddprof_ffi_Tag{
-        .name = char_star_to_CharSlice("profiler_version"),
-        .value = string_view_to_CharSlice(exporter->_input.profiler_version)});
-  }
+  DDRES_CHECK_FWD(
+      add_single_tag_cpp(tags_exporter, std::string_view("language"),
```
nsavoire, line 179: No need to create `std::string_view`, it will be implicitly constructed from string literal.

src/exporter/ddprof_exporter.cc
r1viollet: Fixed

src/exporter/ddprof_exporter.cc
r1viollet: Fixed

src/exporter/ddprof_exporter.cc
r1viollet: Indeed, Fixed.

src/exporter/ddprof_exporter.cc
r1viollet: Removed them.

nsavoire (approved):
LGTM !

PR #60, opened 2022-04-14
Perform a test for PID namespace compatibility

sanchda:
# What does this PR do?

If a `/host/proc` mountpoint is provided and acts like procfs in the sense that`readlink()` is implemented at the `/host/proc/self` VFS node in a way that returns a _relative_ `/host/proc/<PID>` link, for which `PID` (as an integer) is equal to the current process `getpid()`, then prefer to use `/host/proc`.

If any of this is untrue, then proceed as though `/proc` was available, even if it is not.
# Motivation

Previously, when `/host/proc` was available and determined to be a directory, the profiler would replace many calls to `/proc` with calls to `/host/proc` instead.  This appears to be broken when the profiler is not operating within the root PID namespace, since the PID returned by `perf_event_open()` events will be bound to the PID namespace of the instrumenting process (the profiler), and thus `/host/proc/<PID>` will be invalid (or, at worst, a _different_ process).

Confusingly, `/host/proc` works in many situations.  It appears to work when `/host/proc` is provided _and_ the calling process is in the root PID namespace (actually, this is common in the scenarios where full-host profiling is desired and deployed).  Since `/host/proc` is mounted _from_ that PID namespace, `/proc` and `/host/proc` agree.  This actually makes `/host/proc` irrelevant.

# Additional Notes

None :)

# How to test the change?

The system-probe dockerfile for the agent can repro this.  Will send privately.

### DIFF
--- a/src/dso_hdr.cc
+++ b/src/dso_hdr.cc
@@ -6,6 +6,8 @@
 #include "dso_hdr.hpp"
 
 extern "C" {
+#include <fcntl.h>
+#include <unistd.h>
 #include "ddprof_defs.h"
 #include "logger.h"
 #include "procutils.h"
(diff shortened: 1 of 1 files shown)


src/dso_hdr.cc
```diff
@@ -112,7 +114,12 @@ uint64_t DsoStats::sum_event_metric(DsoEventType dso_event) const {
 DsoHdr::DsoHdr() {
   // keep dso_id 0 as a reserved value
   // Test different places for existence of /proc
-  if (check_file_type("/host/proc", S_IFDIR)) {
+  // A given procfs can only work if its PID namespace is the same as mine.
+  // Fortunately, `/proc/self` will return a symlink to my process ID in the
+  // corresponding namespace, so this is easy to check
+  char pid_str[sizeof("42949672960")] = {};
```
r1viollet, line 120: [minor] I thought max pid was `4194304`. Regardless this is fine, I just wanted to understand the logic

r1viollet (dismissed):
LGTM
Thanks for investigating and fixing this issue !

r1viollet:
Should we merge it ?

src/dso_hdr.cc
sanchda: My idea was that the kernel uses the entire `pid_t`, so I thought of using `2^32`, but now that I review https://github.com/torvalds/linux/blob/master/include/linux/threads.h#L30 I see that the actual _numerical_ limit is `2^30` (nothing larger is ever actually used as a PID and will never appear in traces or in userspace).  I'll change it and add a clarifying comment, since otherwise the number is mysterious.  Thanks!

sanchda: Fixed in latest commit

r1viollet (approved):
LGTM

PR #61, opened 2022-04-15
Allocation profiling

nsavoire:
# What does this PR do?

This PR adds allocation profiling support to ddprof.

# Additional Notes

Allocation profiling is only possible through library mode.

`libdd_profiling.so`redefines standard glibc allocations functions in `malloc_wrapper.cc` (malloc, realloc, calloc, free, reallocarray, posix_memalign, memalign, aligned_alloc, pvalloc, valloc) to intercept the allocation calls and redirect them to the original function . The original function is determined with `dlsym(RTLD_NEXT, ...)`, that should make allocation profiling work even when an allocator replacing glibc allocator (jemalloc, tcmalloc, ...). Initialization is a bit tricky and is done on the first call to any allocation function: at first "original" function pointer for any alloc function points on an init function, which does the init by calling dlsym for all alloc functions and update the "original" function pointer to point on the function returned by dlsym. dlsym might call `calloc` therefore during initialization a temporary calloc implementation that just returns NULL is used (glibc does not seem to mind).

Allocation profiling is activated through a new input event option ("sALLOC") which is defined as a new event type (`kDDPROF_TYPE_CUSTOM`), and takes an optional value (profiling interval in bytes).
A new sample type (`DDPROF_PWT_ALLOC_SPACE`) is associated with this event and also a new count sample type (`DDPROF_PWT_ALLOC_SAMPLE`).

Library and profiler communicate during profiler startup through unix UDP sockets created with socketpair. This allows synchronization during startup, profiler sends information to the library through this channel (profiler PID, allocation profiling parameters, evenfd / memfd for ring buffer creation). Sockets are closed once profiler is attached.

Allocation profiling logic is done in `allocation_tracker.cc` by `track_allocation` / `track_deallocation` functions that are called by allocation functions.
Sampling is done using poisson sampling, samples are sent to the profiler through a ring buffer that has exactly the same format as a perf ring buffer. Samples are handled on profiler side exactly as perf samples.
The current design is to have only one process writing to the ring buffer. If process forks, then allocation profiling is stopped for child.
Currently only allocations are tracked, and there is no "live" allocations measure, but this could be added in the future.

Multithreading concerns: locking is done only when a sample is actually created. Poisson sampling is done per thread.
Locking could be reduced even more by using per-thread a random generator and using a lock free MPSC ring buffer.

Another issue is to determine when profiler should stop, there is no easy way to be notified that the target process has exited except repeatedly calling `kill(...,0)` to check that it still exist. If CPU profiling is done alongside allocation profiling, then perf will signal through a POLLHUP that target process and all it's children have exited.
The current solution is to attach perf with a dummy event when allocation profiling is activated without any other perf event.
This solution could be improved:  avoid ring buffer mapping for dummy events or avoid using perf altogether and keep sockets open in connected mode to detect process exit.


cmake/dd_profiling.version
```diff
@@ -1,4 +1,4 @@
 {
-  global: ddprof_start_profiling; ddprof_stop_profiling;
+  global: ddprof_start_profiling; ddprof_stop_profiling; malloc; realloc; free; calloc; reallocarray; posix_memalign; memalign; aligned_alloc; pvalloc; valloc;
```
r1viollet, line 2: I did not know there were so many possible APIs triggering allocations.

include/ddprof_ffi_utils.hpp
```diff
@@ -0,0 +1,22 @@
...
+
+#include "string_view.h"
+
+#include <string_view>
+
+inline ddprof_ffi_CharSlice to_CharSlice(std::string_view str) {
+  return {.ptr = str.data(), .len = str.size()};
+}
+
+inline ddprof_ffi_CharSlice to_CharSlice(string_view slice) {
```
r1viollet, line 20: Extracting this makes sense.
We will have to finish our transition to c++ at one point 😄

include/perf_watcher.h
```diff
@@ -47,18 +49,24 @@ typedef struct PerfWatcher {
   X(NOCOUNT, "nocount", nocount, NOCOUNT)                                      \
   X(TRACEPOINT, "tracepoint", events, NOCOUNT)                                 \
   X(CPU_NANOS, "cpu-time", nanoseconds, CPU_SAMPLE)                            \
-  X(CPU_SAMPLE, "cpu-sample", count, NOCOUNT)
+  X(CPU_SAMPLE, "cpu-samples", count, NOCOUNT)                                 \
+  X(ALLOC_SAMPLE, "alloc-samples", count, NOCOUNT)                             \
+  X(ALLOC_SPACE, "alloc-space", bytes, ALLOC_SAMPLE)
```
r1viollet, line 54: this works well with David's work

include/perf_watcher.h
```diff
@@ -47,18 +49,24 @@ typedef struct PerfWatcher {
...
 typedef enum DDPROF_SAMPLE_TYPES {
   PROFILE_TYPE_TABLE(X_ENUM) DDPROF_PWT_LENGTH,
 } DDPROF_SAMPLE_TYPES;
 #undef X_ENUM
 
-enum DDPROF_PERFOPEN_CONFIGS {
-  IS_FREQ = 1 << 0,
-  IS_KERNEL = 1 << 1,
-};
+enum DDProfTypeId { kDDPROF_TYPE_CUSTOM = PERF_TYPE_MAX + 100 };
```
r1viollet, line 62: It should be easier and easier to add new type of events, nice.
A comment around this definition would be nice, explaining that we are defining our own types different from the perf event API.

include/perf_watcher.h
```diff
@@ -47,18 +49,24 @@ typedef struct PerfWatcher {
...
+
+enum DDProfCustomCountId { kDDPROF_COUNT_ALLOCATIONS = 0 };
+
+#define IS_FREQ                                                                \
```
r1viollet, line 66: so this is a modifier for the perf event setup ? (I'll figure it out once I understand the pre-processor magic)

include/perf_watcher.h
```diff
@@ -68,25 +76,26 @@ enum DDPROF_PERFOPEN_CONFIGS {
 // clang-format off
 //  short    desc              type config                   period/freq  profile type   addtl. configs
```
r1viollet, line 77: we can realign the comments

include/perf_watcher.h
```diff
@@ -12,13 +12,15 @@
 struct PerfWatcherOptions {
   bool is_kernel;
   bool is_freq;
+  uint8_t nb_frames_to_skip; // number of frames to skip during unwinding
```
r1viollet, line 15: Nice, So the number of frames to skip depends on the library mechanism we implement for a given type of event. This means we should not have a different number of frames between malloc and realloc for instance.
(not saying anything, just trying to make sure I understand)

include/perf_watcher.h
```diff
@@ -12,13 +12,15 @@
...
 };
 
 typedef struct PerfWatcher {
   const char *desc;
-  uint64_t sample_type;
-  int type;
-  unsigned long config;
+  uint64_t sample_type; // perf sample type: specifies values included in sample
+  int type; // perf event type (software / hardware / tracepoint / ... or custom
```
r1viollet, line 21: thank you for comments

include/pevent.h
```diff
@@ -12,9 +12,13 @@
 #define MAX_NB_WATCHERS 450
 
 typedef struct PEvent {
-  int pos;       // Index into the sample
-  int fd;        // Underlying perf event FD
-  RingBuffer rb; // metadata and buffers for processing perf ringbuffer
+  int pos; // Index into the sample
+  int fd; // Underlying perf event FD for perf_events, otherwise an eventfd that
```
r1viollet, line 16: Yes, you mentioned we could not do the exact same thing as perf events.

src/exporter/ddprof_exporter.cc
```diff
@@ -52,18 +45,13 @@ static char *alloc_url_agent(const char *protocol, const char *host,
...
-  strftime(time_end, sizeof time_end, "%Y-%m-%dT%H:%M:%SZ", tm_end);
+  char time_start[128] = {};
+  tm tm_storage;
+  tm *tm_start = gmtime_r(&start.seconds, &tm_storage);
+  strftime(time_start, sizeof time_start, "%Y%m%dT%H%M%SZ", tm_start);
 
   char filename[400];
-  snprintf(filename, 400, "%s/ddprof_%s_%s.pprof", dbg_folder, time_start,
-           time_end);
+  snprintf(filename, 400, "%s/ddprof_%s.pprof", dbg_folder, time_start);
```
r1viollet, line 54: The end time is not so relevant indeed.

test/CMakeLists.txt
```diff
@@ -297,3 +297,8 @@ target_link_libraries(savecontext-bench benchmark::benchmark benchmark::benchmar
 target_include_directories(savecontext-bench PRIVATE ${DDPROF_INCLUDE_LIST})
 # clang-tidy does like google-benchmark macros
 disable_clangtidy(savecontext-bench)
+
+
+# add_executable(ringbuffer-ut ringbuffer-ut.cc ../src/ipc.cc)
```
r1viollet, line 302: [minor] Some test code remains around here

include/perf_watcher.h
nsavoire: I am in the process of changing this since it might depend on optimization level. The idea will be to skip all the frames belonging to same mapping.

include/perf_watcher.h
nsavoire: These are additional options for the event setup, I changed the implementation to be able to add my own option.

include/perf_watcher.h
r1viollet: That is a good idea. Though if someone compiles with a static library, this no longer works ? I think the current logic is maybe OK.

include/perf_watcher.h
nsavoire: Yeah, some of them are deprecated, but I included them for the sake of completeness.

test/CMakeLists.txt
nsavoire: I am working on adding tests / benchmarks, this is part of it ;)

include/perf_watcher.h
nsavoire: Done

include/perf_watcher.h
nsavoire: Done

include/perf_watcher.h
nsavoire: Indeed, this won't work with static library.
No even sure than allocation profiling will work with static lib.
Need to think a bit more on this !

r1viollet (approved):
Impressive PR. Fixing the number of frames to ignore can be done later.
LGTM !

include/perf_watcher.h
nsavoire: Tested allocation profiling with static lib and it worked (did no tried by a statically linked jemmalloc/tcmalloc).
I will bring back a fixed number of frames to skip that depends on compilation mode (release/LTO)...

src/lib/allocation_tracker.cc
```diff
@@ -0,0 +1,365 @@
...
+  inline size_t available_size() const { return _available_size; }
+
+  void write(Buffer buf) {
+    assert(buf.size() <= _available_size);
+
+    uint64_t head_linear = _head & _rb.mask;
+    char *dest = const_cast<char *>(_rb.start) + head_linear;
+
+    // check if we need to wrap around end of buffer
+    // \fixme remove memcpy by writing directly into the ring buffer ?
```
sanchda, line 157: What would be the advantage of suppressing `memcpy()`?

sanchda (dismissed):
This is quite a significant PR with many nontrivial contributions to the native profiler.  I tried to do my due diligence to understand potential system failure modes in the IPC loops, but I haven't found anything noteworthy.  I have a few more experiments to run, but I don't think they should block this PR.

LGTM

include/perf_watcher.h
sanchda: The idea is to skip frames particular to the profiling library, right?  I have a sense that indicating a static number of frames to skip is difficult, but I wonder whether we can emit a special frame (or detect our own entrypoints, which we could mark somehow) and we would use it to drop all stored unwinding state and basically "start over" until after that frame.  Is that idea too naive?

r1viollet: Indeed, that is a great idea. We could have profiling conventions for frames. Though I'm not too worried about this. I'm sure we can fix this later.

nsavoire: I was reluctant to base the logic on the name of the frames (what if symbols are stripped?).

nsavoire: Relying on the entry points would work (malloc/realloc/calloc/...) but is a bit of a pain (there a 10 different entry points).

src/lib/allocation_tracker.cc
nsavoire: I see 2 advantages:
* do less work: directly write to ringbuffer
* remove the need for static variable `s_event`

PR #63, opened 2022-05-04
Invert pid/tid_send to suppression

sanchda:
This globally enabled sending PID and TID to the backend.  This can be
changed later once we determine whether this needs to be a
user-configurable setting.

### DIFF
--- a/src/pprof/ddprof_pprof.cc
+++ b/src/pprof/ddprof_pprof.cc
@@ -202,22 +202,18 @@ DDRes pprof_aggregate(const UnwindOutput *uw_output,
   ddprof_ffi_Label labels[PPROF_MAX_LABELS] = {};
   size_t labels_num = 0;
 
-  // The max PID on Linux is approximately 2^22 (~4e6), which also limits TID
-  // cppcheck-suppress variableScope
-  char pid_str[7] = {};
-  // cppcheck-suppress variableScope
-  char tid_str[7] = {};
-
   // Add any configured labels.  Note that TID alone has the same cardinality as
   // (TID;PID) tuples, so except for symbol table overhead it doesn't matter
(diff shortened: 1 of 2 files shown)


r1viollet (approved):
LGTM
This makes a lot of sense. Thank you !

src/pprof/ddprof_pprof.cc
```diff
@@ -202,22 +202,18 @@ DDRes pprof_aggregate(const UnwindOutput *uw_output,
...
-  char pid_str[7] = {};
-  // cppcheck-suppress variableScope
-  char tid_str[7] = {};
-
   // Add any configured labels.  Note that TID alone has the same cardinality as
   // (TID;PID) tuples, so except for symbol table overhead it doesn't matter
   // much if TID implies PID for clarity.
-  if (watcher->send_pid || watcher->send_tid) {
+  if (!watcher->suppress_pid || !watcher->suppress_tid) {
+    char pid_str[sizeof("536870912")] = {}; // reserve space up to 2^29 base-10
```
nsavoire, line 209: Not sure that reducing scope of `pid_str` / `tid_str` is correct: `labels` will store a pointer on a variable that has gone out-of-scope.

r1viollet: Wow, I missed that, thanks ! Indeed, cppcheck is not detecting this correctly. To be fair the design pattern is weird.

r1viollet: Not urgent, though trying to see how we can avoid this issue. Perhaps we should adapt this in some way. Having a general storage object ?

sanchda: Ah, good call--I followed the cppcheck nag without thinking!

PR #64, opened 2022-05-06
Fix static initialization order in allocation profiler

nsavoire:
# What does this PR do?

Compiler was generating a constructor for TrackerStaticState that is called before main and could be called after constructor of g_autostart that is starting allocation profiling.
This commit moves TrackerStaticState inside AllocationTracker as a non-static member. This adds an indirection to access this state, but overhead should be very low.
Also moves all free functions inside AllocationTracker, and exposes only the required functions.
Avoid writing events twice to ring buffer.


r1viollet (approved):
LGTM

PR #65, opened 2022-05-09
Avoid calling poll repeatedly

nsavoire:
# What does this PR do?

* On consumer side, make sure to read from eventfd to clear POLLIN state.
* On consumer side, do not call poll if there are still samples left in the ring buffer
* On producer side, after commit, only signal eventfd if consumer has caught to position of head before commit

# Motivation

Avoid excessive CPU usage.

### DIFF
--- a/include/ringbuffer_utils.hpp
+++ b/include/ringbuffer_utils.hpp
@@ -17,47 +17,59 @@ class RingBufferWriter {
 public:
   explicit RingBufferWriter(RingBuffer &rb) : _rb(rb) {
     _tail = __atomic_load_n(&_rb.region->data_tail, __ATOMIC_ACQUIRE);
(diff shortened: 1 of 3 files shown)


include/ringbuffer_utils.hpp
```diff
@@ -17,47 +17,59 @@ class RingBufferWriter {
...
 
     memcpy(dest, buf.data(), buf.size());
-
-    _available_size -= buf.size();
     _head += buf.size();
   }
 
+  // return true if notification to consumer is necesssary
+  // Notification is necessary only if consumer has caught up with producer
+  // (meaning tail afer commit is at or after head before commit)
```
r1viollet, line 59: smart !

r1viollet (approved):
LGTM

PR #66, opened 2022-05-09
Allow overriding exe used by libdd_profiling

nsavoire:
If set, path stored in DD_PROFILING_NATIVE_DDPROF_EXE env variable is used instead of embedded binary.

### DIFF
--- a/src/lib/dd_profiling.cc
+++ b/src/lib/dd_profiling.cc
@@ -33,6 +33,8 @@ static constexpr const char *k_profiler_active_env_variable =
     "DD_PROFILING_NATIVE_LIBRARY_ACTIVE";
 static constexpr const char *k_profiler_auto_start_env_variable =
     "DD_PROFILING_NATIVE_AUTOSTART";
+static constexpr const char *k_profiler_ddprof_exe_env_variable =
+    "DD_PROFILING_NATIVE_DDPROF_EXE";
 
 // address of embedded ddprof executable
 extern const char _binary_ddprof_start[]; // NOLINT cert-dcl51-cpp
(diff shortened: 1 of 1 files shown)


src/lib/dd_profiling.cc
```diff
@@ -131,11 +134,13 @@ static int exec_ddprof(pid_t target_pid, pid_t parent_pid, int sock_fd) {
 
   kill(parent_pid, SIGTERM);
 
-  if (fexecve(fd, argv, environ) == -1) {
-    return -1;
+  if (const char *ddprof_exe = getenv(k_profiler_ddprof_exe_env_variable);
+      ddprof_exe) {
+    execve(ddprof_exe, argv, environ);
```
r1viollet, line 139: Do we still want to do the write and create an fd operation ?

r1viollet:
Could you clarify the motivation of such an operation ? I'm guessing this is to iterate faster and just sync the exe ?

src/lib/dd_profiling.cc
nsavoire: I will move the memfd inside the if branch.

nsavoire:
> Could you clarify the motivation of such an operation ? I'm guessing this is to iterate faster and just sync the exe ?

The goal is to allow debugging ddprof in library mode (when using embedded exe, mapped with memfd_create, symbol maps are not available).

r1viollet (approved):
Thanks for clarifying the purpose of the PR. 
I'm guessing this makes it hard to profile the profiler without using the variable.
LGTM

nsavoire:
> Thanks for clarifying the purpose of the PR. I'm guessing this makes it hard to profile the profiler without using the variable. LGTM

Yes, these are parts of the downsides to using memfd instead of an actual file. Maybe we could try first to write a temporary file and if it fails fallback to memfd ?

r1viollet:
> > Thanks for clarifying the purpose of the PR. I'm guessing this makes it hard to profile the profiler without using the variable. LGTM
> 
> Yes, these are parts of the downsides to using memfd instead of an actual file. Maybe we could try first to write a temporary file and if it fails fallback to memfd ?

Good idea

PR #67, opened 2022-05-11
Fix ring buffer bug and malloc wrapper init deadlock

nsavoire:
# What does this PR do?

*  Fix computation for ringbuffer available size
 
    Available size should return one less than the actual available space in the ring buffer,
    because a completely full ringbuffer is indistinguishable from an empty one.
* Fix deadlock when allocation tracker is activated before malloc wrappers

    If allocation tracker is activated before first allocation (and therefore before malloc wrappers are initalized), then upon first allocation,
    AllocationTracker::track_allocation would be called during check_init, triggering an allocation that would reenter check_init and cause a
    deadlock.
    Fix is to call malloc before activation of AllocationTracker to force init of malloc wrappers if not done yet.
* Add unit test for ringbuffer
* Extract utils for custom ring buffer into their own file

### DIFF
--- a/src/ddprof.c
+++ b/src/ddprof.c
@@ -58,6 +58,8 @@ DDRes ddprof_setup(DDProfContext *ctx) {
     LG_ERR("Error when printing capabilities, continuing...");
   }
 
+  // Do not mmap events yet because mmap'ings from perf fds are lost after
+  // fork
   DDRES_CHECK_FWD(
       pevent_open(ctx, ctx->params.pid, ctx->params.num_cpu, pevent_hdr));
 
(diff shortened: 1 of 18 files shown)


src/ddprof.c
```diff
@@ -58,6 +58,8 @@ DDRes ddprof_setup(DDProfContext *ctx) {
     LG_ERR("Error when printing capabilities, continuing...");
   }
 
+  // Do not mmap events yet because mmap'ings from perf fds are lost after
```
r1viollet, line 61: thanks for this reminder !

r1viollet (approved):
LGTM
Congrats on finding that bug !

PR #70, opened 2022-05-13
Formalize tracepoint parser

sanchda:
# What does this PR do?

Raw events weren't enabled in the ringbuffer during the tracepoint PR.  This was an omission.

# Motivation

1. We already had the syntax for specifying the offset into a raw event to use
2. raw events are needed to surface contextual information out of tracepoints
3. when raw events are specified, but we lack support to process them, there's no way to maintain ringbuffer state (the events are variable-length) so we segfault

# Additional Notes

# How to test the change?

Test added to the perf ringbuffer ut


src/ddprof_worker.cc
```diff
@@ -168,15 +168,24 @@ DDRes ddprof_pr_sample(DDProfContext *ctx, perf_event_sample *sample, int pos) {
   unsigned long this_ticks_unwind = __rdtsc();
   DDRes res = unwindstate__unwind(us);
 
+  // Usually we want to send the sample_val, but sometimes we need to process
+  // the event to get the desired value
+  PerfWatcher *watcher = &ctx->watchers[pos];
+  uint64_t sample_val = sample->period;
```
r1viollet, line 174: minor, this could belong to the perf_ringbuffer file ?

src/ddprof_worker.cc
```diff
@@ -168,15 +168,24 @@ DDRes ddprof_pr_sample(DDProfContext *ctx, perf_event_sample *sample, int pos) {
...
+  if (PERF_SAMPLE_RAW & watcher->sample_type) {
+    uint64_t raw_offset = watcher->trace_off;
```
r1viollet, line 176: This is pretty neat in terms of design. The watcher knows where to find the value and keeps an offset. 
I guess in some cases, it could make sense to have fixed values, though we could adapt that.
I wonder if the sample->period could be "merged" some way with this design. I did not give it any thought.

src/perf_ringbuffer.c
```diff
@@ -245,12 +250,19 @@ perf_event_sample *hdr2samp(const struct perf_event_header *hdr,
...
+    sample.size_raw = ((flipper *)buf)->half[0];
+    sample.data_raw = sample.size_raw ? &((flipper *)buf)->half[1] : NULL;
+    buf += 1 + (sample.size_raw / sizeof(*buf)); // Advance + align
+  }
   if (PERF_SAMPLE_BRANCH_STACK & mask) {}
   if (PERF_SAMPLE_REGS_USER & mask) {
     sample.abi = *buf++;
-    // In case regs are not available, ignore this sample
-    if (sample.abi == PERF_SAMPLE_REGS_ABI_NONE) {
+    // Since some of the previous events are variable-length, ABI gives us
```
r1viollet, line 262: Interesting !
Instrumentation telemetry would be nice here ! Perhaps we should leave markers of where we want to raise these events.

r1viollet (dismissed):
Thanks for the upgrades !
A few minor comments. LGTM

src/perf_ringbuffer.c
```diff
@@ -245,12 +250,19 @@ perf_event_sample *hdr2samp(const struct perf_event_header *hdr,
...
+    // an opportunity to sanity-check the buffer (only 2 legal values!)
+    if (sample.abi != PERF_SAMPLE_REGS_ABI_32 &&
```
nsavoire, line 264: Not sure to follow the reasoning here, isn't `PERF_SAMPLE_REGS_ABI_NONE` a legal value ?

r1viollet: `PERF_SAMPLE_REGS_ABI_NONE` is legal, though we will not do anything with it. 
The reasoning is to make the check less permissive if somehow we miss-align reads, we won't create a bad sample / crash in the following lines.

nsavoire (approved):

src/perf_ringbuffer.c
nsavoire: [Minor] The comment is a bit misleading then
It could be interesting to separate `PERF_SAMPLE_REGS_ABI_NONE` case (we currently return NULL, but from the ring buffer perspective, sample is valid) and other invalid values.

nsavoire (dismissed):
LGTM, thanks for the fix !

r1viollet:
Should we consider this one in the upcoming release ?

src/ddprof_worker.cc
sanchda: I've thought about this several times and I don't know if I agree.  Right now, the ringbuffer implementation is "pure" in the sense that it returns to the user exactly what is in the buffer.  If we push down the business logic here (e.g., overwriting the sample value), then we've added logic which doesn't feel to me like a natural part of ringbuffer deserialization.

On the other hand, the worker is already responsible for consolidating a variety of user inputs and application states (such as the sample events).  This area is probably overdue for a refactor.

What are your thoughts?  Why/what did you think could go in the ringbuffer implementation?

CMakeLists.txt
```diff
@@ -125,7 +149,7 @@ aux_source_directory(src/exe EXE_SRC)
 # Define all sources
 set(DDPROF_GLOBAL_SRC ${COMMON_SRC} ${PPROF_SRC} ${EXPORTER_SRC} ${EXE_SRC})
 
-set(DDPROF_LIBRARY_LIST llvm-demangle ${ELFUTILS_LIBRARIES} Threads::Threads)
+set(DDPROF_LIBRARY_LIST llvm-demangle ${ELFUTILS_LIBRARIES} Threads::Threads DDProf::Parser)
```
r1viollet, line 152: It is nice that we isolate this in it's own library
We should follow this pattern more. This way when we add new code, we don't recompile it many times.

include/perf_archmap.hpp
```diff
@@ -89,3 +89,15 @@ enum PERF_ARCHMAP_ARM {
...
 #endif
+
+unsigned int dwarf_to_perf_regno(unsigned int i);
+unsigned int param_to_perf_regno(unsigned int param_no);
+
+// C API for use by the tracepoint parser
+#ifdef __cplusplus
+extern "C" {
+#endif
+unsigned int param_to_regno_c(unsigned int param_no);
```
r1viollet, line 100: good helper function !

src/ddprof_worker.cc
r1viollet: The suggestion was indeed not great. Ringbuffer should not use watcher knowledge.

We should indeed avoid adding more responsibility to the worker.
I would suggest a simple refactor to have a function computing the value based on watcher + raw memory / registers. 

Minor: Should we not protect the access to the raw memory ?

src/ddprof_worker.cc
```diff
@@ -267,14 +285,14 @@ DDRes ddprof_pr_sample(DDProfContext *ctx, perf_event_sample *sample,
       ddprof_stats_add(STATS_UNWIND_AVG_TIME, unwind_ticks - ticks0, NULL));
 
   // Aggregate if unwinding went well (todo : fatal error propagation)
-  if (!IsDDResFatal(res)) {
+  if (!IsDDResFatal(res) &&
+      (watcher->output_mode & kPerfWatcherMode_callgraph)) {
```
r1viollet, line 289: Interesting customisation

src/dwfl_thread_callbacks.cc
```diff
@@ -19,70 +20,21 @@ pid_t next_thread(Dwfl *dwfl, void *arg, void **thread_argp) {
...
-  regs[n++] = us->initial_regs.regs[REGNAME(X24)];
-  regs[n++] = us->initial_regs.regs[REGNAME(X25)];
-  regs[n++] = us->initial_regs.regs[REGNAME(X26)];
-  regs[n++] = us->initial_regs.regs[REGNAME(X27)];
-  regs[n++] = us->initial_regs.regs[REGNAME(X28)];
-  regs[n++] = us->initial_regs.regs[REGNAME(FP)];
-  regs[n++] = us->initial_regs.regs[REGNAME(LR)];
-  regs[n++] = us->initial_regs.regs[REGNAME(SP)];
+  unsigned int num_regs = 0;
+  unsigned int regs_idx = dwarf_to_perf_regno(num_regs);
```
r1viollet, line 26: Thanks for factorising.

Could we improve the naming ? I always take 10 minutes to remember how things work. Not sure if something could help or it is just me.

util/event_parser.l
```diff
@@ -0,0 +1,85 @@
+%option noyywrap
```
r1viollet, line 1: I would not mind a small review of this together. I don't do Lex and Yacc on a daily basis 😅

test/ddprofcmdline-ut.cc
```diff
@@ -47,11 +48,224 @@ TEST(CmdLineTst, NullPatterns) {
 TEST(CmdLineTst, FirstEventHit) {
   char const *str = "hCPU";
   PerfWatcher watcher = {};
-  ASSERT_TRUE(watcher_from_event(str, &watcher));
+  ASSERT_TRUE(watcher_from_str(str, &watcher));
   ASSERT_EQ(watcher.type, PERF_TYPE_HARDWARE);
   ASSERT_EQ(watcher.type, PERF_COUNT_HW_CPU_CYCLES);
 }
 
+TEST(CmdLineTst, ParserKeyPatterns) {
```
r1viollet, line 56: very nice !

util/event_parser.l
r1viollet: @sanchda could we have a small doc to describe the syntax ?

util/build.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
```
r1viollet, line 1: I think we forgot about this file ?

r1viollet:
I think this can go in the next release (also it makes my life easier if we merge it). Can we do a small session on this PR ?

CMakeLists.txt
```diff
@@ -104,6 +107,31 @@ include(Findlibcap)
 # llvm demangler
 add_subdirectory(third_party/llvm)
 
+# ---- Event Parser ----
```
nsavoire, line 110: This should probably be moved to its own `CMakeLists.txt` in`util` directory

nsavoire (approved):
Nice work !
I like the grammar idea.

src/perf_archmap.cc
```diff
@@ -0,0 +1,55 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0. This product includes software
+// developed at Datadog (https://www.datadoghq.com/). Copyright 2021-Present
+// Datadog, Inc.
+
+#include "perf_archmap.hpp"
+
+// F[i] = y_i; where i is the DWARF regno and y_i is the Linux perf regno
+unsigned int dwarf_to_perf_regno(unsigned int i) {
+  static unsigned int lookup[] = {
```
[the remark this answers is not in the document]
r1viollet: Minor: This function can be constexpr ?

r1viollet (approved):
LGTM 🥁

sanchda:
I tried to update the build, let's see if this merge works!

PR #71, opened 2022-05-17
Add support for allocation profiling in wrapper mode

nsavoire:
# What does this PR do?
Add support for allocation profiling in wrapper mode
-  In wrapper mode, when allocation profiling is requested, inject with LD_PRELOAD a shared library into target process
- ddprof first looks for libdd_profiling.so in its directory and ../lib, if not found, ddprof uses a library embedded and creates a temp file
- ddprof creates a socket pair and sets DD_PROFILING_NATIVE_LIB_SOCKET env variable with socket fd
 - library when it detects that DD_PROFILING_NATIVE_LIB_SOCKET is set, connects to existing ddprof instead of spawning a new profiler

Other changes:
- Make sure that style-check.sh checks all the source files
- Move constants to their own file (`constants.hpp`)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -146,9 +146,40 @@ endif()
 list(APPEND DDPROF_LIBRARY_LIST libcap)
 list(APPEND DDPROF_INCLUDE_LIST ${LIBCAP_INCLUDE_DIR})
 
+set(DD_PROFILING_SOURCES src/lib/dd_profiling.cc src/daemonize.cc src/ddprof_cmdline.c src/logger.c src/logger_setup.cc 
+                         src/ipc.cc src/pevent_lib.cc src/perf.cc src/perf_ringbuffer.c src/perf_watcher.cc src/ringbuffer_utils.cc
+                         src/user_override.c src/ddres_list.c src/savecontext.cc src/lib/malloc_wrapper.cc src/lib/allocation_tracker.cc)
+
+add_library(dd_profiling-embedded SHARED ${DD_PROFILING_SOURCES})
+target_include_directories(dd_profiling-embedded PUBLIC ${CMAKE_SOURCE_DIR}/include/lib ${CMAKE_SOURCE_DIR}/include ${CMAKE_SOURCE_DIR}/third_party)
+set_target_properties(dd_profiling-embedded PROPERTIES
+   PUBLIC_HEADER "${CMAKE_SOURCE_DIR}/include/lib/dd_profiling.h")
+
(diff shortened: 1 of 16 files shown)


tools/style-check.sh
```diff
@@ -3,13 +3,9 @@
...
-### Set directory names
-CURRENTDIR=$PWD
-SCRIPTPATH=$(readlink -f "$0")
-SCRIPTDIR=$(dirname $SCRIPTPATH)
-cd $SCRIPTDIR/../
-TOP_LVL_DIR=$PWD
-cd $CURRENTDIR
+set -euo pipefail
+
+cd "$(git rev-parse --show-toplevel)"
```
r1viollet, line 8: minor This does not work well in some cases (worktrees / copies of the directory)

nsavoire: Good point, I will change that

nsavoire: It should work with worktrees though

nsavoire: Done

include/constants.hpp
```diff
@@ -0,0 +1,27 @@
+#pragma once
+
```
r1viollet, line 2: I think we have a defs file, that could be somewhat redundant. We can split type definition from constants maybe.

nsavoire: Indeed I forgot about `ddprof_defs.h`, I could move the constants there once migrated to C++.

CMakeLists.txt
```diff
@@ -146,9 +146,40 @@ endif()
 list(APPEND DDPROF_LIBRARY_LIST libcap)
 list(APPEND DDPROF_INCLUDE_LIST ${LIBCAP_INCLUDE_DIR})
 
+set(DD_PROFILING_SOURCES src/lib/dd_profiling.cc src/daemonize.cc src/ddprof_cmdline.c src/logger.c src/logger_setup.cc
```
r1viollet, line 149: thanks for factorising this

include/daemonize.hpp
```diff
@@ -0,0 +1,24 @@
...
+// Datadog, Inc.
+
+#pragma once
+
+#include <functional>
+#include <sys/types.h>
+
+namespace ddprof {
+
+struct DaemonizeResult {
```
r1viollet, line 13: This API looks nice considering what it was in the beginning 👍 . I had trouble figuring out how we could design an interface around this

r1viollet (approved):
LGTM ! 
An impressive PR.

PR #72, opened 2022-05-17
Add input option to set debug pprof prefix

nsavoire:
# What does this PR do?

This replaces DDPROF_PPROFS_FOLDER and allows to control prefix of generated pprof files.


r1viollet (approved):
LGTM
This standardises the debug folder option

PR #74, opened 2022-05-24
Make ddprof fail on bad input

nsavoire:
# What does this PR do?

* Make ddprof fail on bad input event
* Accept to create profile with no sample type (to make `ddprof -e sDUM` work)

### DIFF
--- a/src/pprof/ddprof_pprof.cc
+++ b/src/pprof/ddprof_pprof.cc
@@ -31,15 +31,16 @@ DDRes pprof_create_profile(DDProfPProf *pprof, DDProfContext *ctx) {
   // We also record the watcher with the lowest valid sample_type id, since that
   // will serve as the default for the pprof
   bool active_ids[DDPROF_PWT_LENGTH] = {};
-  PerfWatcher *default_watcher = &watchers[0];
+  PerfWatcher *default_watcher = watchers;
   for (unsigned i = 0; i < num_watchers; ++i) {
     int this_id = watchers[i].sample_type_id;
     int count_id = sample_type_id_to_count_sample_type_id(this_id);
     if (this_id < 0 || this_id == DDPROF_PWT_NOCOUNT ||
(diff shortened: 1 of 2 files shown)


src/pprof/ddprof_pprof.cc
```diff
@@ -91,38 +91,37 @@ DDRes pprof_create_profile(DDProfPProf *pprof, DDProfContext *ctx) {
...
   if (!pprof->_profile) {
-    DDRES_RETURN_ERROR_LOG(DD_WHAT_PPROF, "Unable to allocate profiles");
+    DDRES_RETURN_ERROR_LOG(DD_WHAT_PPROF, "Unable to create profile");
   }
   return ddres_init();
 }
 
 DDRes pprof_free_profile(DDProfPProf *pprof) {
-  ddprof_ffi_Profile_free(pprof->_profile);
+  if (pprof->_profile) {
```
sanchda, line 122: Nit: does this check actually improve readability?  `free(NULL)` is valid as per POSIX (https://pubs.opengroup.org/onlinepubs/009604499/functions/free.html)

sanchda (approved):
LGTM!  I left a stylistic comment, but I'm satisfied either way.

src/pprof/ddprof_pprof.cc
nsavoire: Well Rust has a different opinion on the matter, `ddprof_ffi_Profile_free` crashes if passed a null pointer.
I had a crash when `pprof_create_profile` was failing and `pprof_free_profile` was subsequently trying to free a null profile (just try using `-e sDUM` alone on main).

r1viollet: Interesting !

r1viollet (approved):
LGTM !
Thanks for saving time for the next user who will get things wrong (probably me)

PR #75, opened 2022-05-30
libdatadog 0.7.0 - Use the latest release of the shared library

r1viollet:
# What does this PR do?

- Repository was moved to libdatadog (instead of libddprof)
- APIs now allow specifying time (instead of taking the current time)

# Motivation

- Test the new changes.
- Ensure we point to the new libdatadog repository

# Additional Notes

Some more renaming could help, thoough libddprof still exists within libdatadog. Our downstream naming should be fixed first.

# How to test the change?

Relenv and our usual tests should ensure we upload profiles.


nsavoire (approved):
LGTM !

PR #77, opened 2022-05-31
Fix worker restart issue

nsavoire:
# What does this PR do?

Worker was not restarting because `persistent_worker_state->restart_worker` was unconditionally set to false upon worker exit.


r1viollet (approved):
LGTM
Thanks for investigating

PR #80, opened 2022-06-03
[PROTO] User override when reading proc maps

r1viollet:
# What does this PR do?

When accessing proc maps and receiving a failure opening the file, attempt to switch to the user that owns this file.
This was raised by an internal user. The bug occurs when a gitlab runner was using a different user to run the app.

# Motivation

A User reported a bug

# Additional Notes

We were previously hitting this as we were catching the important mappings at startup.

# How to test the change?

Find an app that switches user


r1viollet:
@nsavoire did a better version of this. Closing.

PR #82, opened 2022-06-10
Add profile presets

nsavoire:
# What does this PR do?

Add preset profile configurations, at the moment: `default`, `cpu_only`, `alloc_only`
`default` preset includes cpu + allocations in lib mode and wrapper mode, and cpu only in global / pid mode.
`default` preset is used when no preset is given and no event are passed in input (`-e`).
Events defined by a preset set in input are added is not already present in events in input (`-e`).
Duplicated events (except tracepoint) is now an error.

### DIFF
--- a/include/ddprof_input.hpp
+++ b/include/ddprof_input.hpp
@@ -35,6 +35,7 @@ typedef struct DDProfInput {
   char *tags;
   char *url;
   char *socket;
+  char *preset;
   // Watcher presets
   PerfWatcher watchers[MAX_TYPE_WATCHER];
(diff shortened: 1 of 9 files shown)


r1viollet (approved):
This is awesome ! Presets will be super helpful to simplify the deployment of allocation profiling when it makes sense.
Not blocking but a few thoughts that could be nice
- tests around the override logics
- defining in documentation interaction between options

r1viollet (approved):
Thanks for adding these tests :-)
An important feature for the next release !

PR #84, opened 2022-06-13
Cause printargs to always print

sanchda:
# What does this PR do?

Two things
 * Allow the logging interfaces to print, regardless of the user's specified log level
 * Have the printargs interfaces always print, presuming that the user's indication to log means they want the configuration regardless of their specified log level

# Motivation

Right now it's really annoying to get the configuration without also signing up for very verbose logs.

# Additional Notes

Questions
* Should configuration be printed _also_ when the user has informational level or greater for logs?  Right now it only prints when the user requests it
* Are there other obvious places where output should be forced (not necessarily at the INFORMATIONAL level)
* Is INFORMATIONAL the correct level for printargs?

# How to test the change?

idk lol (will add)

### DIFF
--- a/include/logger.h
+++ b/include/logger.h
@@ -20,8 +20,16 @@ typedef enum LOG_OPTS {
   LOG_FILE = 4,
 } LOG_OPTS;
 
+
 typedef enum LOG_LVL {
-  LL_EMERGENCY = 0,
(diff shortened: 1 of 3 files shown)


r1viollet (approved):
Thanks ! 👍 
LGTM

PR #85, opened 2022-06-13
Attempt to fix IPC test

nsavoire:
# What does this PR do?

* Make sure that child fork exits
* Round up duration to next multiple of 4ms

### DIFF
--- a/test/ipc-ut.cc
+++ b/test/ipc-ut.cc
@@ -91,8 +91,9 @@ TEST(IPCTest, timeout) {
     // timeout measurement is not very accurate
     auto d = std::chrono::steady_clock::now() - t0;
     if (d < timeout) {
-      printf("Error1: %s, d=%.1fms\n", strerror(ec.value()),
-             d.count() / 1000000.0);
+      LG_ERR("Read timeout error: errno=%s, duration=%.1fms",
+             strerror(ec.value()),
+             std::chrono::duration<double, std::milli>(d).count());
     }
     ASSERT_GE(d, timeout - timeout_tolerance);
     ASSERT_TRUE(ec);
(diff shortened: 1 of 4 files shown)


test/ipc-ut.cc
```diff
@@ -25,13 +25,11 @@ TEST(IPCTest, Positive) {
   // Fork
   pid_t child_pid = fork();
   if (!child_pid) {
-    std::string fileName = UNIT_TEST_DATA "/ipc_test_data_Positive.txt";
+    FILE *tmp_file = std::tmpfile();
```
sanchda, line 28: I think this is good.  Arguably, we should do a better job cleaning up the temporary files in the face of a catastrophic failure.  The standard doesn't specify what needs to happen to the temporaries if the process is closed abruptly when the file is open.  In theory we can `unlink()` the path immediately after calling `fileno()`, but I don't know how much value there is in doing that.

test/ipc-ut.cc
```diff
@@ -42,7 +40,7 @@ TEST(IPCTest, Positive) {
     EXPECT_FALSE(ec);
     close(sockets[kChildIdx]);
     close(fileFd);
-    return;
+    exit(0);
```
sanchda, line 43: I fear that I will have a hard time remembering this.  `return()` actually gives control back up to the unit test handler, so `fork()` will resume.  My brain wants to think of each unit test as an isolated `main()`.

sanchda: (there's nothing to do about that, unless you can propose a patch for my brain)

test/ipc-ut.cc
```diff
@@ -85,8 +88,9 @@ TEST(IPCTest, timeout) {
     auto t0 = std::chrono::steady_clock::now();
     ASSERT_EQ(sock2.receive(buffer, ec), 0);
     // timeout measurement is not very accurate
-    auto d = (std::chrono::steady_clock::now() - t0) * 4;
-    ASSERT_GE(d, timeout);
+    auto d = std::chrono::steady_clock::now() - t0;
+    auto d_rounded = std::chrono::ceil<jiffies>(d);
```
sanchda, line 92: I feel like this is probably the most reasonable lower bound we can give.

sanchda (approved):
LGTM; comments are narrative only, no need to act on any of them.  Thanks for the quick fix!

PR #87, opened 2022-06-14
Fix style nags

sanchda:
This fixes a bad thing David did.  Bad, David, bad!

### DIFF
--- a/include/logger.h
+++ b/include/logger.h
@@ -20,7 +20,6 @@ typedef enum LOG_OPTS {
   LOG_FILE = 4,
 } LOG_OPTS;
 
-
 typedef enum LOG_LVL {
   LL_FORCE_ALERT = -1,
   LL_FORCE_CRITICAL = -2,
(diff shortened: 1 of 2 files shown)


nsavoire (approved):
LGTM, thanks for the quick fix !

include/logger.h
```diff
@@ -94,10 +93,11 @@ int LOG_getlevel();
 void LOG_setfacility(int);
 
 /******************************* Logging Macros *******************************/
-#define ABS(__x) ({                                                            \
-  typeof(__x) _x = (__x);                                                      \
-  _x < 0 ? -1 * _x : _x;                                                           \
-})
+#define ABS(__x)                                                               \
+  ({                                                                           \
+    typeof(__x) _x = (__x);                                                    \
```
sanchda, line 98: BTW, this would break or be weird if the user passes an expression that evalutes the variable `_x` as a LHS.  Such as `ABS(_x=5)`, but that's slightly typical for these sorts of macros anyway and doesn't reflect the way we'd call `ABS()` in our code.

PR #91, opened 2022-06-16
Improve stats

nsavoire:
* express cpu usage as millicores instead of time
* add stats on truncated / incompleted stacks


src/unwind.cc
```diff
@@ -38,6 +44,22 @@ void unwind_init_sample(UnwindState *us, uint64_t *sample_regs,
   us->stack = sample_data_stack;
 }
 
+static bool is_stack_complete(UnwindState *us) {
+  static constexpr std::array s_expected_root_frames{"_start"sv, "__clone"sv};
```
r1viollet, line 48: should this depend on runtime ?
This KPI could be super interesting to measure the quality of the unwinding.

r1viollet (approved):
LGTM
Thanks for these enhancements !

src/unwind.cc
nsavoire: This is not perfect for sure, but lacking a better way, @sanchda and me agreed that it would be a good start.
Looking at profiles from whole host, this will not work for Go processes for example.

PR #93, opened 2022-06-22
Fix warning on ARM

nsavoire:


r1viollet (approved):
Back to no warnings, thanks

PR #97, opened 2022-06-27
Fix DSO memory layout

r1viollet:
# What does this PR do?

We were assuming that the elf layout was the same as the file layout.
This now registers the module in the correct address for the few cases where we do not match between file and elf layout.

Other minor adjustments:
- Find exe file using the proc exe link if we can't find it in the DSOs
- Adjust the symbol caching addresses

# Motivation

Fix infinite unwinding

# Additional Notes

Some amount of time was spent checking for impacts in whole host mode. Performances look the same (or slightly better)

# How to test the change?

Some unit tests are there, the rest of the tests should cover the unwinding.


include/ddprof_module.hpp
```diff
@@ -0,0 +1,40 @@
...
+
+  DDProfMod()
+      : _mod(nullptr), _low_addr(0), _high_addr(0),
+        _sym_bias(static_cast<Offset_t>(-1)), _status(kUnknown) {}
+
+  explicit DDProfMod(Status status) : DDProfMod() { _status = status; }
+
+  // In the current version of dwfl, the dwfl_module addresses are stable
+  Dwfl_Module *_mod;
+  ProcessAddress_t _low_addr;
```
r1viollet, line 34: this should have been a range

include/ddprof_module.hpp
```diff
@@ -0,0 +1,40 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
```
r1viollet, line 1: we could unify naming of files (dwfl_modules /ddprof_module)

include/dso_symbol_lookup.hpp
```diff
@@ -15,8 +15,9 @@ namespace ddprof {
 
 class DsoSymbolLookup {
 public:
-  SymbolIdx_t get_or_insert(ElfAddress_t addr, const Dso &dso,
-                            SymbolTable &symbol_table);
+  SymbolIdx_t get_or_insert(FileAddress_t normalized_addr, const Dso &dso,
```
r1viollet, line 18: Feedback from Nicolas (not fully relevant to this PR): 
We should do the same as Go for the mappings 
- Single mapping per binary,
Low address is mod low (same for high).
Offset is 0

- We don't need to display these addresses (though they are nice to have within the pprof, for debug purposes)

r1viollet:
General feedback : we need to add some test on following cases
- Position dependant
- Split debug (warning several variations)

src/dwfl_module.cc
```diff
@@ -5,90 +5,76 @@
...
-        const char *mod_name = get_filename(main_name);
-        if (strcmp(mod_name, dso_name) != 0) {
-          // A dso replaced the mod - flag as incoherent for future release
-          LG_NTC("Incoherent DSO (%s) %lx != %lx dwfl_module (%s)",
-                 filepath.c_str(), dso._start - dso._pgoff,
-                 ddprof_mod._low_addr, main_name);
-          return DDProfMod(DDProfMod::kInconsistent);
-        }
-      }
+    if (ddprof_mod._low_addr != mod_range._low_addr) {
```
r1viollet, line 33: Feedback from Nicolas: This is an issue and yet we won't notice this.

src/dwfl_module.cc
```diff
@@ -5,90 +5,76 @@
...
+      LG_NTC("Incoherent Modules (%s-%s) %lx != %lx dwfl_module)", module_name,
```
r1viollet, line 34: Test that when a new dso is mapped (within an existing mod_range), we update the ddprof_module

src/dwfl_symbol.cc
```diff
@@ -40,74 +52,39 @@ bool symbol_get_from_dwfl(Dwfl_Module *mod, ProcessAddress_t process_pc,
...
-  } else {
-    symbol._lineno = 0;
+  if (line) {
+    int linep;
+    const char *localsrcpath =
+        dwfl_lineinfo(line, &process_pc, static_cast<int *>(&linep), 0, 0, 0);
+    if (localsrcpath) {
+      symbol._srcpath = std::string(localsrcpath);
+      symbol._lineno = static_cast<uint32_t>(linep);
+    }
```
r1viollet, line 72: do we need to errno ?

src/dwfl_module.cc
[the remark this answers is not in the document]
r1viollet: So this is not correctly handled and this is not totally obvious
There is a difference between the lifetime of the sample and the state of the proc maps.

I think the best way is to list the files that were unloaded during a DSO update. If any of the unloaded DSOs had files linked, we should check for any related modules. 

Sadly, if these were loaded in the dwfl library, we will have to destroy the dwfl object.

include/ddprof_module.hpp
r1viollet: It might be better in the current form

include/dso_symbol_lookup.hpp
r1viollet: Opening a ticket to store information that allows us to symbolize after capture of stack traces

src/dwfl_module.cc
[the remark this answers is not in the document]
r1viollet: Opened a ticket to follow up on this

r1viollet: This is a case where we need better telemetry

src/dwfl_symbol.cc
r1viollet: Yes we do, APIs actually never mention this.

src/dwfl_module.cc
r1viollet: We don't, follow up ticket.

src/dso_hdr.cc
```diff
@@ -125,6 +124,25 @@ DsoHdr::DsoHdr() {
   _file_info_vector.emplace_back(FileInfo(), 0, true);
 }
 
+namespace {
+bool string_readlink(const char *path, std::string &link_name) {
+  char buff[1024];
+  ssize_t len = ::readlink(path, buff, sizeof(buff) - 1);
```
nsavoire, line 130: [Minor] You could use [std::filesystem::read_link](https://en.cppreference.com/w/cpp/filesystem/read_symlink)

nsavoire (approved):
Thanks for fixing this issue !

src/dso_hdr.cc
r1viollet: - [ ] improve readlink using cpp API

PR #100, opened 2022-07-07
Make failure to connect to library non-fatal in wrapper mode

nsavoire:
# What does this PR do?

Make failure to connect to library non-fatal in wrapper mode

# Motivation

LD_PRELOAD does not work for statically linked executables (eg. go binaries)

# Additional Notes


r1viollet (approved):
LGTM. Error message is very clear. Thanks

Issue #88, opened 2022-06-14
Alpine 3.16 build


creationix:
I'm trying to build from source so I can use this in alpine images, but it's failing on an [import](https://github.com/DataDog/ddprof/blob/main/src/pprof/ddprof_pprof.cc#L9):

```c
#include <ddprof/ffi.h>
```

I can't even find a single reference on google for `ddprof/ffi.h` nor can I find it in the git history here.

Here is my Dockerfile for reference:
```dockerfile
FROM --platform=linux/amd64 alpine:3.16 as build-ddprof
RUN apk add curl cmake build-base xz-dev gtest-dev libcap-dev libexecinfo-dev elfutils-dev
RUN curl -L https://github.com/DataDog/ddprof/archive/refs/tags/v0.8.1.tar.gz | tar -xzv
RUN cd ddprof-0.8.1 && mkdir build && cd build && cmake ..
RUN make -C ddprof-0.8.1/build -j $(nprocs)
```

The full error is:

```sh
/ddprof-0.8.1/src/pprof/ddprof_pprof.cc:11:10: fatal error: ddprof/ffi.h: No such file or directory
   11 | #include <ddprof/ffi.h>
      |          ^~~~~~~~~~~~~~
compilation terminated.
```

creationix:
I think I found it.  It appears to be part of https://github.com/datadog/libdatadog

r1viollet:
Yes, would it help if we provided an alpine build ?
The CMake step is meant to pull down all of the dependencies. I need to update the fetch to retrieve the alpine binaries.

r1viollet:
We now have a branch that builds alpine. We will deliver an alpine binary. Thanks for raising this issue.

r1viollet:
This will be included in the upcoming release, if you want an early custom build or want to discuss other requirements, feel free to reach out.

r1viollet:
This is now released and available in v0.9.3.

PR #104, opened 2022-07-13
Alpine build

r1viollet:
# What does this PR do?

Adapt the build process to work on Alpine
This includes a new docker image to pull down the appropriate dependencies for alpine

co-authored with @sanchda 

# Motivation

Compatibility with alpine
A [ticket](https://github.com/DataDog/ddprof/issues/88) was opened requesting this

# Additional Notes

The CI was also adapted to test alpine.
One unit test is still causing issues. savecontext-ut
I had to downgrade it slightly for now. We suspect this is related to the usage of a signal handler.

# How to test the change?

Tests pass on alpine.


tools/fetch_elfutils.sh
```diff
@@ -68,6 +68,18 @@ patch -p1 < "${SCRIPT_DIR}/elfutils.patch"
 if [[ "$(basename "${C_COMPILER}")" == clang* ]]; then
   export CFLAGS="-Wno-xor-used-as-pow -Wno-gnu-variable-sized-type-not-at-end -Wno-unused-but-set-parameter"
 fi
+
+# Detect musl
+# If so, then set the DFNM_EXTMATCH macro to a harmless value.  We don't use the affected
+# codepaths, so it's OK (TODO: can we just disable compilation of those units?)
+MUSL_LIBC=$(ldd /bin/ls | grep 'musl' | head -1 | cut -d ' ' -f1 || true)
```
nsavoire, line 75: You could have taken build recipe / patches from: 
https://gitlab.alpinelinux.org/alpine/aports/-/blob/master/main/elfutils/APKBUILD

r1viollet: Fixed, thanks

tools/fetch_elfutils.sh
```diff
@@ -68,6 +69,23 @@ patch -p1 < "${SCRIPT_DIR}/elfutils.patch"
...
+if [ ! -z ${MUSL_LIBC-""} ]; then
+  CFLAGS="${CFLAGS-""} -DFNM_EXTMATCH=0"
```
nsavoire, line 78: I don't think this is needed aymore

r1viollet: Fixed, I'll update the build caches

nsavoire (approved):
Thanks for all the work !
LGTM

PR #107, opened 2022-07-15
Force backpopulate per PID

r1viollet:
# What does this PR do?

- Force a backpopulate if we never encountered this PID
- Remove the backpopulate state if we exec
- Avoid creating a module if we detect a bad DSO range

# Motivation

Fix a unit tests where events arrive in wrong order

# Additional Notes

This could have some bad performance impacts, but it ensure we have the correct mappings

# How to test the change?

### DIFF
--- a/src/dso_hdr.cc
+++ b/src/dso_hdr.cc
@@ -281,6 +281,8 @@ DDRes DsoHdr::mod_range_or_backpopulate(DsoMapConstIt it, DsoMap &map,
           return ddres_warn(DD_WHAT_DSO);
         }
       }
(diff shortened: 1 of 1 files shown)


sanchda (approved):
Yeah, I think this is inevitable for now.  We'll get more lost events in our tests, but we should focus on correctness.

To recap some random thoughts from an offline thread:
* We don't order events between ringbuffers
* We can lose precious MMAP events
* I _think_ the perf mmap events over-expose the behavior of the exe loader; if we lose the wrong event in process startup,

PR #108, opened 2022-07-15
Fix flaky unit test

nsavoire:
# What does this PR do?

Fix flaky unit test 
* Improve incomplete stack detection
* Make sure that profiler process has exited before reading log
* Disable leak sanitizer because it triggers a fork that causes incomplete frames
* Allow 3 incomplete frames at most


r1viollet (approved):
LGTM.
Thanks for fixing the unit test !

PR #109, opened 2022-07-18
Revert "Sanchda/elfutils 186"

r1viollet:
# Description

Reverts DataDog/ddprof#106

Revert the revert on elfutils. We initially thought the bad performances were related to the new elfutils verison. Instead 
the issue is related to the CFLAGS sent to elfutils.


nsavoire (approved):
LGTM !

PR #110, opened 2022-07-18
Unset the C flags to avoid building in debug mode

r1viollet:
# What does this PR do?

It is important NOT to set CFLAGS if you don't mean to
Otherwise you will not benefit from -O2 in the elfutils compilation

# Motivation

Fix the performance degradation

# Additional Notes

N/A

# How to test the change?

Relenv / CI


tools/fetch_elfutils.sh
```diff
@@ -68,9 +68,13 @@ patch -p1 < "${SCRIPT_DIR}/elfutils.patch"
 if [[ "$(basename "${C_COMPILER}")" == clang* ]]; then
   export CFLAGS="-Wno-xor-used-as-pow -Wno-gnu-variable-sized-type-not-at-end -Wno-unused-but-set-parameter"
 fi
-export CFLAGS="${CFLAGS-""} ${EXTRA_CFLAGS}"
+# It is important NOT to set CFLAGS if you don't mean to
+# Otherwise you will not benefit from -O2 in the elfutils compilation
+if [ -n "${EXTRA_CFLAGS-""}" ]; then
```
nsavoire, line 73: Name `EXTRA_CFLAGS` is misleading, since they replace default CFLAGS

nsavoire, line 73: With this fix, we still compile elfutils without optimzations for clang (because of extra warning flags set in CFLAGS)

nsavoire (changes requested):

tools/fetch_elfutils.sh
r1viollet: Ah yes !

tools/fetch_elfutils.sh
```diff
@@ -68,9 +68,17 @@ patch -p1 < "${SCRIPT_DIR}/elfutils.patch"
...
   export CFLAGS="-Wno-xor-used-as-pow -Wno-gnu-variable-sized-type-not-at-end -Wno-unused-but-set-parameter"
 fi
-export CFLAGS="${CFLAGS-""} ${EXTRA_CFLAGS}"
+# It is important NOT to set CFLAGS if you don't mean to
+# Otherwise you will not benefit from -O2 in the elfutils compilation
+if [ -n "${C_FLAGS_OVERRIDE-""}" ]; then
+  export CFLAGS="${CFLAGS-""} ${C_FLAGS_OVERRIDE}"
+else # Include the default flags for elfutils
+  if [ -n "${CFLAGS-""}" ]; then
+    export CFLAGS="${CFLAGS-""} -D_FORTIFY_SOURCE=2 -g -O2"
```
nsavoire, line 77: No need for `-D_FORTIFY_SOURCE=2`, it will be added automatically

nsavoire (approved):
LGTM !

PR #112, opened 2022-07-19
Compute ranges in mappings ignoring anonymous regions

r1viollet:
# What does this PR do?

Consider that the mappings can have anon regions that separate valid regions

# Motivation

Fix the computation of the range to avoid loading the modules at the bad address.

# Additional Notes

This was suggested by @nsavoire 

# How to test the change?

I added a unit test

### DIFF
--- a/src/dso_hdr.cc
+++ b/src/dso_hdr.cc
@@ -243,21 +243,36 @@ DDProfModRange DsoHdr::compute_mod_range(DsoMapConstIt it, const DsoMap &map) {
     return DDProfModRange();
   }
   DsoMapConstIt first_el = it;
-  while (first_el != map.begin()) {
-    --first_el;
-    if (it->second._filename != first_el->second._filename) {
-      ++first_el;
-      break;
+  {
+    DsoMapConstIt loop_el = first_el;
+    while (loop_el != map.begin()) {
+      --loop_el;
+      // there can be anonymous regions in the middle of the mapping
+      if (loop_el->second._type != dso::DsoType::kAnon &&
+          it->second._filename != loop_el->second._filename) {
+        break;
+      }
+      if (it->second._filename == loop_el->second._filename) {
+        first_el = loop_el;
+      }
     }
   }
-  while (++it != map.end()) {
-    if (it->second._filename != first_el->second._filename) {
-      break;
+
+  DsoMapConstIt last_el = it;
+  {
+    DsoMapConstIt loop_el = last_el;
+    while (++loop_el != map.end()) {
+      if (loop_el->second._type != dso::DsoType::kAnon &&
+          it->second._filename != loop_el->second._filename) {
+        break;
+      }
+      if (it->second._filename == loop_el->second._filename) {
+        last_el = loop_el;
+      }
     }
   }
-  --it; // back up from end element (or different file)
   return DDProfModRange{._low_addr = first_el->second._start,
-                        ._high_addr = it->second._end};
+                        ._high_addr = last_el->second._end};
 }
 
 // Find the lowest and highest for this given DSO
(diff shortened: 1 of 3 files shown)


src/dso_hdr.cc
```diff
@@ -243,21 +243,36 @@ DDProfModRange DsoHdr::compute_mod_range(DsoMapConstIt it, const DsoMap &map) {
...
     }
   }
-  while (++it != map.end()) {
-    if (it->second._filename != first_el->second._filename) {
-      break;
+
+  DsoMapConstIt last_el = it;
+  {
+    DsoMapConstIt loop_el = last_el;
+    while (++loop_el != map.end()) {
```
sanchda, line 264: I understand that we don't consume `loop_el` when `loop_el == map.end()`.  However, since the iterator doesn't have first-order knowledge of the container boundaries and `if (loop_el == map.end()) ++loop_el;` is UB, does this loop need to be structured similarly to lines 248-249?

Genuine question, I really have no idea.

sanchda (dismissed):
Added some comments, but overall LGTM!

src/dso_hdr.cc
r1viollet: Thanks for checking loop structure
- End element is an invalid element after all elements. We should not process it.
- Begin element is valid and should be processed. Also we can not do a -- on the begin element.

The order of operators means that `++` is done prior to the evaluation of the `==`
If I had written on lines 248
```
    while (--loop_el != map.begin()) {
//...
```
I would not have processed the begin element. Alternatively I think a `do...while` would have been better.... Feel free to tell me if you have a more stylish way of writing this

sanchda: Ah, rookie mistake on my part--thanks for clarifying.

src/dso_hdr.cc
```diff
@@ -243,21 +243,36 @@ DDProfModRange DsoHdr::compute_mod_range(DsoMapConstIt it, const DsoMap &map) {
     return DDProfModRange();
   }
   DsoMapConstIt first_el = it;
-  while (first_el != map.begin()) {
-    --first_el;
-    if (it->second._filename != first_el->second._filename) {
-      ++first_el;
-      break;
+  {
```
nsavoire, line 246:
```  
  const ddprof::Dso *dso = nullptr;
  auto func = [&](auto &x) {
    if (x.second._filename == it->second._filename) {
      dso = &x.second;
      return false;
    };
    return x.second._type != dso::DsoType::kAnon;
  };
  std::find_if(it, std::end(map), func);
  uint64_t high_addr = dso->_end;
  std::find_if(std::make_reverse_iterator(std::next(it)), std::rend(map), func);
  uint64_t low_addr = dso->_start;
  return DDProfModRange{._low_addr = low_addr, ._high_addr = high_addr};
```
```suggestion
  const ddprof::Dso *dso = nullptr;
  auto func = [&](auto &x) {
    if (x.second._filename == it->second._filename) {
      dso = &x.second;
      return false;
    };
    return x.second._type != dso::DsoType::kAnon;
  };
  std::find_if(it, std::end(map), func);
  uint64_t high_addr = dso->_end;
  std::find_if(std::make_reverse_iterator(std::next(it)), std::rend(map), func);
  uint64_t low_addr = dso->_start;
  return DDProfModRange{._low_addr = low_addr, ._high_addr = high_addr};  
```

nsavoire: Not sure if it's better ;)

r1viollet: It is more compact and stylish ! Though it took me some time to read it :-)

nsavoire (dismissed):

nsavoire (approved):

PR #114, opened 2022-07-20
Remove backpopulate state reset every 200 samples

r1viollet:
# What does this PR do?

We were keeping a backpopulate state per PID. By reseting, were were looping over all PIDs every 200 samples, this was causing a large overhead. 

This is related to the recent update in the way we cleared backpopulate states which increased overhead.

The overhead was investigated on a large machine, where we need to consider 90 CPUs of activity.
We will now only clear backpopulate state at every cycle (59 seconds). 

# Motivation

Reduce overhead

# Additional Notes

This could reduce accuracy in mappings. 

# How to test the change?

I tested this in whole host mode (where the problem is the most obvious) on debug PODs


nsavoire (approved):
LGTM!

PR #115, opened 2022-07-21
Improve stats

nsavoire:
# What does this PR do?

* change unwind stat from cpu millicores to wall time since that what is measured
* separate unwind / aggregation
* normalize unwing / aggreagtion time per sample
* add average stack depth
* add profile duration


r1viollet (approved):
LGTM
Improved visibility is nice !

PR #116, opened 2022-07-21
Always generate debug information

nsavoire:
# What does this PR do?

Generate debug information even in release mode


r1viollet (approved):
LGTM
We'll rechange it if needed.

PR #118, opened 2022-07-22
Small improvements

nsavoire:
# What does this PR do?
* Handle mmap failures in TSC timer setup
* Map ring buffers both in parent and worker to avoid losing mmap events at startup
* Use mirroring for perf ring buffers
* Fix use of iterator after free


r1viollet (approved):
Thanks ! More than small improvements
LGTM

PR #119, opened 2022-07-25
Compute elf base address from text segment mapping and segment vaddr / offset

nsavoire:
# What does this PR do?

* Only keep track of executable segments and remove module range computation
* Base address is computed as dso.start - dso.pgoff + seg.offset - seg.vaddr + seg#0.vaddr
  Where `seg` is the elf segment corresponding to the DSO (we make the assumption that there
  is only one executable segment per ELF file, if that's not enough we could add a size check).
  `seg#0` is the first LOAD segment in the ELF file.
* Bias is computed as dso.start - dso.pgoff + seg.offset - seg.vaddr


src/ddprof_module_lib.cc
```diff
@@ -20,34 +114,36 @@ DDRes update_module(Dwfl *dwfl, ProcessAddress_t pc,
...
-      ddprof_mod._status = DDProfMod::kInconsistent;
-      return ddres_warn(DD_WHAT_MODULE);
-    }
-    return ddres_init();
+    Dwarf_Addr low_addr, high_addr;
+    dwfl_module_info(ddprof_mod._mod, 0, &low_addr, &high_addr, 0, 0,
+                     &main_name, 0);
+    LG_NTC("Incoherent modules: module %s [%lx-%lx] already exists at %lx (%s)",
+           main_name, low_addr, high_addr, pc, module_name);
+    ddprof_mod._status = DDProfMod::kInconsistent;
```
r1viollet, line 127: this is actually more robust

src/ddprof_module_lib.cc
```diff
@@ -6,12 +6,106 @@
...
+#include <sys/types.h>
+#include <unistd.h>
 
 namespace ddprof {
 
-DDRes update_module(Dwfl *dwfl, ProcessAddress_t pc,
-                    const DDProfModRange &mod_range,
+static bool get_elf_offsets(const std::string &filepath, Offset_t &start_offset,
+                            Offset_t &bias_offset) {
+  int fd = ::open(filepath.c_str(), O_RDONLY);
```
r1viollet, line 25: we could have a single fd per file that we keep open and give to dwfl (this seems to generate less file opens)
I'll add that in later commit

r1viollet (approved):
LGTM
Tried the branch on my end, everything looks good after these last few fixes

PR #122, opened 2022-07-28
Fix LD_PRELOAD failure due to permissions on /proc/<ddprof_pid>/fd/

nsavoire:
# What does this PR do?

Permissions to read /proc/<pid>/fd are influenced by process's dumpable attribute
(cf. https://man7.org/linux/man-pages/man5/proc.5.html).
When changing the effective user ID, dumpable attribute is reset to /proc/sys/fs/suid_dumpable
(usually 0, cf. https://man7.org/linux/man-pages/man2/prctl.2.html). With dumpable
attribute set to 0, ownership of /proc<pid>/ is set to root, this should to be an issue
since we change euid only when root, but strangely doing this, parent process loses the permission
to read /proc/<ddprof_pid>/fd/* (but not /proc/<ddprof_pid>/maps).
The workaround is to reset dumpable attribute to its initial value after each effective
user ID change.

### DIFF
--- a/src/exe/main.cc
+++ b/src/exe/main.cc
@@ -211,6 +211,7 @@ static int start_profiler_internal(DDProfContext *ctx, bool &is_profiler) {
           preload_str.append(";");
           preload_str.append(s);
         }
+        LG_DBG("Setting LD_PRELOAD=%s", preload_str.c_str());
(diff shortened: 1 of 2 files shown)


r1viollet (approved):
LGTM
To be included in the release

PR #124, opened 2022-08-01
Fix deadlock in pthread_getattr_np

nsavoire:
# What does this PR do?

Deadlock occurs when user code in created thread calls pthread_getattr:
 - pthread_getattr takes a lock in pthread object
 - pthread_getattr itself does an allocation
 - AllocationTracker tracks the allocation and calls savecontext
 - savecontext calls pthread_getattr to get stack end address
 - pthread_getattr is reentered and attempts to take the lock again leading to
   a deadlock.

Workaround is to hook pthread_create and call `cache_stack_end` to
cache stack end address while temporarily disabling allocation profiling for
current thread before calling user code.

### DIFF
--- a/src/lib/pthread_hook.cc
+++ b/src/lib/pthread_hook.cc
@@ -1,0 +1,50 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0. This product includes software
+// developed at Datadog (https://www.datadoghq.com/). Copyright 2021-Present
+// Datadog, Inc.
+
+#include "allocation_tracker.hpp"
+
+#include <dlfcn.h>
+#include <pthread.h>
+#include <tuple>
+
+namespace {
+using Args = std::tuple<void *(*)(void *), void *>;
+
+void *mystart(void *arg) {
+  ddprof::AllocationTracker::notify_thread_start();
+  Args *args = reinterpret_cast<Args *>(arg);
+  auto [start_routine, start_arg] = *args;
+  delete args;
+  return start_routine(start_arg);
+}
+
+} // namespace
+
+/** Hook pthread_create to cache stack end address just after thread start.
+ *
+ * The rationale is to fix a deadlock that occurs when user code in created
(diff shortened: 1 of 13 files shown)


src/lib/pthread_hook.cc
```diff
@@ -0,0 +1,54 @@
...
+ * - AllocationTracker tracks the allocation and calls savecontext
+ * - savecontext calls pthread_getattr to get stack end address
+ * - pthread_getattr is reentered and attempts to take the lock again leading to
+ * a deadlock.
+ *
+ * Workaround is to hook pthread_create and call `cache_stack_end` to
+ * cache stack end address while temporarily disabling allocation profiling for
+ * current thread before calling user code.
+ * */
+int pthread_create(pthread_t *thread, const pthread_attr_t *attr,
```
r1viollet, line 44: I love this mechanism.
I also think it will be helpful to identify thread lifetimes.

src/savecontext.cc
```diff
@@ -31,17 +41,19 @@ DDPROF_NOINLINE static const std::byte *get_stack_end_address() {
...
 static DDPROF_NO_SANITIZER_ADDRESS size_t
 save_stack(const std::byte *stack_ptr, ddprof::span<std::byte> buffer) {
-  // Use a thread local variable to cache the stack end address per thread
-  thread_local static const std::byte *stack_end{};
-
+  const std::byte *stack_end = s_stack_end;
   if (unlikely(!stack_end)) {
     // Slow path, takes ~30us
-    stack_end = get_stack_end_address();
+    stack_end = cache_stack_end();
```
r1viollet, line 47: I would possibly just disable any kind of custom events profiling if we don't have this filled.
Though this code path is unlikely:
- We did not hook into pthread_create correctly ?
- We are not using pthreads ?
I'll look more into this

r1viollet (approved):
LGTM. Nice fix ! 
Left a minor comment, not sure if relevant yet

r1viollet (approved):
LGTM

PR #129, opened 2022-08-03
Refactor ring buffer creation

nsavoire:
# What does this PR do?

* Remove mirror option from ring buffer creation, ring buffers are always mirrored
* Add a ring buffer type to accommodate for different types of ring buffers
* Move all ring buffer read / write logic to ringbuffer_utils.hpp

### DIFF
--- a/include/perf_ringbuffer.hpp
+++ b/include/perf_ringbuffer.hpp
@@ -7,8 +7,6 @@
 
 #include "perf.hpp"
 
-#include <atomic>
-
 enum class RingBufferType : uint8_t { kPerfRingBuffer };
 
 struct RingBuffer {
--- a/include/perf_ringbuffer.hpp
+++ b/include/perf_ringbuffer.hpp
@@ -7,26 +7,28 @@
 
 #include "perf.hpp"
 
-typedef struct RingBuffer {
-  const char *start;
-  unsigned long offset;
-  size_t mask;
-  size_t meta_size; // size of the metadata page
-  size_t data_size; // size of data pages
-  size_t size;      // size of the entire region (data + metadata)
-  struct perf_event_mmap_page *region;
-  bool is_mirrored;
-  unsigned char *wrbuf;
-} RingBuffer;
+#include <atomic>
 
-bool rb_init(RingBuffer *rb, struct perf_event_mmap_page *page, size_t size,
-             bool is_mirrored);
+enum class RingBufferType : uint8_t { kPerfRingBuffer };
+
+struct RingBuffer {
+  RingBufferType type;
+  uint64_t *writer_pos;
+  uint64_t *reader_pos;
+  uint64_t mask;
+
+  size_t meta_size; // size of the metadata
+  size_t data_size; // size of data
+  std::byte *data;
+  void *base;
+};
+
+bool rb_init(RingBuffer *rb, void *base, size_t size, RingBufferType type);
 void rb_free(RingBuffer *rb);
-void rb_clear(RingBuffer *rb); // does not deallocate the buffer storage
-uint64_t rb_next(RingBuffer *rb);
-struct perf_event_header *rb_seek(RingBuffer *rb, uint64_t offset);
+
 bool samp2hdr(struct perf_event_header *hdr, perf_event_sample *sample,
               size_t sz_hdr, uint64_t mask);
+
 perf_event_sample *hdr2samp(const struct perf_event_header *hdr, uint64_t mask);
 
 uint64_t hdr_time(struct perf_event_header *hdr, uint64_t mask);
--- a/include/pevent.hpp
+++ b/include/pevent.hpp
@@ -20,6 +20,7 @@ typedef struct PEvent {
   int mapfd;               // FD for ring buffer, same as `fd` for perf events
   int attr_idx;            // matching perf_event_attr
   size_t ring_buffer_size; // size of the ring buffer
+  RingBufferType ring_buffer_type;
   bool custom_event; // true if custom event (not handled by perf, eg. memory
                      // allocations)
   RingBuffer rb;     // metadata and buffers for processing perf ringbuffer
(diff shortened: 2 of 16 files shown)


include/perf_ringbuffer.hpp
```diff
@@ -7,26 +7,28 @@
...
-  size_t data_size; // size of data pages
-  size_t size;      // size of the entire region (data + metadata)
-  struct perf_event_mmap_page *region;
-  bool is_mirrored;
-  unsigned char *wrbuf;
-} RingBuffer;
-
-bool rb_init(RingBuffer *rb, struct perf_event_mmap_page *page, size_t size,
-             bool is_mirrored);
+#include <atomic>
```
r1viollet, line 10: minor: I don't see usage of std::atomic 
Needed ?

include/perf_ringbuffer.hpp
```diff
@@ -7,26 +7,28 @@
...
 
-typedef struct RingBuffer {
-  const char *start;
-  unsigned long offset;
-  size_t mask;
-  size_t meta_size; // size of the metadata page
-  size_t data_size; // size of data pages
-  size_t size;      // size of the entire region (data + metadata)
-  struct perf_event_mmap_page *region;
-  bool is_mirrored;
```
r1viollet, line 18: interesting, so the mirroring never accounts for more pinned memory

include/ringbuffer_utils.hpp
```diff
@@ -17,55 +17,78 @@ namespace ddprof {
 
 struct RingBufferInfo;
 
-class RingBufferWriter {
+// Return `x` rounded up to next multiple of `y`
+// `y` must be a power of 2
+// Return 0 for x==0 or x > std::numeric_limits<uint64_t>::max()-7
+constexpr inline uint64_t round_up_to_mutiple_of_pow2(uint64_t x, uint64_t y) {
```
r1viollet, line 23: minor, naming is a little confusing, it is a multiple of `y`
Nice use of constexpr

include/perf_ringbuffer.hpp
nsavoire: Bad split of changes on my change, I will remove this

include/perf_ringbuffer.hpp
nsavoire: Yes, I checked in the kernel source.
cf. https://github.com/DataDog/ddprof/pull/118/commits/fc3a2c1fd4e167b868478d56d2e97ed716e9943b

src/perf_mainloop.cc
```diff
@@ -142,52 +143,44 @@ static DDRes signalfd_setup(pollfd *pfd) {
...
+
+      ddprof::ConstBuffer buffer = reader.read_all_available();
+      while (!buffer.empty()) {
+        auto *hdr = reinterpret_cast<const perf_event_header *>(buffer.data());
         DDRes res = ddprof_worker_process_event(hdr, pes[i].watcher_pos, ctx);
 
-        // We've processed the current event, so we can advance the ringbuffer
-        tail += hdr->size;
+        // \fixme{nsavoire} free slot as soon as possible ?
+        // reader.advance(hdr->size);
```
r1viollet, line 165: I'm missing why would we not advance here ?

include/ringbuffer_utils.hpp
nsavoire: `constexpr` is not useful for my use here, but it does not hurt :)
Comments clearly explain what this function does. Is it the name of the function or the name of the `y` argument that bothers you ? Do you have an idea for a better name ?

src/perf_mainloop.cc
nsavoire: For the moment, I am keeping the same behaviour as before, advance is done once all available samples of the current ring buffer are processed (it's done by `reader`'s destructor).
Since unwinding is not that fast, it could be advantageous to advance for each sample, doing more writes to the ring buffer reader positions but freeing space faster for the writer.

include/pevent.hpp
```diff
@@ -20,6 +20,7 @@ typedef struct PEvent {
   int mapfd;               // FD for ring buffer, same as `fd` for perf events
   int attr_idx;            // matching perf_event_attr
   size_t ring_buffer_size; // size of the ring buffer
+  RingBufferType ring_buffer_type;
```
nsavoire, line 23: I don't really like the duplication of `ring_buffer_type` and `ring_buffer_size` in `PEvent` and `RingBuffer`, but I am not sure how to improve this.

src/perf_ringbuffer.cc
```diff
@@ -9,73 +9,28 @@
...
   rb->data_size = size - rb->meta_size;
   rb->mask = get_mask_from_size(size);
-  rb->is_mirrored = is_mirrored;
-
-  // If already allocated just free it
-  if (rb->wrbuf)
-    free(rb->wrbuf);
-  rb->wrbuf = NULL; // in case alloc fails
-
-  if (!is_mirrored) {
```
r1viollet, line 28: nice, this simplifies code paths

include/ringbuffer_utils.hpp
r1viollet: Indeed, the comment is good. 
I was just confused as to why the function would take two arguments with current naming, but I can't think of anything better. The pow2 is a pre-requisit of y, so the real naming would be something ugly like
`round_up_to_multiple_of_y_assuming_pow2`

src/perf_mainloop.cc
r1viollet: Agreed, I think this will make things better 👍

include/pevent.hpp
r1viollet: The data we have is fine. Optionally, we could improve the grouping / naming.
We have the data to instrument the ring buffer and the instantiation of the ring buffer itself. We could make that clearer.

test/ringbuffer-ut.cc
```diff
@@ -8,28 +8,251 @@
 #include "ringbuffer_utils.hpp"
```
r1viollet, line 9: thank you for adding good tests to these structures 🙇

r1viollet (dismissed):
Thanks for doing a gentle refactor PR.
This was very understandable. LGTM !

r1viollet (approved):
ACK

PR #130, opened 2022-08-04
Do not use `-o pipefail`

nsavoire:
# What does this PR do?
`-o pipefail` makes `command -v clang-format ... | head -n1` fail with SIGPIPE when
 `head` closes the reading end of the pipe before `command` has finished writing.


r1viollet (approved):
Thanks for looking into this ! This was very mysterious.
LGTM

PR #131, opened 2022-08-04
MPSC ring buffer

nsavoire:
# What does this PR do?
Add MPSC ring buffer

Design is based on BPF ring buffer (https://www.kernel.org/doc/html/latest/bpf/ringbuf.html).
    * Writer first reserves some space at ring buffer head by first taking a lock, advancing
      writer position, marking sample as busy and releasing lock. Lock is required because
      advancing position / marking sample cannot be done atomically.
    * Writer commits a sample by atomically setting its state to ready or discarded
    * Reader is almost the same as for Perf ring buffer, it just additionally check sample state:
      if sample is busy it continues waiting, if sample is discarded, it silently drops it

### DIFF
--- a/test/simple_malloc.cc
+++ b/test/simple_malloc.cc
@@ -3,12 +3,14 @@
 // developed at Datadog (https://www.datadoghq.com/). Copyright 2021-Present
 // Datadog, Inc.
 
+#include <alloca.h>
 #include <chrono>
 #include <cmath>
 #include <cstdio>
 #include <cstdlib>
 #include <cstring>
-#include <future>
+#include <functional>
+#include <sstream>
 #include <thread>
 #include <time.h>
 #include <unistd.h>
@@ -46,36 +48,49 @@ struct Stats {
   pid_t tid;
 };
 
-extern "C" DDPROF_NOINLINE void
-do_lot_of_allocations(uint64_t loop_count, std::chrono::microseconds sleep,
-                      std::chrono::microseconds spin,
-                      std::chrono::milliseconds timeout, Stats *stats) {
+struct Options {
+  uint64_t malloc_size;
+  uint64_t realloc_size;
+  uint64_t loop_count;
+  std::chrono::microseconds spin_duration_per_loop;
+  std::chrono::microseconds sleep_duration_per_loop;
+  std::chrono::milliseconds timeout_duration;
+  uint32_t callstack_depth;
+  uint32_t frame_size;
+};
+
+extern "C" DDPROF_NOINLINE void do_lot_of_allocations(const Options &options,
+                                                      Stats &stats) {
   uint64_t nb_alloc{0};
   uint64_t alloc_bytes{0};
 
   auto start_time = std::chrono::steady_clock::now();
-  auto end_time = start_time + timeout;
+  auto deadline_time = start_time + options.timeout_duration;
   auto start_cpu = thread_cpu_clock::now();
-  for (uint64_t i = 0; i < loop_count; ++i) {
-    void *p = malloc(1000);
-    ddprof::DoNotOptimize(p);
-    ++nb_alloc;
-    alloc_bytes += 1000;
-    void *p2 = realloc(p, 2000);
-    ddprof::DoNotOptimize(p2);
-    ++nb_alloc;
-    alloc_bytes += 2000;
-    free(p2);
-    void *p3 = calloc(1, 512);
-    ddprof::DoNotOptimize(p3);
-    ++nb_alloc;
-    alloc_bytes += 512;
-    free(p3);
-    if (sleep.count()) {
-      std::this_thread::sleep_for(sleep);
+  for (uint64_t i = 0; i < options.loop_count; ++i) {
+    void *p = nullptr;
+    if (options.malloc_size) {
+      p = malloc(options.malloc_size);
+      ++nb_alloc;
+      alloc_bytes += options.malloc_size;
     }
-    if (spin.count()) {
-      auto target_time = std::chrono::steady_clock::now() + spin;
+    ddprof::DoNotOptimize(p);
+    void *p2;
+    if (options.realloc_size) {
+      p2 = realloc(p, options.realloc_size);
+      ++nb_alloc;
+      alloc_bytes += options.realloc_size;
+    } else {
+      p2 = p;
+    }
+    ddprof::DoNotOptimize(p2);
+    free(p2);
+    if (options.sleep_duration_per_loop.count()) {
+      std::this_thread::sleep_for(options.sleep_duration_per_loop);
+    }
+    if (options.spin_duration_per_loop.count()) {
+      auto target_time =
+          std::chrono::steady_clock::now() + options.spin_duration_per_loop;
       do {
         volatile uint64_t sum = 1;
         for (uint64_t j = 0; j < 100; ++j) {
@@ -84,14 +99,35 @@ do_lot_of_allocations(uint64_t loop_count, std::chrono::microseconds sleep,
       } while (std::chrono::steady_clock::now() < target_time);
     }
 
-    if (timeout.count() > 0 && std::chrono::steady_clock::now() >= end_time) {
+    if (options.timeout_duration.count() > 0 &&
+        std::chrono::steady_clock::now() >= deadline_time) {
       break;
     }
   }
   auto end_cpu = thread_cpu_clock::now();
-  end_time = std::chrono::steady_clock::now();
-  *stats = {nb_alloc, alloc_bytes, end_time - start_time, end_cpu - start_cpu,
-            ddprof::gettid()};
+  auto end_time = std::chrono::steady_clock::now();
+  stats = {nb_alloc, alloc_bytes, end_time - start_time, end_cpu - start_cpu,
+           ddprof::gettid()};
+}
+
+extern "C" DDPROF_NOINLINE void recursive_call(const Options &options,
+                                               Stats &stats, uint32_t depth) {
+  void *stack_alloc = nullptr;
+  if (options.frame_size) {
+    stack_alloc = alloca(options.frame_size);
+  }
+  ddprof::DoNotOptimize(stack_alloc);
+
+  if (depth == 0) {
+    do_lot_of_allocations(options, stats);
+  } else {
+    recursive_call(options, stats, depth - 1);
+  }
+  DDPROF_BLOCK_TAIL_CALL_OPTIMIZATION();
+}
+
+extern "C" DDPROF_NOINLINE void wrapper(const Options &options, Stats &stats) {
+  recursive_call(options, stats, options.callstack_depth);
 }
 
 void print_header() {
@@ -111,10 +147,8 @@ int main(int argc, char *argv[]) {
 
     unsigned int nb_forks{1};
     unsigned int nb_threads{1};
-    uint64_t sleep_us{0};
-    uint64_t spin_us{0};
-    int timeout_ms = -1;
-    uint64_t loop_count = std::numeric_limits<uint64_t>::max();
+
+    Options opts;
     std::vector<std::string> exec_args;
 
     app.add_option("--fork", nb_forks, "Number of processes to create")
@@ -123,14 +157,35 @@ int main(int argc, char *argv[]) {
         ->default_val(1);
     app.add_option("--exec", exec_args, "Exec the following command")
         ->expected(-1);
-    app.add_option("--loop", loop_count, "Number of loops")->default_val(-1);
-    app.add_option("--timeout", timeout_ms, "Timeout after N milliseconds")
+    app.add_option("--loop", opts.loop_count, "Number of loops")
+        ->default_val(-1);
+    app.add_option("--malloc", opts.malloc_size,
+                   "Malloc allocation size per loop")
+        ->default_val(1000);
+    app.add_option("--realloc", opts.realloc_size,
+                   "Realloc allocation size per loop")
+        ->default_val(2000);
+    app.add_option("--call-depth", opts.callstack_depth, "Callstack depth")
         ->default_val(0);
-    app.add_option("--sleep", sleep_us,
-                   "Time to sleep (us) between allocations")
+    app.add_option("--frame-size", opts.frame_size,
+                   "Size to allocate on the stack for each frame")
         ->default_val(0);
-    app.add_option("--spin", spin_us, "Time to spin (us) between allocations")
-        ->default_val(0);
+
+    app.add_option<std::chrono::milliseconds, int64_t>(
+           "--timeout", opts.timeout_duration, "Timeout after N milliseconds")
+        ->default_val(0)
+        ->check(CLI::NonNegativeNumber);
+    app.add_option<std::chrono::microseconds, int64_t>(
+           "--sleep", opts.sleep_duration_per_loop,
+           "Time to sleep (us) between allocations")
+        ->default_val(0)
+        ->check(CLI::NonNegativeNumber);
+    app.add_option<std::chrono::microseconds, int64_t>(
+           "--spin", opts.spin_duration_per_loop,
+           "Time to spin (us) between allocations")
+        ->default_val(0)
+        ->check(CLI::NonNegativeNumber);
+
 #ifdef USE_DD_PROFILING
     bool start_profiling = false;
     app.add_flag("--profile", start_profiling, "Enable profiling")
@@ -145,7 +200,6 @@ int main(int argc, char *argv[]) {
       return 1;
     }
 #endif
-
     if (exec_args.empty()) {
       print_header();
     }
@@ -169,14 +223,9 @@ int main(int argc, char *argv[]) {
     std::vector<std::thread> threads;
     std::vector<Stats> stats{nb_threads};
     for (unsigned int i = 1; i < nb_threads; ++i) {
-      threads.emplace_back(do_lot_of_allocations, loop_count,
-                           std::chrono::microseconds{sleep_us},
-                           std::chrono::microseconds{spin_us},
-                           std::chrono::milliseconds{timeout_ms}, &stats[i]);
+      threads.emplace_back(wrapper, std::cref(opts), std::ref(stats[i]));
     }
-    do_lot_of_allocations(loop_count, std::chrono::microseconds{sleep_us},
-                          std::chrono::microseconds{spin_us},
-                          std::chrono::milliseconds{timeout_ms}, &stats[0]);
+    wrapper(opts, stats[0]);
     for (auto &t : threads) {
       t.join();
     }


include/ringbuffer_utils.hpp
```diff
@@ -152,6 +153,151 @@ class PerfRingBufferReader {
...
+
+  static bool is_busy(uint64_t size) { return size & k_busy_bit; }
+  static bool is_discarded(uint64_t size) { return size & k_discard_bit; }
+};
+
+class MPSCRingBufferWriter {
+public:
+  explicit MPSCRingBufferWriter(RingBuffer &rb) : _rb(rb) {
+    assert(rb.type == RingBufferType::kMPSCRingBuffer);
+    _tail = __atomic_load_n(_rb.reader_pos, __ATOMIC_ACQUIRE);
```
ivoanjo, line 169: Minor: I think this can be slightly simplified:

```suggestion
    update_tail();
```

include/ringbuffer_utils.hpp
```diff
@@ -152,6 +153,151 @@ class PerfRingBufferReader {
...
+  ~MPSCRingBufferReader() {
+    if (_tail != _initial_tail) {
+      __atomic_store_n(_rb.reader_pos, _tail, __ATOMIC_RELEASE);
+    }
+  }
+
+  inline size_t available_size() const { return _head - _tail; }
+
+  ConstBuffer read_sample() {
+    size_t n = _head - _tail;
```
ivoanjo, line 258: Minor: 

```suggestion
  inline size_t available_size() const { return _head - _tail; }

  ConstBuffer read_sample() {
    size_t n = available_size();
```

src/perf_ringbuffer.cc
```diff
@@ -17,12 +18,23 @@ bool rb_init(RingBuffer *rb, void *base, size_t size,
...
     rb->writer_pos = reinterpret_cast<uint64_t *>(&meta->data_head);
-  } else {
+    break;
+  }
+  case RingBufferType::kMPSCRingBuffer: {
+    ddprof::MPSCRingBufferMetaDataPage *meta =
+        reinterpret_cast<ddprof::MPSCRingBufferMetaDataPage *>(rb->base);
+    rb->reader_pos = &meta->reader_pos;
+    rb->writer_pos = &meta->writer_pos;
+    rb->spinlock = &meta->spinlock;
```
ivoanjo, line 34: I'm kinda curious about this one -- why store pointers to the `MPSCRingBufferMetaDataPage` struct, rather than putting that struct directly inside `struct RingBuffer`? (potentially as a union, to support the multiple ring buffer types)

test/ringbuffer-ut.cc
```diff
@@ -251,3 +251,134 @@ TEST(ringbuffer, full) {
...
+
+void mpsc_reader_fun(RingBuffer *rb, size_t nb_elements, size_t nb_producers,
+                     bool use_new_object, bool advance_eagerly) {
+  std::optional<MPSCRingBufferReader> reader(*rb);
+
+  std::vector<size_t> counts(nb_producers);
+  size_t total_count = 0;
+  while (total_count < nb_elements * nb_producers) {
+    if (use_new_object) {
+      reader.emplace(*rb);
```
ivoanjo, line 263: Minor: I had to admit I had to look up [emplace](https://www.macmillandictionary.com/dictionary/british/emplace) in the dictionary 😅

ivoanjo (commented):
Hey! @r1viollet suggested I may learn a bit from this PR, and indeed I did!

I also left a few n00b questions, hopefully they aren't too bad :)

(I'm also more scared of C++ than I was before!)

include/mpscringbuffer.hpp
```diff
@@ -0,0 +1,52 @@
...
+#else
+        asm volatile("yield");
+#endif
+      } else {
+        // If active spin fails, yield
+        std::this_thread::sleep_for(k_yield_sleep);
+      }
+      // Wait for lock to be released without generating cache misses
+    } while (lock->test(std::memory_order_relaxed));
+  }
```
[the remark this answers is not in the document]
r1viollet: My understanding is that the order of arrival of the threads and 500 microseconds might allow the threads not to go back to spinning. Though perhaps in some apps all threads arrive at the same time.

include/mpscringbuffer.hpp
```diff
@@ -0,0 +1,52 @@
...
+
+inline void lock(std::atomic_flag *lock) {
+  static constexpr uint32_t k_max_active_spin = 4000;
+  static constexpr std::chrono::nanoseconds k_yield_sleep =
+      std::chrono::microseconds(500);
+
+  uint32_t spincount = 0;
+
+  for (;;) {
+    if (!lock->test_and_set(std::memory_order_acquire)) {
```
r1viollet, line 29: should we consider a second safety mechanism to get out ?

include/ringbuffer_utils.hpp
```diff
@@ -152,6 +153,151 @@ class PerfRingBufferReader {
...
+struct MPSCRingBufferHeader {
+  uint64_t size;
+  static constexpr uint64_t k_discard_bit = 1UL << 62;
+  static constexpr uint64_t k_busy_bit = 1UL << 63;
+
+  static bool is_busy(uint64_t size) { return size & k_busy_bit; }
+  static bool is_discarded(uint64_t size) { return size & k_discard_bit; }
+};
+
+class MPSCRingBufferWriter {
```
sanchda, line 165: I think I understand.  In the common setting, the underlying RingBuffer is going to be shared--either in a `mmap(...MAP_SHARED);` or in a separate thread (multiprocess is the worst-case, so I'll focus on that).  On the other hand, the MPSCRingBufferWriter/Reader classes are _views_ into the Ringbuffer, with all private variables intended to be local to the current thread.

include/ringbuffer_utils.hpp
sanchda, line 169: Got it.  Instantiate the view by taking a snapshot of the current tail.  There's no reason to have a view (or in initialization) of the head of the ringbuffer, since that needs to be sequenced carefully at time of reservation.

include/ringbuffer_utils.hpp
```diff
@@ -152,6 +153,151 @@ class PerfRingBufferReader {
...
+    }
+
+    uint64_t head_linear = writer_pos & _rb.mask;
+    MPSCRingBufferHeader *hdr =
+        reinterpret_cast<MPSCRingBufferHeader *>(_rb.data + head_linear);
+
+    // Mark the sample as busy
+    hdr->size = n | MPSCRingBufferHeader::k_busy_bit;
+
+    // Atomic operation required to synchronize with reader load_acquire
```
sanchda, line 205: Right!  Even though we're locked, we need to push down the write to the _next_ reader, in case the write doesn't go through before we unlock.

include/mpscringbuffer.hpp
```diff
@@ -0,0 +1,52 @@
...
+    do {
+      if (spincount < k_max_active_spin) {
+        ++spincount;
+#ifdef __x86_64__
+        asm volatile("pause");
+#else
+        asm volatile("yield");
+#endif
+      } else {
+        // If active spin fails, yield
```
sanchda, line 41: Since this spinlock is also supposed to support multi-process coordination, I think we might need to make it more robust.  Here are some random ideas.

1.  As Ivo notes in the other comment, when `n_waiters >> n_schedulable_cores`, then it becomes more likely for the thread holding the lock to slow down in a way that compounds contention.  The way that some of the pthreads types handle this is by having the slow-path do an atomic increment indicating the number of "interested" (haha) threads.  Unfortunately, the pthreads code I checked only uses this field for arithmetic, but I guess we could use a combination of the lock-defined `num_interested` and thread-local `interested_idx` to sequence the waits?  This isn't a complete solution, just an idea.
2.  As Erwan indicated, since we're at the mercy of other processes, which may be killed in various situations (user culling heavy processes, OOM, or even in the case where the lock is held across container boundaries and an unrelated service gracefully closes), I wonder if we need to structure a check during the slow path.  This is actually really hard to do because the lock-holder may be in a separate PID namespace.  I'd say that after a certain number of passes (plus the addition of a sequence number or something, so we don't get a lot of threads thrashing on the lock), one of the threads needs to make a call to either take ownership of the lock or declare the lock invalid so the application can figure out what to do.

Lots of words with nothing specific, sorry :(

include/mpscringbuffer.hpp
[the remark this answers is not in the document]
sanchda: This type of fast-path/slow-path spinlock is surprisingly durable.  I'd structure a test which sits on a tight loop over the lock in a pool of threads, doubling the number of threads every iteration from 1x to 1024x the number of CPUs and observe the scaling properties of the lock itself after a token amount of on-CPU work.  I'd expect
* Median wait time per thread to remain linear up to a big multiple of the number of cores.
* After breaking linear scaling, we get quadratic rather than exponential

Why do I think this?  Jiffies are big and CPUs are fast.  All bets are off the table if the work loop has a context switch though, in that case I'm with Ivo.

But again, I think the point is to look at the scaling properties rather than a given a priori overhead value for a fixed number of threads.

I do think that ceiling should be evaluated though.  We're adding a spinlock to a library which is trying to infect its children, and some processes are more, uh, prodigious breeders than others.  :)

include/ringbuffer_utils.hpp
```diff
@@ -152,6 +153,151 @@ class PerfRingBufferReader {
...
+  }
+
+  MPSCRingBufferWriter(const MPSCRingBufferWriter &) = delete;
+  MPSCRingBufferWriter &operator=(const MPSCRingBufferWriter &) = delete;
+
+  void update_tail() {
+    _tail = __atomic_load_n(_rb.reader_pos, __ATOMIC_ACQUIRE);
+  }
+
+  Buffer reserve(size_t n) {
```
sanchda, line 179: I really like `reserve()` semantics here.

One of the challenges though is that a consumer doesn't know if a given element is ready, only if the whole ringbuffer is ready.  I think the current strategy for that problem is good, since the alternative is to have a bitmask for potential reader positions indicating whether the given read position is ready (since we have variable-size elements).  I don't think that's a favorable alternative unless we start hitting issues with sparse updates (lagging thread in the middle of the ringbuffer) or high write contention with fast drains (in which case the buffer is never full but always busy), but I very much doubt that will happen given how big our samples are.

include/ringbuffer_utils.hpp
```diff
@@ -152,6 +153,151 @@ class PerfRingBufferReader {
...
+    size_t n2 = align_up(n + sizeof(MPSCRingBufferHeader), 8);
+    if (n2 == 0) {
+      return {};
+    }
+
+    // \fixme{nsavoire} Not sure if spinlock is the best option here
```
sanchda, line 185: I also don't know whether it's the best, but I do think it's quite lovely.  My $0.02 USD (using USD right now because it's weaker than the Euro and my opinion shouldn't be worth much) is that any decisions about different lock architecture or modifying the reservation sequence to use CAS+atomics depends entirely on the usage pattern, and for that we need to deploy what we have and observe it in its natural setting.

include/ringbuffer_utils.hpp
```diff
@@ -152,6 +153,151 @@ class PerfRingBufferReader {
...
+    unlock(reinterpret_cast<std::atomic_flag *>(_rb.spinlock));
+
+    return {reinterpret_cast<std::byte *>(hdr) + sizeof(MPSCRingBufferHeader),
+            n};
+  }
+
+  // Return true if notification to consumer is necesssary
+  // Notification is necessary only if consumer has caught up with producer
+  // (meaning tail afer commit is at or after head before commit)
+  bool commit(Buffer buf, bool discard = false) {
```
sanchda, line 216: Beautiful.  Again, I love reservation semantics.

include/ringbuffer_utils.hpp
```diff
@@ -152,6 +153,151 @@ class PerfRingBufferReader {
...
+private:
+  RingBuffer &_rb;
+  uint64_t _tail;
+};
+
+class MPSCRingBufferReader {
+public:
+  explicit MPSCRingBufferReader(RingBuffer &rb) : _rb(rb) {
+    assert(rb.type == RingBufferType::kMPSCRingBuffer);
+    _head = __atomic_load_n(rb.writer_pos, __ATOMIC_ACQUIRE);
```
sanchda, line 243: Hey, is head/tail read-after-write sequencing an issue here?  I see that we set the MO for the head, and I see that the reserve/commit sequence by all means *should* update the head before the tail.  My gut tells me that this is fine.  Any reason to believe otherwise?

include/ringbuffer_utils.hpp
```diff
@@ -152,6 +153,151 @@ class PerfRingBufferReader {
...
+      return {};
+    }
+
+    uint64_t tail_linear = _tail & _rb.mask;
+    std::byte *start = _rb.data + tail_linear;
+    MPSCRingBufferHeader *hdr = reinterpret_cast<MPSCRingBufferHeader *>(start);
+    uint64_t sz = __atomic_load_n(&hdr->size, __ATOMIC_ACQUIRE);
+
+    // Sample not committed yet, bail out
+    if (MPSCRingBufferHeader::is_busy(sz)) {
```
sanchda, line 269: Makes sense to me.  Leave it to the calling application to determine how to handle the scenario where the ringbuffer is in the middle of a reserve-commit sequence.

test/ringbuffer-ut.cc
sanchda: lol, also https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2008/n2642.pdf

But fundamentally, the C++ standards committee has a precise technical goal for using the term "emplace" to qualify a particular type of insertion.

sanchda (dismissed):
Some pretty good stuff here!  I like the spinlock, I think it's going to scale pretty well.  I like the reserve-commit semantics.  I think we should all do a little bit of scaling analysis, but I don't think that work needs to block merging into main.

My only reservation is that this is a multiprocess MPCS structure, but because of its application domain we don't have a mechanism to detect lock-holder expiry/deadlock.  I think we can structure a solution, but I would be very happy making that improvement a subsequent PR rather than overloading this one with too many expectations.

include/mpscringbuffer.hpp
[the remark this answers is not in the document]
sanchda: IMHO, this is pretty idiomatic.
1. Use of pause/yield on x86/arm to tell the processor that this is a spin-wait loop
2. Use of OS sched yield to sleep thread after fixed number of iterations in the fast path, back to lock

One of the problems with the classic spinlock is, as Ivo notes, suppose that the spinlock is the only thing on CPU, with a bunch of on- and off-cpu threads in contention.  Draining the spinlock will depend on the active thread being able to progress, and whenever it has to yield to the scheduler it might have to round-robin behind all of the waiting threads as they clash on the lock.

test/ringbuffer-ut.cc
ivoanjo: Someone wrote 8 pages on this matter....

/me backs away slowly and silently

include/mpscringbuffer.hpp
```diff
@@ -0,0 +1,52 @@
...
+// Datadog, Inc.
+#pragma once
+
+#include <atomic>
+#include <chrono>
+#include <thread>
+
+namespace ddprof {
+//  mimic: std::hardware_destructive_interference_size, C++17
+constexpr std::size_t hardware_destructive_interference_size = 128;
```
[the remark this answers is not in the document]
nsavoire: C++ standards are often not fully implemented (even when compilers claim support), especially on the standard library side !
clang has not support yet for hardware_destructive_interference_size yet (https://libcxx.llvm.org/Status/Cxx17.html) and only recent gcc versions (>=12.1) have support for it (https://gcc.gnu.org/onlinedocs/libstdc++/manual/status.html#status.iso.2017)

src/perf_ringbuffer.cc
nsavoire: I hesitated some time between both solutions. I think at the start I thought I could factorize code between ring buffer implementations by having a common pointer that expose the same semantic but points to different parts depending on the type.
But I agree that since I was not able to factorize code, clearly separating the data for different implementations might make the code mode readable.

test/ringbuffer-ut.cc
nsavoire: C++ is too complicated for its own good !
The idea behind `emplace` is to be able to construct an object in-place in a container, potentially saving the construction / copy of a temporary object.

ivoanjo:
Thanks for the explanations @nsavoire! Learned a bunch of modern C++ things with this PR :)

include/mpscringbuffer.hpp
nsavoire: I see @r1viollet's point:  if a thread/process dies while having the lock, then other processes will wait endlessly.
I will add a timeout.

include/mpscringbuffer.hpp
```diff
@@ -0,0 +1,52 @@
...
+  static constexpr std::chrono::nanoseconds k_yield_sleep =
+      std::chrono::microseconds(500);
+
+  uint32_t spincount = 0;
+
+  for (;;) {
+    if (!lock->test_and_set(std::memory_order_acquire)) {
+      break;
+    }
+    do {
```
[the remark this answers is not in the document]
nsavoire: Not my idea ;)
https://probablydance.com/2019/12/30/measuring-mutexes-spinlocks-and-how-bad-the-linux-scheduler-really-is/

include/ringbuffer_utils.hpp
nsavoire: The object model is not clear because there is not object that represents the ring buffer and is responsible for its creation/destruction.
`RingBuffer` holds the data necessary to access the ring buffer but does not act on it. The only requirement is a zero-initialized chunk of memory. And it is indeed shared explicitly (several threads accessing the same object) or implicitly (pointers in different `RingBuffer` objects pointing to the same underlying memory).
Writer/Reader are not really views (IMHO `RingBuffer` is the view) but local short-lived objects that modify the state of the ring buffer in a consistent manner.

include/ringbuffer_utils.hpp
nsavoire: Indeed the lock is only used by writers, reader/writer synchronization relies on atomic loads/stores.

include/mpscringbuffer.hpp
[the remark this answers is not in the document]
ivoanjo: Ahhhh you're right about `spincount`, I completely missed that part and read it as the loop starting as-if from scratch 😅

include/mpscringbuffer.hpp
nsavoire: Concerning point 1, I will add a benchmark to be able to base our decisions on measures.
Point 2 is very important, I definitely need to add safeguards. Disabling allocation profiling when we detect that a lock is never released seems the safest choice to me.

include/ringbuffer_utils.hpp
nsavoire: For the calling application, we behave as if the ringbuffer was empty, we don't convey the information that a sample is being finalized.
We could try reading the next samples but this would add a lot more complexity. WDYT ?

include/ringbuffer_utils.hpp
nsavoire: I am not sure to understand your concern.
`_tail` is only written by reader and read by writer, reads of `_tail` in reader can be relaxed since writes are done by the same thread.

include/ringbuffer_utils.hpp
nsavoire: I would be interested to discuss your idea !
Initially I started an implementation without a spinlock, the idea was to use a ready bit, at the start the ready bit would be unset, writer would reserve memory by advancing head through a CAS operation (as it is done here), but reader would not read the sample since its ready is not set yet. Upon commit writer would do an exchange on the ready bit to set it, and reader would do an exchange on it with 0 to check if sample is ready and at the same time reset it to not ready when ready. Sadly this does not work with variable-size samples :(
The other idea was to have a "pre-head" that ahead of head, reserving would advance the pre-head through CAS. Here there is not need for ready/busy state, but commit is more complex because only the writer that has reserved the sample just after head is allowed to advance head, which leads to a wait loop.

nsavoire:
Thanks @sanchda, @ivoanjo and @r1viollet for all your comments !

include/ringbuffer_utils.hpp
```diff
@@ -152,6 +154,160 @@ class PerfRingBufferReader {
...
+    MPSCRingBufferHeader *hdr = reinterpret_cast<MPSCRingBufferHeader *>(
+        buf.data() - sizeof(MPSCRingBufferHeader));
+
+    // Clear busy bit
+    uint64_t new_size = hdr->size ^ MPSCRingBufferHeader::k_busy_bit;
+    if (discard) {
+      new_size |= MPSCRingBufferHeader::k_discard_bit;
+    }
+
+    __atomic_store_n(&hdr->size, new_size, __ATOMIC_RELAXED);
```
nsavoire, line 236: Ooops, `relaxed` is not good here, since we need that all previous writes to be visible to the reader once we change `hdr->size`. Need to be changed to `release`

r1viollet (dismissed):
LGTM
An impressive PR ! Thanks for splitting this up

r1viollet (approved):
LGTM

r1viollet (dismissed):
LGTM

r1viollet (approved):
LTGM -- Thanks for adding extra tests on forking

PR #132, opened 2022-08-04
Fix for dd_profiling library not found in unwinding

r1viollet:
# What does this PR do?

- Store the file descriptor to the matching binaries to limit number of file opens on the same binary
- Store the file descriptor to the dd_profiling library when relevant
- Give libdwfl API a file descriptor instead of a path

# Motivation

As we were storing the library in a temporary file (getting deleted), we were not able to re-open the elf file when profiling the allocations. This was breaking allocation profiling.

# Additional Notes

This fixes allocation profiling when the library is not available

# How to test the change?

Allocation profiling (while deleting the dd_profiling library)

### DIFF
--- a/src/ddprof_file_info.cc
+++ b/src/ddprof_file_info.cc
@@ -5,14 +5,13 @@
 
 #include "ddprof_file_info.hpp"
 
+#include "ddres_exception.hpp"
 #include <fcntl.h>
 #include <unistd.h>
-#include "ddres_exception.hpp"
 
 #include "logger.hpp"
 
 namespace ddprof {
-
 
 bool FileInfoInodeKey::operator==(const FileInfoInodeKey &o) const {
   // default comparison is c++20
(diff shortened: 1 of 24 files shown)


src/ddprof_file_info.cc
```diff
@@ -5,11 +5,81 @@
...
 
+
 bool FileInfoInodeKey::operator==(const FileInfoInodeKey &o) const {
   // default comparison is c++20
   return _inode == o._inode && _sz == o._sz;
 }
 
+FileInfoValue::FileInfoValue(FileInfo &&info, FileInfoId_t id, int fd)
+    : _info(info), _id(id) {
+  _fd = dup(fd);
```
nsavoire, line 24: It makes me slightly uneasy to duplicate file descriptors knowing that the underlying file offset is shared.
Are we guaranteed that dwfl will independently of the file descriptor current offset ?

src/ddprof_file_info.cc
```diff
@@ -5,11 +5,81 @@
...
+}
+
+FileInfoValue::FileInfoValue(const FileInfoValue& other) {
+  copy_values(other);
+  if (other._fd >= 0) {
+    this->_fd = dup(other._fd);
+  }
+}
+
+FileInfoValue& FileInfoValue::operator=(const FileInfoValue& other) {
```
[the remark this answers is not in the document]
r1viollet: I don't think I would factorise much and I would trade a single assign for a std::swap (unless I'm missing something)

src/ddprof_file_info.cc
r1viollet: This is a very valid concern.

nsavoire (approved):
LGTM

PR #136, opened 2022-08-08
Fix libdd_profiling.a

nsavoire:
# What does this PR do?

Emebedded exe was missing from lib


r1viollet (approved):
LGTM
Should we see this in relenv ? Can we add a monitor ?

nsavoire:
> LGTM Should we see this in relenv ? Can we add a monitor ?
No, this is not tested in relenv, I am planning to add a unit test.

PR #138, opened 2022-08-09
Add an option to print samples as folded stacks

nsavoire:


include/ddprof_input.hpp
```diff
@@ -97,7 +97,8 @@ typedef struct DDProfInput {
   XX(DD_PROFILING_NATIVE_GLOBAL,        global,             g, 'g', 1, input, NULL, "", )                      \
   XX(DD_PROFILING_INTERNAL_STATS,       internal_stats,     b, 'b', 1, input, NULL, "", )                      \
   XX(DD_PROFILING_NATIVE_SOCKET,        socket,             Z, 'Z', 1, input, NULL, "", )                      \
-  XX(DD_PROFILING_NATIVE_PRESET,        preset,             D, 'D', 1, input, NULL, "", )
+  XX(DD_PROFILING_NATIVE_PRESET,        preset,             D, 'D', 1, input, NULL, "", ) \
+  XX(DD_PROFILING_NATIVE_SHOW_SAMPLES,  show_samples,       Y, 'Y', 0, input, NULL, "", )
```
sanchda, line 101: We're running out of single-character short options :)

This is a gray area, but capital letters elsewhere designate options that affect the profiler's interaction with Datadog's backend (metadata) whereas lowercase characters designate options that affect the profiler's behavior itself (incl. logging).  This is a slightly silly convention--do we want to keep it?

src/pprof/ddprof_pprof.cc
```diff
@@ -253,3 +254,41 @@ DDRes pprof_reset(DDProfPProf *pprof) {
   }
   return ddres_init();
 }
+
+void ddprof_print_sample(const UnwindOutput &uw_output,
```
sanchda, line 258: Looks good to me--sensible!

sanchda (approved):
This is great!  LGTM

r1viollet (approved):
LGTM
A good feature for tests !

include/ddprof_input.hpp
nsavoire: Sorry I was not aware of that convention !
I will make the change.
We are definitely running out of characters, it would be nice to have only long format for some options, but it seems difficult with the current design.

sanchda: Well, the convention isn't really documented anywhere and doesn't even make that much sense.

I was asking a genuine question--does that convention even make sense?  Maybe we should just get rid of this longopt nonsense and do proper input handling in a future PR?

PR #139, opened 2022-08-10
Add basic tests

nsavoire:
# What does this PR do?

* Build an executable that does lot of allocations in different modes (static/shared profiling lib, standalone).
* Perform some basic checks on profiling this exe


r1viollet (dismissed):
Yay ! New tests 👍 
I don't think I saw any usage of these options internally, so we should not break APIs.
LGTM

PR #159, opened 2022-08-22
Disable part of test only for musl libc

nsavoire:
# What does this PR do?

* Detect libc type by compiling a snippet file with verbose output and parsing -dynamic-linker option.
* Define MUSL_LIBC when musl libc is detected
* Modify savecontext-ut depending if MUSL_LIBC is defined

### DIFF
--- a/cmake/Helperfunc.cmake
+++ b/cmake/Helperfunc.cmake
@@ -20,22 +20,25 @@ function(add_exe name)
    set(exe_libraries "")
    set(exe_definitions "")
    set(exe_include_dirs "")
+
    foreach(arg IN LISTS ARGN)
-      if (arg STREQUAL "LIBRARIES")
+      if(arg STREQUAL "LIBRARIES")
          set(cur_var "libraries")
-      elseif (arg STREQUAL "DEFINITIONS")
+      elseif(arg STREQUAL "DEFINITIONS")
          set(cur_var "definitions")
       else()
          list(APPEND exe_${cur_var} ${arg})
-         if (cur_var STREQUAL "sources")
+
+         if(cur_var STREQUAL "sources")
             get_filename_component(src_dir ${arg} DIRECTORY)
             list(APPEND exe_include_dirs ${src_dir})
          endif()
       endif()
    endforeach()
+
    add_executable(${name} ${exe_sources})
    set_target_properties(${name} PROPERTIES
-       COMPILE_DEFINITIONS "${exe_definitions}")
+      COMPILE_DEFINITIONS "${exe_definitions}")
    target_link_libraries(${name} PRIVATE ${exe_libraries})
    list(REMOVE_DUPLICATES exe_include_dirs)
    target_include_directories(${name} PRIVATE ${exe_include_dirs})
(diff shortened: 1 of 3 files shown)


nsavoire:
Current dependencies on/for this PR:
* main
  * **PR #159** <a href="https://app.graphite.dev/github/pr/DataDog/ddprof/159" target="_blank">Graphite</a>  👈
    * **PR #160** <a href="https://app.graphite.dev/github/pr/DataDog/ddprof/160" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/DataDog/ddprof/159?utm_source=stack-comment).

cmake/Helperfunc.cmake
```diff
@@ -20,37 +20,58 @@ function(add_exe name)
    set(exe_libraries "")
    set(exe_definitions "")
    set(exe_include_dirs "")
+
    foreach(arg IN LISTS ARGN)
-      if (arg STREQUAL "LIBRARIES")
+      if(arg STREQUAL "LIBRARIES")
```
sanchda, line 25: We should probably settle on one style or the other, right?  Without any of your changes (just looking at main):
```
sanchda@dev:~/dev/ddprof$ grep 'if (' CMakeLists.txt | wc -l
8
sanchda@dev:~/dev/ddprof$ grep 'if(' CMakeLists.txt | wc -l
16
```

Similarly,
```
sanchda@dev:~/dev/ddprof_metrics$ grep -nri 'if (' cmake/ | wc -l
7
sanchda@dev:~/dev/ddprof_metrics$ grep -nri 'if(' cmake/ | wc -l
80
```

It doesn't have to happen in this PR, but maybe we can agree that Cmake control flow parameter braces should "touch" the statement name?  `if()` not `if ()`?

cmake/Helperfunc.cmake
```diff
@@ -20,37 +20,58 @@ function(add_exe name)
...
+
+   if(NOT STATIC_LINK_CXX_TARGET)
+      message("You must supply a TARGET to static_link_cxx")
+   endif()
+
+   target_link_options(${STATIC_LINK_CXX_TARGET} PUBLIC $<$<COMPILE_LANGUAGE:CXX>:-static-libstdc++>)
+   target_link_options(${STATIC_LINK_CXX_TARGET} PUBLIC $<$<COMPILE_LANGUAGE:CXX>:-static-libgcc>)
+endfunction()
+
+function(detect_libc output_variable)
```
sanchda, line 63: Right.  I think the only reliable way to really predict which libc we're dealing with is to build and then see what happened--so this seems reasonable.

sanchda (approved):
See note about cmake stylistic consistency, but that's just a comment for us to think about rather than anything that needs to be done in this PR.

LGTM!

cmake/Helperfunc.cmake
nsavoire: I agree !
I might just have done a `Format Document` in VS Code (which is provided by [CMake Language Support](https://marketplace.visualstudio.com/items?itemName=josetr.cmake-language-support-vscode) extension I think).
I also found another extension ([cmake-format](https://marketplace.visualstudio.com/items?itemName=cheshirekow.cmake-format)) based on a CLI tool, which might be better suited (since we could run cmake formatting checks in CI build).

PR #164, opened 2022-08-25
Fix Dockerfile ADD statements

sanchda:
# What does this PR do?
Dockerfile `ADD` statements have their relative paths expanded based on the position of the Dockerfile, not based on the current working directory when `docker build` is invoked.  This patch fixes the ADD statements in the Alpine Dockerfile, which had an over-qualified relative path fragment.

# Motivation
The Alpine Dockerfile is really useful; it's easy to build this image and then launch a local dev environment based on it.  I'd like to keep it for musl testing.

### DIFF
--- a/app/base-env-alpine/Dockerfile
+++ b/app/base-env-alpine/Dockerfile
@@ -9,7 +9,8 @@ ENV CXX=g++
 # libjemalloc
 # libexecinfo-dev # No static lib
 RUN apk update \
-    && apk add --no-cache aws-cli \
+    && apk add --no-cache \
+       aws-cli \
        bash \
        binutils-dev \
        ca-certificates \
--- a/app/base-env-alpine/Dockerfile
+++ b/app/base-env-alpine/Dockerfile
@@ -100,6 +100,6 @@ RUN VERSION="1.6.1" \
 
 # Patch to fix compilation of elfutils
 RUN mkdir /patch
-ADD ./app/base-env-alpine/*.patch /patch/
-ADD ./app/base-env-alpine/error.h /patch/
-ADD ./app/base-env-alpine/libintl.h /patch/
+ADD *.patch /patch/
+ADD error.h /patch/
+ADD libintl.h /patch/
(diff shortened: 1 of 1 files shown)


app/base-env-alpine/Dockerfile
```diff
@@ -100,6 +101,6 @@ RUN VERSION="1.6.1" \
 
 # Patch to fix compilation of elfutils
 RUN mkdir /patch
-ADD ./app/base-env-alpine/*.patch /patch/
-ADD ./app/base-env-alpine/error.h /patch/
-ADD ./app/base-env-alpine/libintl.h /patch/
+ADD *.patch /patch/
```
nsavoire, line 104: For this to work, you will need to change how `docker build` is invoked in CI from to `docker build -f ./app/base-env/Dockerfile .` to `docker build ./app/base-env` (and also make sure that not `ADD` statement requiring files from other dirs are present).

sanchda: Good catch, I see it's not quite working properly.  Obviously the `-f` flag changes the behavior of path expansion compared to just giving `docker build` a path.

I'm going to scratch my head over this a little bit.  I'll probably just retract this PR if I can't think of a compelling reason to use `Docker build` over `docker build -f` (thinking about it in the moment, I understand that changing the semantics of path expansion might be useful, but I am confused why the `docker build` CLI would bundle it into the `-f` parameter instead of making it a distinct option).

nsavoire: The positional path given to docker build (currently `.` in the CI) gives the context that is sent to the docker daemon (all paths in COPY/ADD statements are relative to this path and COPY/ADD cannot refer to paths outside this context). By default docker build will use the `Dockerfile` present in this path, except if `-f` options is given, which will be used in that case.
Usually all files required by `Dockerfile` are put in subdirectories of `Dockerfile` directory, and `-f` is useless.
`-f` is useful when a file is required by different `Dockerfile`s:
- img1 
  - Dockerfile
- img2
  - Dockerfile
- common_files
  - file1

sanchda: Ah, that makes a lot more sense--thank you for the explanation!

PR #167, opened 2022-08-25
Add pragmaonce to a header

sanchda:
We've been adding `#pragma once` to all headers, even those which are called by only a single source file, since it's annoying ti fix transitive issues downstream.  Noticed one header was missing the pragma, so here's the fix.


nsavoire (approved):
LGTM !

PR #168, opened 2022-08-26
Sanchda/use gcc 12

sanchda:
# What does this PR do?

As per a previous PR comment from @nsavoire, it's not a bad idea to include gcc-12 support in the environment setup script.  However, gcc-12 (and clang-13) are less permissive around the `find()` invocation in `src/unwind.cc` which requires the inclusion of `<algorithm>`.

This PR
* Adds gcc-12 defaulting to the setup script
* Fixes a typo I introduced in a previous PR (oops)
* Adds `<algorithm>` to `src/unwind.cc`

### DIFF
--- a/setup_env.sh
+++ b/setup_env.sh
@@ -8,8 +8,8 @@ export PATH=$PATH:${PWD}/tools:${PWD}/bench/runners
 # SanCMake ../ 
 
 # Attempt to use the latest known working version of GCC as the default
-DDPROF_CC_DEFAULT=$(command -v gcc{-11,-10,-9,} | head -n1)
-DDPROF_CXX_DEFAULT=$(command -v g++{-11,-10,-9,} | head -n1)
+DDPROF_CC_DEFAULT=$(command -v gcc{-12,-11,-10,-9,} | head -n1)
+DDPROF_CXX_DEFAULT=$(command -v g++{-12,-11,-10,-9,} | head -n1)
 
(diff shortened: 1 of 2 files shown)


src/unwind.cc
```diff
@@ -17,6 +17,7 @@
 #include "unwind_metrics.hpp"
 #include "unwind_state.hpp"
 
+#include <algorithm>
```
r1viollet, line 20: Was this intended, or is this part of another change ?

setup_env.sh
```diff
@@ -8,8 +8,8 @@ export PATH=$PATH:${PWD}/tools:${PWD}/bench/runners
 # SanCMake ../ 
 
 # Attempt to use the latest known working version of GCC as the default
-DDPROF_CC_DEFAULT=$(command -v gcc-{-11,-10,-9,} | head -n1)
-DDPROF_CXX_DEFAULT=$(command -v g++-{-11,-10,-9,} | head -n1)
+DDPROF_CC_DEFAULT=$(command -v gcc{-12,-11,-10,-9,} | head -n1)
+DDPROF_CXX_DEFAULT=$(command -v g++{-12,-11,-10,-9,} | head -n1)
```
r1viollet, line 12: Thanks, this also includes a fix (as things were slightly broken right now)

r1viollet (approved):
LGTM 
Edit: Ok this was intended, read the commit messages a little too late
Thanks for the fixes

src/unwind.cc
r1viollet: Ah, yeah, intended, my bad.

PR #169, opened 2022-09-02
Allow custom symbols to be published by runtimes

r1viollet:
# What does this PR do?

Allow publication of custom symbols by runtimes like .NET. We look for the map when we attempt to symbolise an anonymous region.

When a file with symbols (perf map file) matching this pid is found within `/tmp`, we will consider the symbols within the map.
Some limitations:
- The map is loaded a single time. We do not refresh the map when there are changes
- When we do not find the map, we add a fake symbol to make sure we don't always look for this file.

# Motivation

This way .NET can push their symbols to us.

# Additional Notes

A further step would be to ensure we can have the symbolisation step in backend

# How to test the change?

I added a unit test for all the parsing of the map.
It would be nice to have dedicated .NET tests within the CI.


src/unwind_dwfl.cc
```diff
@@ -87,7 +88,35 @@ static void trace_unwinding_end(UnwindState *us) {
...
+  RuntimeSymbolLookup &runtime_symbol_lookup =
+      unwind_symbol_hdr._runtime_symbol_lookup;
+  SymbolIdx_t symbol_idx =
+      runtime_symbol_lookup.get_or_insert(dso._pid, pc, symbol_table);
+  if (symbol_idx == -1) {
+    add_dso_frame(us, dso, pc, "pc");
+    return ddres_init();
+  }
+  UnwindOutput *output = &us->output;
+  int64_t current_loc_idx = output->nb_locs;
```
nsavoire, line 107: Check for `UnwindOutput::locs` overflow is missing, the code for adding a new location could probably be factorized.

src/dso.cc
```diff
@@ -59,7 +61,10 @@ Dso::Dso(pid_t pid, ElfAddress_t start, ElfAddress_t end, ElfAddress_t pgoff,
              // ends with .jsa
              ((_filename.length() > s_jsa_str.length() + 1) &&
               _filename.substr(_filename.length() - s_jsa_str.length(),
-                               _filename.length()) == s_jsa_str)) {
```
nsavoire, line 62: Why not create a new DSO type for these files ?

src/runtime_symbol_lookup.cc
```diff
@@ -0,0 +1,128 @@
...
+RuntimeSymbolLookup::RuntimeSymbolFindRes
+RuntimeSymbolLookup::find_closest(SymbolMap &map, ProcessAddress_t pc) {
+  bool is_within = false;
+
+  // First element not less than (can match exactly a start addr)
+  auto it = map.lower_bound(pc);
+  if (it != map.end()) { // map is empty
+    is_within = symbol_is_within(pc, *it);
+    if (is_within) {
+      return std::make_pair<SymbolMap::iterator, bool>(std::move(it),
```
nsavoire, line 95: No need to move iterator / bool.

include/runtime_symbol_lookup.hpp
```diff
@@ -0,0 +1,47 @@
...
+
+#include "ddprof_defs.hpp"
+#include "symbol_table.hpp"
+
+#include <map>
+#include <unordered_map>
+
+namespace ddprof {
+
+class RumtimeSymbolVal {
```
nsavoire, line 16: `Runtime` instead of `Rumtime`.
Not sure it's the best name, why not reuse `DwflSymbolVal` ?

include/runtime_symbol_lookup.hpp
```diff
@@ -0,0 +1,47 @@
...
+public:
+  RumtimeSymbolVal(ProcessAddress_t end, SymbolIdx_t symbol_idx)
+      : _end(end), _symbol_idx(symbol_idx) {}
+  Offset_t get_end() const { return _end; }
+
+  SymbolIdx_t get_symbol_idx() const { return _symbol_idx; }
+
+private:
+  // symbol end within the segment (considering file offset)
```
nsavoire, line 25: Comment seems wrong

src/runtime_symbol_lookup.cc
```diff
@@ -0,0 +1,128 @@
...
+    return std::make_pair<SymbolMap::iterator, bool>(map.end(), false);
+  }
+  // element can not be end (as we reversed or exit)
+  is_within = symbol_is_within(pc, *it);
+
+  return std::make_pair<SymbolMap::iterator, bool>(std::move(it),
+                                                   std::move(is_within));
+}
+
+bool RuntimeSymbolLookup::symbol_is_within(ProcessAddress_t pc,
```
nsavoire, line 113: Duplicated code from `DwflSymbolLookup_V2::dwfl_symbol_is_within` ?

src/runtime_symbol_lookup.cc
```diff
@@ -0,0 +1,128 @@
...
+
+  RuntimeSymbolFindRes find_res = find_closest(symbol_map, pc);
+  if (find_res.second) {
+    return find_res.first->second.get_symbol_idx();
+  }
+  return -1;
+}
+
+RuntimeSymbolLookup::RuntimeSymbolFindRes
+RuntimeSymbolLookup::find_closest(SymbolMap &map, ProcessAddress_t pc) {
```
nsavoire, line 87: Duplicated code from `DwflSymbolLookup_V2::find_closest` ?

src/runtime_symbol_lookup.cc
```diff
@@ -0,0 +1,128 @@
...
+      : address{addr}, code_size{cs}, function_name{std::move(fn)} {}
+
+  std::uint64_t address;
+  std::size_t code_size;
+  std::string function_name;
+};
+
+FILE *RuntimeSymbolLookup::perfmaps_open(int pid, const char *path_to_perfmap = "") {
+  char buf[1024] = {0};
+  auto n = snprintf(buf, 1024, "%s/proc/%d/mount%s/perf-%d.map", _path_to_proc.c_str(), pid, path_to_perfmap, pid);
```
nsavoire, line 20: What is `/proc/<pid>/mount` ?

src/runtime_symbol_lookup.cc
```diff
@@ -0,0 +1,128 @@
...
+
+  LG_DBG("Loading runtime symbols from (PID%d)", pid);
+  char *line = NULL;
+  size_t sz_buf = 0;
+  char buffer[2048];
+  auto it = symbol_map.end();
+  while (-1 != getline(&line, &sz_buf, pmf)) {
+    uint64_t address;
+    uint32_t code_size;
+    if (3 != sscanf(line, spec, &address, &code_size, buffer) ||
```
nsavoire, line 55: I find more readable to have scanf spec inlined in the function call

nsavoire, line 55: `%[^\n\t]` seems unsafe since it could overflow `buffer`

src/runtime_symbol_lookup.cc
```diff
@@ -0,0 +1,128 @@
...
+        should_skip_symbol(buffer)) {
+      continue;
+    }
+    // elements are ordered
+    it = symbol_map.emplace_hint(
+        it, address,
+        RumtimeSymbolVal(address + code_size - 1, symbol_table.size()));
+    symbol_table.emplace_back(
+        Symbol(std::string(buffer), std::string(buffer), 0, "unknown"));
+  }
```
nsavoire, line 65: free of `line` is missing

src/dso.cc
r1viollet: I guess it is slightly confusing.
I wanted to make sure that any type of symbols managed by runtimes could be overridden. 
The cleaner way could be to have 2 types, and allow both to default to the perf-map symbols.

include/runtime_symbol_lookup.hpp
r1viollet: I will propose `SymbolSpan` if that is ok.

r1viollet:
Thanks for the thorough review. I was essentially trying to figure out if this was a good direction. 
I think having a mechanism for this is nice, so I'll clean up PR as per reviews.

src/runtime_symbol_lookup.cc
r1viollet: CI agrees with this 👍

sanchda (approved):
LGTM.  This will require a little bit of generalization in the future, but the "tech debt" is constrained to this feature and of no major consequence.

include/dso_type.hpp
```diff
@@ -15,6 +15,7 @@ enum DsoType {
   kHeap,
   kUndef,
   kAnon,
+  kRuntime,
```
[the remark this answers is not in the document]
r1viollet: +1, I was thinking of introducing attributes. Then I thought it might be overkill.

include/dso_type.hpp
```diff
@@ -29,6 +30,12 @@ static inline bool has_relevant_path(dso::DsoType dso_type) {
   }
   return false;
 }
+
+// some runtimes such as java or .NET can publish maps to populate the symbols
+static inline bool has_runtime_symbols(dso::DsoType dso_type) {
```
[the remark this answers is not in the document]
r1viollet: I'm really waiting for new use cases. I agree this does not cover all possibilities. Mostly the .NET use case.
As soon as we have new use cases I will be happy to come back to this.

src/dso.cc
```diff
@@ -23,6 +23,8 @@ static const std::string_view s_anon_str = "//anon";
 static const std::string_view s_jsa_str = ".jsa";
 // Example of these include : anon_inode:[perf_event]
 static const std::string_view s_anon_inode_str = "anon_inode";
+// dll should not be considered as elf files
```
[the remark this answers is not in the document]
r1viollet: Indeed. I'm not very comfortable with this.
We could open all the binaries to check for headers.

sanchda (approved):
review bump:  still LGTM

src/dso.cc
```diff
@@ -55,12 +57,16 @@ Dso::Dso(pid_t pid, ElfAddress_t start, ElfAddress_t end, ElfAddress_t pgoff,
...
              _filename.substr(0, s_dev_zero_str.length()) == s_dev_zero_str ||
-             _filename.substr(0, s_dev_null_str.length()) == s_dev_null_str ||
-             // ends with .jsa
-             ((_filename.length() > s_jsa_str.length() + 1) &&
-              _filename.substr(_filename.length() - s_jsa_str.length(),
-                               _filename.length()) == s_jsa_str)) {
+             _filename.substr(0, s_dev_null_str.length()) == s_dev_null_str) {
     _type = dso::kAnon;
+  } else if ( // ends with .jsa
+      ((_filename.length() > s_jsa_str.length() + 1) &&
```
nsavoire, line 63:
```c++
      _filename.ends_with(s_jsa_str) || ...
```

src/dso_hdr.cc
```diff
@@ -154,7 +154,7 @@ DsoFindRes DsoHdr::dso_find_first_std_executable(pid_t pid) {
   if (it == map.end()) {
     return find_res_not_found(map);
   }
-  return std::make_pair<DsoMapConstIt, bool>(std::move(it), true);
+  return std::pair(it, true);
```
nsavoire, line 157:
```suggestion
  return {t, true};
```

src/common_symbol_lookup.cc
```diff
@@ -35,8 +35,8 @@ SymbolIdx_t CommonSymbolLookup::get_or_insert(SymbolErrors lookup_case,
   } else { // insert things
     symbol_idx = symbol_table.size();
     symbol_table.push_back(symbol_from_common(lookup_case));
-    _map.insert(std::pair<SymbolErrors, SymbolIdx_t>(lookup_case, symbol_idx));
+    _map.insert(std::pair(lookup_case, symbol_idx));
```
nsavoire, line 38:
```suggestion
    _map.insert({lookup_case, symbol_idx});
```

src/common_mapinfo_lookup.cc
```diff
@@ -27,8 +27,7 @@ MapInfoIdx_t CommonMapInfoLookup::get_or_insert(
   } else { // insert things
     res = mapinfo_table.size();
     mapinfo_table.push_back(mapinfo_from_common(lookup_case));
-    _map.insert(std::pair<CommonMapInfoLookup::MappingErrors, MapInfoIdx_t>(
-        lookup_case, res));
+    _map.insert(std::pair(lookup_case, res));
```
nsavoire, line 30:
```suggestion
    _map.insert({lookup_case, res});
```

src/dso_hdr.cc
```diff
@@ -165,8 +165,7 @@ DsoFindRes DsoHdr::dso_find_closest(const DsoMap &map, pid_t pid,
   if (it != map.end()) {
     is_within = it->second.is_within(pid, addr);
     if (is_within) { // exact match
-      return std::make_pair<DsoMapConstIt, bool>(std::move(it),
-                                                 std::move(is_within));
+      return std::pair(it, is_within);
```
nsavoire, line 168:
```suggestion
      return{it, is_within};
```

src/runtime_symbol_lookup.cc
```diff
@@ -0,0 +1,110 @@
...
+    char address_buff[33]; // max size of 16 (as it should be hexa for uint64)
+    char size_buff[33];
+    // Avoid considering any symbols beyond 300 chars
+    if (3 !=
+            sscanf(line, "%16s %8s %300[^\t\n]", address_buff, size_buff,
+                   buffer) ||
+        should_skip_symbol(buffer)) {
+      continue;
+    }
+    ProcessAddress_t address = std::strtoul(address_buff, nullptr, 16);
```
nsavoire, line 71: Success of strtoul is not checked.
why not use scanf to do the conversion ?

include/dso_hdr.hpp
```diff
@@ -121,12 +121,12 @@ class DsoHdr {
   static Dso dso_from_procline(int pid, char *line);
 
   static DsoFindRes find_res_not_found(const DsoMap &map) {
-    return std::make_pair<DsoMapConstIt, bool>(map.end(), false);
+    return std::pair(map.end(), false);
```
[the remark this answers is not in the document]
sanchda: Whoopsies!  Thank you, I also find copy-list-initialization to be much cleaner here.

src/dso_hdr.cc
sanchda: Actually should be
```suggestion
return {it, true};
```

sanchda:
Hey @nsavoire and @r1viollet, I pushed dcfb88b which replaces the `return std::pair()` pattern with `return {...}`.  I also replaced two instances of implicit `std::string()` in `tags.cc`, which is a separate matter of taste (I didn't have the forethought to separate into two commits for easy reversal, though).

@nsavoire already provided this suggestion in all of the right places, but I checked for a few more of them in code and ended up having a commit handy.  Please feel free to revert or modify, I was only trying to get the suggestions through CI for convenience.

src/runtime_symbol_lookup.cc
r1viollet: Compiler was nagging :-)
You are correct though. I'll check for success.

src/dso.cc
r1viollet: I guess we embrace c++ 20 fully ? 😄

sanchda (approved):
Since the last review, several streamline calls, movement away from V2 dwfl_symbol_lookup, nice clarification of the dso.cc checks, additional fencing in runtime_symbol_lookup.cc--looks good to me!

PR #172, opened 2022-09-13
Small fixes

nsavoire:


sanchda (approved):
LGTM, thank you for these fixes!

cmake/CodeCoverage.cmake
```diff
@@ -1,3 +1,4 @@
+# cmake-format: off
```
sanchda, line 1: Good call.  I believe we're in the right to do it this way and I can't find a convenient way to ignore certain files from a cmake-format configuration file.

cmake/Jemalloc.cmake
```diff
@@ -1,47 +1,35 @@
-# Source : https://github.com/c9s/r3/blob/2.0/cmake/Modules/FindJemalloc.cmake
-# Licence : MIT license
+# Source : https://github.com/c9s/r3/blob/2.0/cmake/Modules/FindJemalloc.cmake Licence : MIT license
```
sanchda, line 1: This is another case where we're pulling in code, but we don't spare it from cmake-format.  We should probably try to be good citizens by keeping the substance of the copied code unmodified.

sanchda (dismissed):
LGTM.  Left one advisory comment, but it's open for discussion and doesn't need to block merging.

cmake/Jemalloc.cmake
```diff
@@ -1,3 +1,4 @@
+# cmake-format: off
```
sanchda, line 1: nice

sanchda (approved):
LGTM

PR #173, opened 2022-09-13
Add build-id information to the pprof mappings

r1viollet:
# What does this PR do?

Ensure we send build-id information in the pprof data

# Motivation

Allow backend to symbolize the pprofs remotely 

# Additional Notes

N/A

# How to test the change?

I tested manually. This would be a relevant test to add.


include/build_id.hpp
```diff
@@ -0,0 +1,16 @@
...
+// developed at Datadog (https://www.datadoghq.com/). Copyright 2021-Present
+// Datadog, Inc.
+
+#include <span>
+#include <vector>
+#include <string>
+
+namespace ddprof {
+using BuildIdSpan = std::span<const unsigned char>;
+using BuildId = std::vector<unsigned char>;
```
nsavoire, line 12: `BuildId` does not appear to be used

include/ddprof_module.hpp
```diff
@@ -29,6 +31,12 @@ struct DDProfMod {
 
   explicit DDProfMod(Status status) : DDProfMod() { _status = status; }
 
+  void set_build_id(std::span<const unsigned char> x) {
```
nsavoire, line 34: Why not accept directly a `BuildIdStr` and do the conversion when retrieving it from dwfl ?

include/mapinfo_lookup.hpp
```diff
@@ -20,7 +20,7 @@ namespace ddprof {
 class MapInfoLookup {
 public:
   MapInfoIdx_t get_or_insert(pid_t pid, MapInfoTable &mapinfo_table,
-                             const Dso &dso);
+                             const Dso &dso, const BuildIdStr *build_id = nullptr);
```
nsavoire, line 23: Maybe use an `std::optional` if the argument is truly optional

src/unwind.cc
```diff
@@ -111,7 +111,7 @@ DDRes unwindstate__unwind(UnwindState *us) {
 void unwind_pid_free(UnwindState *us, pid_t pid) {
   us->dso_hdr.pid_free(pid);
```
nsavoire, line 112: Seeing this makes me wonder if having a single mapping `pid -> {dso, symbols, dwfl}` instead would make more sense.

include/ddog_profiling_utils.hpp
```diff
@@ -12,6 +12,11 @@ extern "C" {
 #include "string_view.hpp"
 
 #include <string_view>
+#include <span>
+
+inline ddog_CharSlice to_CharSlice(std::span <const unsigned char>span) {
```
nsavoire, line 17: This function seems useless (buildid is a `std::string` in MapInfo, and therefore `ddog_CharSlice to_CharSlice(std::string_view str)` should be able to convert it to a slice).

include/ddprof_module.hpp
r1viollet: I don't like having it as a string and still want to leave the path open to have this as a number

r1viollet:
Ready to review 👍

include/build_id.hpp
```diff
@@ -0,0 +1,17 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0. This product includes software
+// developed at Datadog (https://www.datadoghq.com/). Copyright 2021-Present
+// Datadog, Inc.
+
+#include "span.hpp"
+
+#include <string>
+#include <vector>
```
nsavoire, line 9: **🤓 nitpick:** useless include‏

include/build_id.hpp
```diff
@@ -0,0 +1,17 @@
...
+
+namespace ddprof {
+using BuildIdSpan = ddprof::span<const unsigned char>;
```
nsavoire, line 12: **🤓 nitpick:** ‏you might want to use `const std::byte` instead of `const unsigned char` to make explicit that these are not characters

include/ddprof_module.hpp
```diff
@@ -29,6 +31,12 @@ struct DDProfMod {
 
   explicit DDProfMod(Status status) : DDProfMod() { _status = status; }
 
+  void set_build_id(ddprof::span<const unsigned char> x) {
```
nsavoire, line 34:
```suggestion
  void set_build_id(BuildIdSpan x) {
```

src/build_id.cc
```diff
@@ -0,0 +1,25 @@
...
+#include "build_id.hpp"
+
+#include <iomanip>
+
+namespace ddprof {
+
+BuildIdStr format_build_id(BuildIdSpan build_id_span) {
+  std::string build_id_str;
+  // We encode every char on 2 hexa string characters
+  build_id_str.resize(build_id_span.size() * 2 + 1);
```
nsavoire, line 15: Why `+1` ?

src/build_id.cc
```diff
@@ -0,0 +1,25 @@
...
+  for (unsigned i = 0; i < build_id_span.size(); ++i) {
+    snprintf(&build_id_str[static_cast<unsigned>(2 * i)], 3, "%02x",
```
nsavoire, line 17: **🤓 nitpick:** ‏No sure `snprintf` is needed here since `%02x` completely defines the output length

src/ddprof_module_lib.cc
```diff
@@ -148,6 +148,16 @@ DDRes report_module(Dwfl *dwfl, ProcessAddress_t pc, const Dso &dso,
...
       dwfl_report_elf(dwfl, module_name, filepath.c_str(), fd, start, false);
 
+  // Retrieve build id
+  const unsigned char *bits = nullptr;
+  GElf_Addr vaddr;
+  if (int size = dwfl_module_build_id(ddprof_mod._mod, &bits, &vaddr);
+      size > 0) {
+    // ensure we called dwfl_module_getelf first (or this can fail)
+    // returns the size
+    ddprof_mod.set_build_id(ddprof::span(bits, size));
```
nsavoire, line 158:
```suggestion
    ddprof_mod.set_build_id(BuildIdSpan{bits, size});
```

nsavoire (commented):
LGTM !

nsavoire (approved):

src/build_id.cc
r1viollet: snprintf adds an extra null termination

nsavoire: Is it a problem? It would overwrite the existing null terminator of `build_id_str` and this is ok.
From https://en.cppreference.com/w/cpp/string/basic_string/operator_at:

> If pos == size(), a reference to the character with value CharT() (the null character) is returned.
For the first (non-const) version, the behavior is undefined if this character is modified to any value other than CharT() .

include/build_id.hpp
r1viollet: Thanks, c++ 17 things !

nsavoire: What went wrong ?

PR #175, opened 2022-09-15
Avoid closing std out on exit

r1viollet:
# What does this PR do?

Avoid closing the wrong file descriptor

# Motivation

It sounds like a good idea

# Additional Notes

N/A

# How to test the change?

I'm not sure


nsavoire (approved):
LGTM !

PR #176, opened 2022-09-19
R1viollet/centos7

r1viollet:
# What does this PR do?

Build on older libc

# Motivation

Get a version to help with agent leak issues


app/base-env-centos/scl_enable
```diff
@@ -0,0 +1,2 @@
+unset BASH_ENV PROMPT_COMMAND ENV
+source scl_source enable devtoolset-11
```
sanchda, line 2: Ah right, you do need this type of thing on rhel-variants--good catch.

sanchda (approved):
Looks like a pretty straight translation of the Ubuntu Dockerfile for centos7.  Unlike Alpine, can't just defer to what's in the repo since centos7 is so old.

I didn't check SHAs, but LGTM.

r1viollet:
Closing for now in favour of the musl strategy.

PR #178, opened 2022-09-21
mmap profiling

nsavoire:


r1viollet (approved):
LGTM

include/perf.hpp
```diff
@@ -17,7 +17,7 @@
 // defaut ring buffer size expressed as a power-of-two in number of pages
 #define DEFAULT_BUFF_SIZE_SHIFT 6
 // this does not count as pinned memory, use a larger size
-#define MPSC_BUFF_SIZE_SHIFT 7
+#define MPSC_BUFF_SIZE_SHIFT 8
```
sanchda, line 20: This doubles the size of the ringbuffer, but that should be fine since it's not pinned memory and allocations might be frequent.

sanchda (approved):
Very elegant, very nice!

PR #181, opened 2022-09-22
Create unwinding tables 

r1viollet:
# What does this PR do?

Ongoing work to check if the async profiler's libraries would make sense in terms of performance and object model
Your probably don't want to read this

# Motivation

Checking performances of async profiler's unwinding


src/async-profiler/cgotraceback.cpp
```diff
@@ -0,0 +1,223 @@
+#include <cstring>
```
nsrip-dd, line 1: I think you can delete this file and cgotraceback.go, they're specifically for my Go library :)

r1viollet:
Closing in favour of : https://github.com/DataDog/ddprof/pull/275

PR #183, opened 2022-09-26
Leak detection: Avoid unbounded growth of address map

r1viollet:
# What does this PR do?

Minor update to clear the map at 200000 elements

# Notes

To be improved: The sync is not guaranteed between lib and worker.


r1viollet:
Closing as we made a better version of this.

PR #186, opened 2022-09-27
Add tag information to downstream pipeline

r1viollet:
# What does this PR do?

Add tag information to downstream pipeline

# Motivation

The idea is that we can trigger benchmarks when tags are placed


nsavoire (approved):
LGTM

PR #187, opened 2022-09-28
Switch user if failing to read /proc/<pid>/maps

nsavoire:


r1viollet:
I think we can close this one ? As we are running as actual root on the customer infra.

nsavoire:
> I think we can close this one ? As we are running as actual root on the customer infra.

Indeed, and functionality is

PR #192, opened 2022-09-30
Add system allocation profiling

r1viollet:
# What does this PR do?

Adds support for system allocation profiling using the new `tALLOCSYS1` event type.
* Adds an override (`is_overloaded` in PerfWatcherOptions) which allows watchers to receive special post-processing in `watcher_from_event`.  This is useful for watchers whose setup might have runtime requirements or nontypical complexities.
* Adds linked file descriptor support
* Adds example using sys_exit kprobes to perform analysis

# Motivation

Another tool for helping locate memory leaks

# Discussion

An early attempt at this PR added a simple profile type (a wrapper over `raw_syscalls:sys_exit`) and a processing hook in the worker.  This was insufficient because the kernel would copy the stack into the ringbuffer for every syscall, and on a system where nthreads >= system_cores, this could easy prevent the profiler from getting on-cpu often enough not to drop a ton of events.  Because "live" allocation analysis requires knowledge of state transitions, this wasn't working for the target application.

Splitting that event into individual syscall-events is an idea, but because of our ringbuffer visitation strategy the events would have to be ordered.  There are many ways of doing this, but I didn't want to do any of them upside_down_face.jpeg

@nsavoire helpfully reminded me of the `ioctl()` for sharing multiple `perf_event_open()` file descriptors, which is the direction I chose.  One complication is that the `ioctl()` has to be emitted *after* the `mmap()` has happened on the first fd on a given CPU, so I needed to add a yucky mechanism for propagating that information, but it's very easy to repair.  I omit such a fix in this PR since #70 will change several other aspects of watcher substantially, and it's nice not to overtighten the object model right before a big change.


r1viollet:
Tracepoints from David

image

include/perf.hpp
```diff
@@ -15,7 +15,7 @@
 #include <vector>
 
 // defaut ring buffer size expressed as a power-of-two in number of pages
-#define DEFAULT_BUFF_SIZE_SHIFT 6
+#define DEFAULT_BUFF_SIZE_SHIFT 8
```
sanchda, line 18: Debugging operation.  Will remove.

src/perf.cc
```diff
@@ -28,7 +28,7 @@
 static long s_page_size = 0;
 
 struct perf_event_attr g_dd_native_attr = {
-    .size = sizeof(struct perf_event_attr),
+    .size = PERF_ATTR_SIZE_VER5,
```
sanchda, line 31: Tools like strace seem to process `perf_event_open()` better when we stick to the documented attribute size macro rather than use `sizeof()`

src/perf_ringbuffer.cc
```diff
@@ -214,12 +219,20 @@ perf_event_sample *hdr2samp(const perf_event_header *hdr, uint64_t mask) {
...
+  }
   if (PERF_SAMPLE_BRANCH_STACK & mask) {}
   if (PERF_SAMPLE_REGS_USER & mask) {
     sample.abi = *buf++;
-    // In case regs are not available, ignore this sample
-    if (sample.abi == PERF_SAMPLE_REGS_ABI_NONE) {
+    // Since some of the previous events are variable-length, ABI gives us
+    // an opportunity to sanity-check the buffer (only 2 legal values!)
+    if (sample.abi != PERF_SAMPLE_REGS_ABI_32 &&
+        sample.abi != PERF_SAMPLE_REGS_ABI_64) {
```
sanchda, line 235: I cherry-picked this from PR #70, but I forgot if we found it an agreeable change.  This is a sanity check since the interface only provides two legal values.

src/unwind.cc
```diff
@@ -126,4 +126,12 @@ void unwind_cycle(UnwindState *us) {
...
 }
 
+//bool read_dso(UnwindState *us, ProcessAddress_t addr, unsigned char *buf, size_t sz) {
+//  DsoHdr::DsoFindRes find_res = us->dso_hdr.dso_find_closest(us->pid, us->current_ip);
+//  if (!find_res.second)
+//    return false;
+//  Dso *dso = &find_res.first->second;
+//}
+
+
```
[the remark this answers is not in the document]
sanchda: Removed.

include/live_sysallocations.hpp
```diff
@@ -0,0 +1,87 @@
...
+    // Convert addr to page idx, then page-align size and decimate
+    uintptr_t page_start = to_page(addr);
+    uintptr_t page_end = to_page(addr + size);
+
+    for (auto i = page_start; i <= page_end; ++i) {
+      stack_map[i] = stack;
+    }
+  }
+
+  void move_allocs(uintptr_t addr0, uintptr_t addr1, size_t size, pid_t pid) {
```
sanchda, line 31: This is unused for now, but I'd like to keep it for a while since it isn't 100% clear to me that bulk movement (delete old + add new) is appropriate.

src/ddprof_cmdline.cc
```diff
@@ -78,6 +78,11 @@ bool watcher_from_event(const char *str, PerfWatcher *watcher) {
     watcher->tracepoint_name = watcher->desc;
     watcher->tracepoint_group = event_groupname;
   }
+
+  // Certain watcher configs get additional event information
```
sanchda, line 82: @r1viollet You had inlined this in the worker code in your PR.  I moved it here so it's more of an upfront configuration change.  Curious to learn whether this feels like a better change to you.

src/ddprof_cmdline.cc
```diff
@@ -136,11 +141,10 @@ bool watcher_from_tracepoint(const char *_str, PerfWatcher *watcher) {
   // period is a number
   char *str = strdup(_str);
   size_t sz_str = strlen(str);
-  const char *groupname;
-  const char *tracename;
+  const char *gname;
```
sanchda, line 144: Don't stare _too_ hard at this `watcher_from_tracepoint()` stuff.  It's all going completely away when #70 is merged.

src/ddprof_cmdline.cc
```diff
@@ -224,30 +240,33 @@ bool watcher_from_tracepoint(const char *_str, PerfWatcher *watcher) {
...
   }
 
   // OK done
   *watcher = *twatcher_default();
   watcher->config = trace_id;
   watcher->sample_period = period;
-  if (is_raw)
-    watcher->sample_type |= PERF_SAMPLE_RAW;
+
+  // das221004 the raw field is a very small part of the overall event and
```
sanchda, line 258: This is unfortunate, but I didn't want to fix it in this PR.  Should only affect tracepoints.

include/live_sysallocations.hpp
```diff
@@ -0,0 +1,94 @@
...
+    // Convert addr to page idx, then page-align size and decimate
+    uintptr_t page_start = to_page(addr);
+    uintptr_t page_end = to_page(addr + size);
+
+    for (auto i = page_start; i <= page_end; ++i) {
+      stack_map.erase(i);
+    }
+  }
+
+  void do_mmap(const UnwindOutput &stack, uintptr_t addr, size_t size,
```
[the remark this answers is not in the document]
sanchda: Thank you for looking--you mean errors in the allocation function itself (in this case `mmap()`)?

If so, then it's slightly strange, but this is guarded for in src/ddprof_worker.cc lines 390-393.

sanchda:
Merging into Erwan's branch

PR #194, opened 2022-10-06
Add more m(un)map overrides

nsavoire:


sanchda (approved):
Ah, let me see if I have this right.  Some glibc functions, for some versions of glibc, will call certain glibc private `__`-prefixed symbols, which are present in the .so even if they aren't exported as the standard API.

I didn't check (sorry...), but e.g., for `mmap()`; is `mmap()` a wrapper over `__mmap()` when both exist, and if so will we double-count things?

If not, no need to bother with this comment--otherwise LGTM!

PR #196, opened 2022-10-11
Replace head with awk in CI

sanchda:
# What does this PR do?

The `setup_env.sh` file is usually `source`'d by CI during the build process of the profiler.  The `command ... | head -n1` sequence is problematic, because it is common practice in CI to execute bash scripts with `pipefail` and `head -n1` _by design_ terminates as soon as it reads one line.

Instead, it is preferable to use `awk`, which will wait for the entire input stream before terminating.

# Motivation

Make things better for downstream builds!

# Additional Notes

I don't have proof that this actually works.  I'll update the PR later maybe with some evidence.


nsavoire (dismissed):
LGTM !

nsavoire (approved):

PR #203, opened 2022-11-08
Leak analysis - aggregate info before cleanup

r1viollet:
# What does this PR do?

Ensure we aggregate live allocation information before cleaning pid info

# Motivation

Ensure we don't loose information on short lived pids

# Additional Notes

This needs more work to ensure we have a global pid object.

# How to test the change?

ongoing

### DIFF
--- a/include/live_sysallocations.hpp
+++ b/include/live_sysallocations.hpp
@@ -85,9 +85,7 @@ public:
     add_allocs(stack, addr1, size1, pid);
   }
 
-  void clear_pid(pid_t pid) {
-    _pid_map.erase(pid);
-  }
+  void clear_pid(pid_t pid) { _pid_map.erase(pid); }
 
   using StackMap = std::unordered_map<uintptr_t, UnwindOutput>;
   using PidMap = std::unordered_map<pid_t, StackMap>;
--- a/include/live_sysallocations.hpp
+++ b/include/live_sysallocations.hpp
@@ -85,22 +85,8 @@ public:
     add_allocs(stack, addr1, size1, pid);
   }
 
-  void do_exit(pid_t pid) {
-    StackMap &stack_map = _pid_map[pid];
-    stack_map.clear();
-    _visited_recently.erase(pid);
-  }
-
-  void sanitize_pids() {
-    for (auto &stack_map : _pid_map) {
-      if (!_visited_recently.contains(stack_map.first)) {
-        // This PID wasn't visited recently.  Is it still around?
-        if (kill(stack_map.first, 0)) {
-          _pid_map[stack_map.first].clear();
-        }
-      }
-    }
-    _visited_recently.clear();
+  void clear_pid(pid_t pid) {
+    _pid_map.erase(pid);
   }
 
   using StackMap = std::unordered_map<uintptr_t, UnwindOutput>;
--- a/src/ddprof_worker.cc
+++ b/src/ddprof_worker.cc
@@ -16,6 +16,7 @@
 #include "pevent_lib.hpp"
 #include "pprof/ddprof_pprof.hpp"
 #include "procutils.hpp"
+#include "signal_helper.hpp"
 #include "stack_handler.hpp"
 #include "tags.hpp"
 #include "timer.hpp"
@@ -45,6 +46,8 @@ static const DDPROF_STATS s_cycled_stats[] = {
     STATS_TARGET_CPU_USAGE};
 
 static const long k_clock_ticks_per_sec = sysconf(_SC_CLK_TCK);
+
+static DDRes worker_pid_free(DDProfContext *ctx, pid_t el);
 
 /// Human readable runtime information
 static void print_diagnostics(const DsoHdr &dso_hdr) {
@@ -264,6 +267,12 @@ static DDRes ddprof_unwind_sample(DDProfContext *ctx, perf_event_sample *sample,
     ddprof_stats_add(STATS_UNWIND_TRUNCATED_INPUT, 1, nullptr);
   }
 
+  if (us->_dwfl_wrapper->_inconsistent) {
+    // error detected on this pid
+    LG_WRN("(Inconsistent DWFL/DSOs)%d - Free associated objects", us->pid);
+    DDRES_CHECK_FWD(worker_pid_free(ctx, us->pid));
+  }
+
   return res;
 }
 
@@ -349,8 +358,8 @@ DDRes ddprof_pr_allocation_tracking(DDProfContext *ctx,
 
 DDRes ddprof_pr_sysallocation_tracking(DDProfContext *ctx,
                                        perf_event_sample *sample,
-                                       int watcher_pos) {
-
+                                       int watcher_pos, bool &clear_pid) {
+  clear_pid = false;
   // Syscall parameters.  Suppressing nags because it's annoying to look these
   // up and it isn't totally appropriate to spin out a new header just
   // for this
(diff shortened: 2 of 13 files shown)


src/ddprof_worker.cc
```diff
@@ -462,56 +472,101 @@ static DDRes aggregate_stack(const LiveAllocation::AllocationInfo &alloc_info,
...
+  unwind_pid_free(us, el);
+  ctx->worker_ctx.sys_allocation.clear_pid(el);
+  ctx->worker_ctx.live_allocation.clear_pid(el);
+  return ddres_init();
+}
 #endif
 
-static void clear_unvisted_pids(DDProfWorkerContext &worker_ctx) {
-  UnwindState *us = worker_ctx.us;
+static DDRes clear_unvisted_pids(DDProfContext *ctx) {
```
sanchda, line 560: unvisted->unvisited?

src/ddprof_worker.cc
```diff
@@ -790,8 +847,12 @@ DDRes ddprof_worker_process_event(const perf_event_header *hdr, int watcher_pos,
           // Handle special profiling types first
           if (watcher->ddprof_event_type == DDPROF_PWE_tALLOCSYS1 ||
               watcher->ddprof_event_type == DDPROF_PWE_tALLOCSYS2) {
-            DDRES_CHECK_FWD(
-                ddprof_pr_sysallocation_tracking(ctx, sample, watcher_pos));
+            bool clear_pid = false;
+            DDRES_CHECK_FWD(ddprof_pr_sysallocation_tracking(
```
sanchda, line 851: This is good for now, but making some irrelevant observations for future development.

Consider when a PID is invalidated in one of these sample-handlers.
1. When a PID is invalidated for the purposes of heap/allocation tracking, it may not be invalid for the purpose of other tracking.
2. When a PID has seen an invalidating event in one ringbuffer, there may be valid events referencing that PID in other ringbuffers.

I think it would be premature to act on these observations now.

sanchda (approved):
LGTM--thank you!

src/ddprof_worker.cc
r1viollet: True. This is not good enough.
Let's talk more about it.

include/live_sysallocations.hpp
```diff
@@ -85,23 +85,7 @@ class SystemAllocation {
...
-  void sanitize_pids() {
-    for (auto &stack_map : _pid_map) {
-      if (!_visited_recently.contains(stack_map.first)) {
-        // This PID wasn't visited recently.  Is it still around?
-        if (kill(stack_map.first, 0)) {
-          _pid_map[stack_map.first].clear();
-        }
-      }
-    }
-    _visited_recently.clear();
```
nsavoire, line 103: Is `_visited_recently` still necessary ?

src/ddprof_worker.cc
```diff
@@ -264,6 +267,12 @@ static DDRes ddprof_unwind_sample(DDProfContext *ctx, perf_event_sample *sample,
     ddprof_stats_add(STATS_UNWIND_TRUNCATED_INPUT, 1, nullptr);
   }
 
+  if (us->_dwfl_wrapper->_inconsistent) {
```
nsavoire, line 270: Just curious, what is the reason for moving this block ?

src/ddprof_worker.cc
```diff
@@ -462,56 +472,101 @@ static DDRes aggregate_stack(const LiveAllocation::AllocationInfo &alloc_info,
...
   return ddres_init();
 }
 
-static DDRes aggregate_sys_allocations(DDProfContext *ctx) {
+DDRes aggregate_sys_allocation_stack(const UnwindOutput *uw_output,
+                                     const SymbolHdr *symbol_hdr,
+                                     const PerfWatcher *watcher,
+                                     DDProfPProf *pprof) {
+  DDRES_CHECK_FWD(
+      pprof_aggregate(uw_output, symbol_hdr, 4096, 1, watcher, pprof));
```
nsavoire, line 503: Should we use a constant somewhere instead of hardcoding `4096` ?

src/ddprof_worker.cc
```diff
@@ -412,7 +421,7 @@ DDRes ddprof_pr_sysallocation_tracking(DDProfContext *ctx,
   } else if (id == 60 || id == 231 || id == 59 || id == 322 || id == 520 ||
              id == 545) {
     // Erase upon exit or exec
-    sysalloc.do_exit(sample->pid);
+    clear_pid = true;
```
nsavoire, line 424: All the syscalls here seem hardcoded for x86_64 / x86.
Do we really want to extract pid lifetime information from the syscalls (there are already perf events + process alive check) ?

src/ddprof_worker.cc
```diff
@@ -631,22 +686,24 @@ void ddprof_pr_lost(DDProfContext *ctx, const perf_event_lost *lost,
...
 }
 
-void ddprof_pr_fork(DDProfContext *ctx, const perf_event_fork *frk,
-                    int watcher_pos) {
+DDRes ddprof_pr_fork(DDProfContext *ctx, const perf_event_fork *frk,
+                     int watcher_pos) {
   LG_DBG("<%d>(FORK)%d -> %d/%d", watcher_pos, frk->ppid, frk->pid, frk->tid);
   if (frk->ppid != frk->pid) {
     // Clear everything and populate at next error or with coming samples
-    unwind_pid_free(ctx->worker_ctx.us, frk->pid);
```
nsavoire, line 648: Why do we need to call `worker_pid_free` on newly created process upon fork ?

src/ddprof_worker.cc
nsavoire, line 1: Why not use perf exit events to clean pid information ?

src/ddprof_worker.cc
```diff
@@ -607,7 +662,7 @@ DDRes ddprof_worker_cycle(DDProfContext *ctx, int64_t now,
   }
   unwind_cycle(ctx->worker_ctx.us);
 
-  clear_unvisted_pids(ctx->worker_ctx);
+  DDRES_CHECK_FWD(clear_unvisted_pids(ctx));
```
nsavoire, line 665: It might make more sense to check pids before sending profiles, that way processes that died in this cycle won't be reported on the next cycle.

src/ddprof_worker.cc
r1viollet: I agree this is probably redundant

sanchda: 1. yeah, these are hardcoded x86-specific syscalls.  I have a patch in a different branch that makes this better, although it doesn't add arm support...

2. I agree that the global PID lifetime probably shouldn't be inferred in pieces all over the application.  If an individual profiling type, which is respons

src/ddprof_worker.cc
r1viollet: Good point, this is better indeed. So
- Clear anything that died. Report it at the time of clean up
- Do the cycle after
This ensures we don't see the information in the next cycle.

src/ddprof_worker.cc
r1viollet: So the cleanup before was only removing unwinding information.
Now that we clean up all information related to the pid, we want to do it at a higher level.

src/ddprof_worker.cc
r1viollet: in case PID was already used by something else ? (yes, very unlikely)

r1viollet: I'm guessing this is probably bad actually, we might be forcing backpopulate mechanisms in case things arrive out of order.

r1viollet:
I integrated the feedbacks within
https://github.com/DataDog/ddprof/pull/240

PR #206, opened 2022-11-10
Simplify elf report

nsavoire:
Use `add_p_vaddr = true` when calling dwfl_report_elf, this allows to pass bias (ie. offset between a process address and the corresponding elf virtual address) instead of base (ie. process address of the start of first LOAD vegment).

This does not change the behaviour of ddprof (libdwfl will compute the same start/end/bias), it's only simplifies the code a little bit.

### DIFF
--- a/src/ddprof_module_lib.cc
+++ b/src/ddprof_module_lib.cc
@@ -21,8 +21,7 @@
 
 namespace ddprof {
 
-static bool get_elf_offsets(int fd, Offset_t &start_offset,
-                            const std::string &filepath,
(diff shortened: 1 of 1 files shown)


sanchda (approved):
Wow, this seems too good to be true.  Like, it's literally a 1-bit change and it drops 28 lines of code.

I expect that all that happens is the check we always did ourselves (L60-L78, etc) are now always done by libdwfl.  Does that feel correct?  If so, no reason to do it ourselves!

PR #207, opened 2022-11-13
Allow usage of jemalloc in ddprof

r1viollet:
# What does this PR do?

- Install jemalloc on alpine
- Remove prints of jemalloc stats
- Define jemalloc as the default and use a function to retrieve this from downstream build pipeline

# Motivation

Solve performance issue of musl allocator


r1viollet:
image

tools/style-check.sh
```diff
@@ -10,7 +10,7 @@ SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
 cd "$SCRIPT_DIR/.."
 
 # Find most recent clang-format, defaulting to an unqualified default
-CLANG_FORMAT=$(command -v clang-format{-13,-12,-11,-10,-9,} | head -n1)
+CLANG_FORMAT=$(command -v clang-format{-15,-13,-12,-11,-10,-9,} | head -n1)
```
sanchda, line 13: Good call, 15 hit earlier this year.

sanchda (approved):
LGTM.  I can't predict any difficulties from using jemalloc and we know from previous experience that it actually has a tremendous impact on RSS increases from libdwfl.  Good idea, thank you!

nsavoire:
When a library is available both as a shared and static library, gcc will prefer linking with the shared one except if `-static` option is used. As a result, `jemalloc` will be linked dynamically on ubuntu, which is not the desired behaviour.

sanchda:
> `jemalloc` will be linked dynamically on ubuntu which might not be the desired behaviour.

Good call, @nsavoire, since we can't assume the user will have jemalloc installed.

tools/style-check.sh
nsavoire, line 13: Is `clang-format-14`  ostracised :wink: ?

PR #211, opened 2022-11-29
Zsh minor fix

r1viollet:
# What does this PR do?

Minor fixes to compile with
 - zsh 
 - gcc-12

# Motivation

I recently setup zsh and could not compile

### DIFF
--- a/setup_env.sh
+++ b/setup_env.sh
@@ -1,6 +1,8 @@
 # Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
 # This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2021-Present Datadog, Inc.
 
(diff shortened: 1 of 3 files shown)


setup_env.sh
```diff
@@ -49,19 +57,24 @@ GetDefaultAllocatorOptions() {
 }
 
 GetDirectoryExtention() {
-  echo "_${EXTENSION_CC,,}_${EXTENSION_OS,,}_${1}"
+  echo "_${EXTENSION_CC}_${EXTENSION_OS}_${1}"
```
sanchda, line 60: Yeah, I think it's fine to remove case-normalization here.  In fact, I forget why we even enforced lowercase in the first place.

sanchda (approved):
LGTM

Issue #212, opened 2022-12-03
Support perf jitdump format


vchuravy:
Multiple languages support [perf's jitdump](https://elixir.bootlin.com/linux/v4.10/source/tools/perf/Documentation/jitdump-specification.txt) format. This includes:

- Node.js/Chrome's V8
- Java through https://github.com/torvalds/linux/tree/master/tools/perf/jvmti
- dotnet/runtime
- LLVM based languages:
  - Julia 
  -  [Wasmtime](https://docs.wasmtime.dev/examples-profiling-perf.html)
  - https://github.com/llvm/llvm-project/blob/main/llvm/lib/ExecutionEngine/PerfJITEvents/PerfJITEventListener.cpp

r1viollet:
Thanks highlighting this gap.

I have started adding some support to read symbols taking dotnet as an example. Would you have a language in particular I should test against ? Not all languages publish frame pointers (or dwarf) to unwind through stacks. I would need to test the quality of the unwinding.

Also I would need to understand where this is published. For now we check in 
`/tmp/perf-<PID>.map`

Note that we also have in process profilers dedicated to languages (for Node, Java, .NET) which will offer better results and other features like memory profiling.

aristotelhs:
This looks like a very useful feature. We are using Julia and it supports publishing jitdump format (https://docs.julialang.org/en/v1/manual/profile/#External-Profiling). Is there a patch for this support which we can help by testing?

vchuravy:
> Would you have a language in particular I should test against ? Not all languages publish frame pointers (or dwarf) to unwind through stacks. I would need to test the quality of the unwinding.

You could test against julia

```
ENABLE_JITPROFILING=1 julia test.jl
```

test.kl:
```
function fibonacci(n)
    if n == 1 || n == 2
        return 1
    else
        return fibonacci(n - 1) + fibonacci(n - 2)
    end
end


while true
    @show fibonacci(42)
    sleep(1)
end
```

The function `fibonacci` should symbolize in the profile.

aristotelhs:
I did and I created 2 flamegraphs (that is what we want in the end to do using ddprof) to show what it captures. The first one is without the perf inject --jit command, the second is after running the command. Both files are attached.

What I did notice is that Julia did not create a /tmp/perf-$pid.map file but on the other hand it created some files in `.debug/jit/llvm-IR-jit-$date-$random` whose names were in the format: `jit-$pid.dump`
[...]

PR #213, opened 2023-01-23
Fix crash when using custom stack

r1viollet:
# What does this PR do?

    When the user is using a custom stack (different from the thread's stack)
    we would crash when copying the stack to the ring buffer.
    This adds a bound check (checking that the current SP is within the stack).
    If we are not within the thread's stack, we will push an empty sample. 
    We would need access to the runtime's API to understand the stack's structure.
    For now the user will see incomplete stack traces within the UI.

# Motivation

Avoid crashes when profiling runtimes with fibers (like Julia).
When making tests for [this](https://github.com/DataDog/ddprof/issues/212) issue, we encountered this crash when enabling allocation profiling.

# Additional Notes

    There is still a case where the stack could grow and we would not detect
    the growth of the stack, however we would not crash, we would simply not
    sample.

# How to test the change?

I will deliver a PR with Julia tests in prof-correctness repository.

### DIFF
--- a/include/lib/allocation_tracker.hpp
+++ b/include/lib/allocation_tracker.hpp
@@ -24,6 +24,7 @@ struct TrackerThreadLocalState {
   int64_t remaining_bytes; // remaining allocation bytes until next sample
   bool remaining_bytes_initialized; // false if remaining_bytes is not
                                     // initialized
+  const std::byte *stack_start;     // thread stack bounds
   const std::byte *stack_end;
   pid_t tid; // cache of tid
 
--- a/include/lib/allocation_tracker.hpp
+++ b/include/lib/allocation_tracker.hpp
@@ -8,6 +8,7 @@
 #include "ddprof_base.hpp"
 #include "ddres_def.hpp"
 #include "pevent.hpp"
+#include "span.hpp"
 #include "unlikely.hpp"
 
 #include <atomic>
@@ -24,8 +25,7 @@ struct TrackerThreadLocalState {
   int64_t remaining_bytes; // remaining allocation bytes until next sample
   bool remaining_bytes_initialized; // false if remaining_bytes is not
                                     // initialized
-  const std::byte *stack_start;     // thread stack bounds
-  const std::byte *stack_end;
+  ddprof::span<const byte> stack_bounds;
   pid_t tid; // cache of tid
 
   bool reentry_guard; // prevent reentry in AllocationTracker (eg. when
--- a/include/lib/savecontext.hpp
+++ b/include/lib/savecontext.hpp
@@ -13,12 +13,10 @@
 #include "span.hpp"
 
 /** Cache stack end from current thread in TLS for future use */
-DDPROF_NOIPO int retrieve_stack_bounds(const std::byte *&start,
-                                       const std::byte *&end);
+DDPROF_NOINLINE ddprof::span<const std::byte> retrieve_stack_bounds();
 
 /** Save registers and stack for remote unwinding
  * Return saved stack size */
-DDPROF_NOIPO size_t save_context(const std::byte *stack_start,
-                                 const std::byte *stack_end,
+DDPROF_NOIPO size_t save_context(ddprof::span<const std::byte>,
                                  ddprof::span<uint64_t, PERF_REGS_COUNT> regs,
                                  ddprof::span<std::byte> buffer);
--- a/include/lib/savecontext.hpp
+++ b/include/lib/savecontext.hpp
@@ -18,7 +18,7 @@ DDPROF_NOIPO int retrieve_stack_bounds(const std::byte *&start,
 
 /** Save registers and stack for remote unwinding
  * Return saved stack size */
-DDPROF_NOIPO size_t save_context(const std::byte *start,
+DDPROF_NOIPO size_t save_context(const std::byte *stack_start,
                                  const std::byte *stack_end,
                                  ddprof::span<uint64_t, PERF_REGS_COUNT> regs,
                                  ddprof::span<std::byte> buffer);
--- a/include/lib/savecontext.hpp
+++ b/include/lib/savecontext.hpp
@@ -13,7 +13,7 @@
 #include "span.hpp"
 
 /** Cache stack end from current thread in TLS for future use */
-DDPROF_NOINLINE ddprof::span<const std::byte> retrieve_stack_bounds();
+DDPROF_NOIPO ddprof::span<const std::byte> retrieve_stack_bounds();
 
 /** Save registers and stack for remote unwinding
  * Return saved stack size */
--- a/include/lib/savecontext.hpp
+++ b/include/lib/savecontext.hpp
@@ -13,10 +13,12 @@
 #include "span.hpp"
 
 /** Cache stack end from current thread in TLS for future use */
-DDPROF_NOIPO const std::byte *retrieve_stack_end_address();
+DDPROF_NOIPO int retrieve_stack_bounds(const std::byte *&start,
+                                       const std::byte *&end);
 
 /** Save registers and stack for remote unwinding
  * Return saved stack size */
-DDPROF_NOIPO size_t save_context(const std::byte *stack_end,
+DDPROF_NOIPO size_t save_context(const std::byte *start,
+                                 const std::byte *stack_end,
                                  ddprof::span<uint64_t, PERF_REGS_COUNT> regs,
                                  ddprof::span<std::byte> buffer);
--- a/include/lib/savecontext.hpp
+++ b/include/lib/savecontext.hpp
@@ -12,7 +12,8 @@
 #include "perf_archmap.hpp"
 #include "span.hpp"
 
-/** Cache stack bounds from current thread in TLS for future use */
+/**Retrieve stack bounds from current thread.
+   Return an empty span in case of failure.*/
 DDPROF_NOIPO ddprof::span<const std::byte> retrieve_stack_bounds();
 
 /** Save registers and stack for remote unwinding
--- a/include/lib/savecontext.hpp
+++ b/include/lib/savecontext.hpp
@@ -12,11 +12,11 @@
 #include "perf_archmap.hpp"
 #include "span.hpp"
 
-/** Cache stack end from current thread in TLS for future use */
+/** Cache stack bounds from current thread in TLS for future use */
 DDPROF_NOIPO ddprof::span<const std::byte> retrieve_stack_bounds();
 
 /** Save registers and stack for remote unwinding
  * Return saved stack size */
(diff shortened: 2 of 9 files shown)


src/lib/allocation_tracker.cc
```diff
@@ -265,18 +265,21 @@ DDRes AllocationTracker::push_sample(uint64_t allocated_size,
...
   event->size = PERF_SAMPLE_STACK_SIZE;
 
-  event->dyn_size = save_context(tl_state.stack_end, event->regs,
-                                 ddprof::Buffer{event->data, event->size});
-
-  if (writer.commit(buffer) || notify_consumer) {
+  event->dyn_size =
+      save_context(tl_state.stack_start, tl_state.stack_end, event->regs,
+                   ddprof::Buffer{event->data, event->size});
+  // discard if dyn_size == 0, this way, we will skip over it
```
r1viollet, line 281: We could consider some form of lost samples. I'm not how meaningful (or confusing) it would be.

include/lib/allocation_tracker.hpp
```diff
@@ -24,6 +24,7 @@ struct TrackerThreadLocalState {
   int64_t remaining_bytes; // remaining allocation bytes until next sample
   bool remaining_bytes_initialized; // false if remaining_bytes is not
                                     // initialized
+  const std::byte *stack_start;     // thread stack bounds
```
nsavoire, line 27: stack could be represented as a `span<const std::byte>`

include/lib/savecontext.hpp
```diff
@@ -13,10 +13,12 @@
 #include "span.hpp"
 
 /** Cache stack end from current thread in TLS for future use */
-DDPROF_NOIPO const std::byte *retrieve_stack_end_address();
+DDPROF_NOIPO int retrieve_stack_bounds(const std::byte *&start,
```
nsavoire, line 16: `retrieve_stack_bounds` could return a `span`

src/lib/allocation_tracker.cc
nsavoire: I agree it would be nice to have some stats on this

src/lib/savecontext.cc
```diff
@@ -51,13 +54,18 @@ save_stack(const std::byte *stack_end, const std::byte *stack_ptr,
...
                     ddprof::span<uint64_t, PERF_REGS_COUNT> regs,
                     ddprof::span<std::byte> buffer) {
   save_registers(regs);
-  // save the stack just after saving registers, stack part above saved SP must
-  // no be changed between call to save_registers and call to save_stack
-  return save_stack(stack_end,
-                    reinterpret_cast<const std::byte *>(regs[REGNAME(SP)]),
-                    buffer);
+  // Safety check to ensure we are not in a fiber using a different stack
+  if (reinterpret_cast<const std::byte *>(regs[REGNAME(SP)]) > stack_start &&
```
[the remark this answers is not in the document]
nsavoire: Why not pass a `span` to `save_stack` and let `save_stack` handle all the checking ?

src/lib/allocation_tracker.cc
r1viollet: We do have information, we have the size of the allocation. 
Should we add a frame saying that we could not capture the buffer ? 

`malloc();[broken]`

r1viollet: This adds another new form of stack which might be confusing for the user. However this exposes the limitation and makes sure we can explain the lack of visibility. What do you think about this name ? I will have to add a new type of event.

src/lib/savecontext.cc
```diff
@@ -14,33 +14,38 @@
...
 // when we are grabbing the stack. But this is not enough, because ASAN
 // intercepts memcpy and reports a satck underflow there, empirically it appears
 // that both attribute and a suppression are required.
 static DDPROF_NO_SANITIZER_ADDRESS size_t
-save_stack(const std::byte *stack_end, const std::byte *stack_ptr,
-           ddprof::span<std::byte> buffer) {
+save_stack(ddprof::span<const std::byte> stack_bounds,
+           const std::byte *stack_ptr, ddprof::span<std::byte> buffer) {
+  // Safety check to ensure we are not in a fiber using a different stack
+  if (stack_ptr <= stack_bounds.begin() && stack_ptr >= stack_bounds.end()) {
```
r1viollet, line 43: Perhaps this should be
`stack_ptr > stack_bounds.end()`

src/lib/savecontext.cc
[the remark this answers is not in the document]
r1viollet: Thanks for suggesting

src/lib/allocation_tracker.cc
nsavoire: We could reuse the `[incomplete]` virtual frame

test/savecontext-ut.cc
```diff
@@ -27,7 +27,8 @@ std::byte stack[PERF_SAMPLE_STACK_SIZE];
 void funcB() {
   UnwindState state;
   uint64_t regs[K_NB_REGS_UNWIND];
-  size_t stack_size = save_context(retrieve_stack_end_address(), regs, stack);
+  ddprof::span<const std::byte> stack_bounds = retrieve_stack_bounds();
+  size_t stack_size = save_context(stack_bounds, regs, stack);
```
nsavoire, line 31:
```suggestion
  size_t stack_size = save_context(retrieve_stack_bounds(), regs, stack);
```

test/savecontext-ut.cc
```diff
@@ -65,7 +66,8 @@ static uint64_t regs[K_NB_REGS_UNWIND];
 static size_t stack_size;
 
 DDPROF_NO_SANITIZER_ADDRESS void handler(int sig) {
-  stack_size = save_context(retrieve_stack_end_address(), regs, stack);
+  ddprof::span<const std::byte> stack_bounds = retrieve_stack_bounds();
+  stack_size = save_context(stack_bounds, regs, stack);
```
nsavoire, line 70:
```suggestion
  stack_size = save_context(retrieve_stack_bounds(), regs, stack);
```

test/savecontext-bench.cc
```diff
@@ -36,10 +36,12 @@ DDPROF_NOINLINE static void *get_stack_start_tls() {
 static void BM_SaveContext(benchmark::State &state) {
   uint64_t regs[PERF_REGS_COUNT];
   std::byte stack[PERF_SAMPLE_STACK_SIZE];
-  auto *stack_end = retrieve_stack_end_address();
-
+  ddprof::span<const std::byte> stack_bounds = retrieve_stack_bounds();
+  if (stack_bounds.size() == 0) {
```
nsavoire, line 40:
```suggestion
  if (stack_bounds.empty()) {
```

src/lib/savecontext.cc
```diff
@@ -51,13 +56,14 @@ save_stack(const std::byte *stack_end, const std::byte *stack_ptr,
...
-size_t save_context(const std::byte *stack_end,
+size_t save_context(ddprof::span<const std::byte> stack_bounds,
                     ddprof::span<uint64_t, PERF_REGS_COUNT> regs,
                     ddprof::span<std::byte> buffer) {
   save_registers(regs);
-  // save the stack just after saving registers, stack part above saved SP must
-  // no be changed between call to save_registers and call to save_stack
-  return save_stack(stack_end,
+  // save the stack just after saving registers, stack part above saved SP
+  // must no be changed between call to save_registers and call to save_stack
```
nsavoire, line 64:
```suggestion
  // must not be changed between call to save_registers and call to save_stack
```

src/lib/savecontext.cc
nsavoire:
```suggestion
  if (!(stack_ptr >= stack_bounds.begin() && stack_ptr < stack_bounds.end())) {
```

src/lib/savecontext.cc
```diff
@@ -14,33 +14,38 @@
 #include <cstring>
 #include <pthread.h>
 
-// Return stack end address (stack end address is the start of the stack since
-// stack grows down)
-DDPROF_NOINLINE const std::byte *retrieve_stack_end_address() {
+// Returns -1 in case of failure
```
nsavoire, line 17:
```suggestion
// Returns an empty span in case of failure
```

include/lib/savecontext.hpp
```diff
@@ -13,10 +13,10 @@
 #include "span.hpp"
 
 /** Cache stack end from current thread in TLS for future use */
-DDPROF_NOIPO const std::byte *retrieve_stack_end_address();
+DDPROF_NOIPO ddprof::span<const std::byte> retrieve_stack_bounds();
 
 /** Save registers and stack for remote unwinding
  * Return saved stack size */
-DDPROF_NOIPO size_t save_context(const std::byte *stack_end,
+DDPROF_NOIPO size_t save_context(ddprof::span<const std::byte>,
```
nsavoire, line 20:
```suggestion
DDPROF_NOIPO size_t save_context(ddprof::span<const std::byte> stack_bounds,
```

src/lib/allocation_tracker.cc
```diff
@@ -263,20 +263,23 @@ DDRes AllocationTracker::push_sample(uint64_t allocated_size,
     tl_state.tid = ddprof::gettid();
   }
 
-  if (tl_state.stack_end == nullptr) {
+  if (tl_state.stack_bounds.size() == 0) {
```
nsavoire, line 266:
```suggestion
  if (tl_state.stack_bounds.empty()) {
```

src/lib/allocation_tracker.cc
```diff
@@ -263,20 +263,23 @@ DDRes AllocationTracker::push_sample(uint64_t allocated_size,
...
     // This call should only occur on main thread
-    tl_state.stack_end = retrieve_stack_end_address();
+    tl_state.stack_bounds = retrieve_stack_bounds();
+    if (tl_state.stack_bounds.size() == 0) {
```
nsavoire, line 269:
```suggestion
    if (tl_state.stack_bounds.empty()) {
```

r1viollet:
Thanks for the review @nsavoire, comments were all taken into account :bowing_man:

include/lib/savecontext.hpp
```diff
@@ -12,11 +12,11 @@
 #include "perf_archmap.hpp"
 #include "span.hpp"
 
-/** Cache stack end from current thread in TLS for future use */
-DDPROF_NOIPO const std::byte *retrieve_stack_end_address();
+/** Cache stack bounds from current thread in TLS for future use */
```
nsavoire, line 15:
```suggestion
/** Retrieve stack bounds from current thread.
      Return an empty span in case of failure.*/
```

nsavoire (approved):
LGTM !

PR #215, opened 2023-02-07
JITDump support

r1viollet:
# What does this PR do?

Use JITDump to find the symbol information.
Slight improvement in the way we lookup perfmap information.

# Motivation

Discussion in this issue : https://github.com/DataDog/ddprof/issues/212

# Additional Notes

- Jitdump detection
LLVM is very nice here, as they mmap the jitdump file to ensure we have a marker to locate this file.

- This is a partial support based on what is emitted in LLVM
I do not read the code_move structures and the unwinding information.

- Strategy to avoid bouncing on failed lookups
For now we can only fail once per Cycle (~60 seconds). After that we will stop attempting to read the perfmaps / jitdump files.
A failure can be several things: failure to find a symbol, failure to read.

- Remaining work
Some optimizations are needed to avoid fully re-reading the file
Some work to consider what happens when symbols overlap (with existing symbols). Do we evict the already existing symbols ? For now I skip over them (which is probably not the correct strategy).

# How to test the change?

I added a Julia test case in the correctness repository.
Some unit tests were added.


vchuravy:
How does this handle partially written jitdump files? IIRC parca saw cases in particular from the JVM where the writes weren't completed when they tried to do the parsing. It's basically snooping on a write ahead log :) 

FYI from a Julia perspective we will only call a function after it's information has been written to the jitdump (modulo kernel buffering writes or similar things).

sanchda:
> How does this handle partially written jitdump files?
I think your concern is covered, but I'll let @r1viollet weigh in on what the code actually does.

In terms of the operating assumptions, a few simplifications were made to keep the code sane for now, but there's some room for improvement if it turns out that writers are doing clever things.

For instance, we assumed that records might not be appended _atomically_, but operations which resize the underlying file are atomic with operations that write data.

This covers the case where the writer is using a mechanism which is ultimately a write-append.  I took a cursory look at some jitdump writer implementations across the industry and this _appeared_ to be a good assumption, but we would welcome any discussion or suggestions on the topic.

With this model, incomplete records will be detected and discarded based on the remaining size of the file, but there are several writing strategies that can't be guaranteed this way:
* Writer does something like `ftruncate()` + `mmap()/mremap()`.  This might be used for multithreaded serialization where each thread knows its serialized payload size and the threads coordinate to write to separate regions of a file.  This defeats the current implementation because a record may be comprised of an uninitialized/written segment.
* Writer uses sparse files.  This defeats the implementation in an analogous way and might be utilized for an analogous reason.

r1viollet:
@vchuravy As David mentioned we are trying to check that records are complete before reading them. 

I have however encountered situations where garbage records can lead to a bad parsing situation. Guarding against all possible content in the file would be pretty complex. This implementation probably does not scale well with high frequency updates.
I'm happy with a version that works well for Julia for now. We can look into more complex languages if we need it. We do have a very awesome dedicated Java profiler (so JVM is probably not our number one priority).

Thanks for helping us identify some of the complex cases !

include/runtime_symbol_lookup.hpp
```diff
@@ -9,26 +9,85 @@
...
+
   void erase(pid_t pid) { _pid_map.erase(pid); }
 
+  void cycle() { ++_cycle_counter; }
+
 private:
-  using PidUnorderedMap = std::unordered_map<pid_t, SymbolMap>;
-  void fill_perfmap_from_file(int pid, SymbolMap &symbol_map,
-                              SymbolTable &symbol_table);
+  using FailedCycle = std::unordered_map<std::string, uint32_t>;
```
sjanel, line 35: You could use transparent comparator here to avoid creating a `std::string` at each call to the map (you will be able to use `std::string_view` directly), with some tricks (it's a bit painful for `std::unordered_map` although easier for std::map).

Have a look at [this](https://www.reddit.com/r/cpp/comments/phqxiq/comment/hbkis8h/?utm_source=share&utm_medium=web2x&context=3) and [this godbolt link](https://godbolt.org/z/hqYxx868W) to make it compile.

r1viollet: I really liked this comment ! This was bugging me. C++ still makes it a little complex though ^^

include/runtime_symbol_lookup.hpp
```diff
@@ -9,26 +9,85 @@
...
+      return it->second == _cycle_counter;
+    }
+    return false;
+  }
+
+  void flag_lookup_failure(SymbolInfo &symbol_info, std::string_view path) {
+    symbol_info._failed_cycle[std::string(path)] = _cycle_counter;
+  }
+
+  bool should_skip_symbol(const std::string &symbol);
```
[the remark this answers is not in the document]
r1viollet: Agreed on the const. For the string_view, I'm not sure it has so much value for a private API.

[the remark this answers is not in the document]
sjanel: If the caller has a `std::string_view` and not a `std::string`, it will avoid the creation of a new `std::string`. In general, it's better to have `const std::string &` parameters replaced with `std::string_view` for this reason (a particular case when it's not possible is when we expect the input string to be null-terminated - it's not guaranteed for a `std::string_view`).

test/dso-ut.cc
```diff
@@ -217,6 +222,8 @@ static const char *s_jsa_line = "0x800000000-0x800001fff rw-p 00000000 00:00 0
 
 static const char *s_dd_profiling = "0x800000000-0x800001fff rw-p 00000000 00:00 0                          /tmp/libdd_profiling.so.1234";
 static const char *s_dotnet_line = "7fbd4f1e4000-7fbd4f1ec000 r--s 00000000 ca:01 140372                     /usr/share/dotnet/shared/Microsoft.NETCore.App/6.0.5/System.Runtime.dll";
+static const char *s_jitdump_line = "7b5242e44000-7b5242e45000 r-xp 00000000 fd:06 22295230                   /home/r1viollet/.debug/jit/llvm-IR-jit-20230131-981d92/jit-3237589.dump";
```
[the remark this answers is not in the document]
r1viollet: good reminder.

include/runtime_symbol_lookup.hpp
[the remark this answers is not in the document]
r1viollet: I see some value in unifying the code (though not a lot of value in the current example). Thanks. fixed.

include/runtime_symbol_lookup.hpp
```diff
@@ -9,26 +9,133 @@
...
 #include <unordered_map>
 
 namespace ddprof {
 
 class RuntimeSymbolLookup {
 public:
+  struct Stats {
+    uint32_t _nb_jit_reads = {};
+    uint32_t _nb_failed_lookups = {};
+    mutable uint32_t _count = {};
```
nsavoire, line 23: It's weird to have a mutable field here

include/runtime_symbol_lookup.hpp
```diff
@@ -9,26 +9,133 @@
...
                             SymbolTable &symbol_table);
+
   void erase(pid_t pid) { _pid_map.erase(pid); }
 
+  void cycle() {
+    ++_cycle_counter;
+    _stats = {};
+  }
+
+  const Stats &get_stats() const {
```
nsavoire, line 43: Return by value this would avoid the weird mutable count

include/runtime_symbol_lookup.hpp
```diff
@@ -9,26 +9,133 @@
...
+    }
+    return false;
+  }
+
+  void flag_lookup_failure(SymbolInfo &symbol_info, std::string_view path) {
+    const auto it = symbol_info._failed_cycle.find(path);
+    // Written this way, we save up on creating strings
+    // only the slow path will create a string for the path
+    if (it != symbol_info._failed_cycle.end()) {
+      ++(it->second);
```
nsavoire, line 116: Seems wrong, this also should set value to `_cycle_counter`

include/runtime_symbol_lookup.hpp
```diff
@@ -9,26 +9,133 @@
...
+    ++_stats._nb_failed_lookups;
+  }
+
+  bool should_skip_symbol(std::string_view symbol) const;
+
+  bool insert_or_replace(const std::string &symbol, ProcessAddress_t address,
+                         Offset_t size, SymbolMap &symbol_map,
+                         SymbolTable &symbol_table);
+
+  static constexpr std::array<const std::string_view, 1>
```
nsavoire, line 129: Could be moved in `.cc` file

include/ddprof_stats.hpp
```diff
@@ -23,6 +23,9 @@
   X(UNWIND_INCOMPLETE_STACK, "unwind.stack.incomplete", STAT_GAUGE)            \
   X(UNWIND_AVG_STACK_SIZE, "unwind.stack.avg_size", STAT_GAUGE)                \
   X(UNWIND_AVG_STACK_DEPTH, "unwind.stack.avg_depth", STAT_GAUGE)              \
+  X(SYMBOLS_JIT_READS, "symbols.jit.reads", STAT_GAUGE)                        \
+  X(SYMBOLS_JIT_FAILED_LOOKUPS, "symbols.jit.failed_lookups", STAT_GAUGE)      \
+  X(SYMBOLS_JIT_COUNT, "symbols.jit.count", STAT_GAUGE)                        \
```
nsavoire, line 28: `symbol_count` instead of `count` might be clearer

include/runtime_symbol_lookup.hpp
nsavoire, line 23: `symbol_count` instead of `count` might be clearer

src/jit/jitdump.cc
```diff
@@ -0,0 +1,215 @@
...
+  return true;
+}
+
+DDRes jit_read_code_load(std::ifstream &file_stream,
+                         JITRecordCodeLoad &code_load,
+                         std::vector<char> &buff) {
+#ifdef DEBUG
+  LG_DBG("----  Read code load  ----");
+#endif
+  buff.resize(code_load.prefix.total_size - sizeof(JITRecordPrefix));
```
nsavoire, line 84: if `code_load.prefix.total_size < sizeof(JITRecordPrefix)`, this will be incorrect

src/jit/jitdump.cc
```diff
@@ -0,0 +1,215 @@
...
+  file_stream.read(buff.data(),
+                   code_load.prefix.total_size - sizeof(JITRecordPrefix));
+  if (!file_stream.good()) {
+    // can happen if we are in the middle of a write
+    DDRES_RETURN_WARN_LOG(DD_WHAT_JIT, "Incomplete code load structure");
+  }
+  uint32_t *buf_32 = reinterpret_cast<uint32_t *>(buff.data());
+  code_load.pid = *buf_32++;
```
nsavoire, line 92: You should check that you don't overrun buffer end when reading

src/runtime_symbol_lookup.cc
```diff
@@ -10,16 +10,23 @@
...
 
+// 00007F78F5230000 d8 stub<1> AllocateTemporaryEntryPoints<PRECODE_STUB>
+// 00007F78F52300D8 78 stub<2> AllocateTemporaryEntryPoints<PRECODE_STUB>
+// 00007F78F5230150 18 stub<3> AllocateTemporaryEntryPoints<PRECODE_STUB>
+
 FILE *RuntimeSymbolLookup::perfmaps_open(int pid,
                                          const char *path_to_perfmap = "") {
-  char buf[1024] = {0};
+  char buf[1024];
   auto n = snprintf(buf, 1024, "%s/proc/%d/root%s/perf-%d.map",
```
nsavoire, line 30:
```suggestion
  auto n = snprintf(buf, std::size(buf), "%s/proc/%d/root%s/perf-%d.map",
```

src/runtime_symbol_lookup.cc
```diff
@@ -31,86 +38,171 @@ FILE *RuntimeSymbolLookup::perfmaps_open(int pid,
   }
   // attempt in local namespace
   snprintf(buf, 1024, "%s/perf-%d.map", path_to_perfmap, pid);
```
nsavoire, line 40:
```suggestion
  snprintf(buf, std::size(buf), "%s/perf-%d.map", path_to_perfmap, pid);
```

src/runtime_symbol_lookup.cc
```diff
@@ -31,86 +38,171 @@ FILE *RuntimeSymbolLookup::perfmaps_open(int pid,
...
+  return true;
 }
 
-void RuntimeSymbolLookup::fill_perfmap_from_file(int pid, SymbolMap &symbol_map,
-                                                 SymbolTable &symbol_table) {
+DDRes RuntimeSymbolLookup::fill_from_jitdump(std::string_view jitdump_path,
+                                             pid_t pid, SymbolMap &symbol_map,
+                                             SymbolTable &symbol_table) {
+  char buf[1024];
+  auto n = snprintf(buf, 1024, "%s/proc/%d/root%s", _path_to_proc.c_str(), pid,
```
nsavoire, line 100:
```suggestion
  auto n = snprintf(buf, std::size(buf), "%s/proc/%d/root%s", _path_to_proc.c_str(), pid,
```

src/runtime_symbol_lookup.cc
```diff
@@ -31,86 +38,171 @@ FILE *RuntimeSymbolLookup::perfmaps_open(int pid,
...
+                    jitdump_path.data());
+  if (n >= 1024) { // unable to snprintf everything
```
nsavoire, line 102:
```suggestion
  if (n >= std::size(buf)) { // unable to snprintf everything
```

src/runtime_symbol_lookup.cc
```diff
@@ -31,86 +38,171 @@ FILE *RuntimeSymbolLookup::perfmaps_open(int pid,
...
     char size_buff[33];
     // Avoid considering any symbols beyond 300 chars
     if (3 !=
-            sscanf(line, "%16s %8s %300[^\t\n]", address_buff, size_buff,
-                   buffer) ||
-        should_skip_symbol(buffer)) {
+        sscanf(line, "%16s %8s %300[^\t\n]", address_buff, size_buff, buffer)) {
       continue;
     }
+    std::string sym = std::string(buffer);
```
nsavoire, line 156: Why not pass `buffer` directly to `insert_or_replace` ?

r1viollet: better indeed

include/runtime_symbol_lookup.hpp
r1viollet: I'm not sure what I prefer. I adjusted according to your proposal.

src/jit/jitdump.cc
```diff
@@ -0,0 +1,215 @@
...
+    // todo everything should be swapped throughout the parsing (not handled)
+    DDRES_RETURN_ERROR_LOG(DD_WHAT_JIT, "Swap data not handled");
+  } else {
+    DDRES_RETURN_ERROR_LOG(DD_WHAT_JIT, "Unknown jit format");
+  }
+  auto remaining_size = header.total_size - sizeof(header);
+  if (remaining_size) {
+    std::vector<char> read_buf;
+    read_buf.resize(remaining_size);
+    file_stream.read(read_buf.data(), remaining_size);
```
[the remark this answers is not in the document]
r1viollet: better again!

src/jit/jitdump.cc
```diff
@@ -0,0 +1,215 @@
...
+    }
+  } while (valid_entry);
+  return ddres_init();
+}
+
+DDRes jitdump_read(const std::string_view file, JITDump &jit_dump) {
+  std::ifstream file_stream(file.data(), std::ios::binary);
+  // We are not locking, assumption is that even if we fail to read a given
+  // section we can always retry later. The aim is not to slow down the app
+  if (!file_stream.good()) {
```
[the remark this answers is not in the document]
r1viollet: I'm not sure this is better.

include/runtime_symbol_lookup.hpp
r1viollet: I see value in having this more visible in header.

r1viollet:
@nsavoire / @sjanel I think this one is good to go. Thanks for the amazing reviews.

nsavoire (approved):
LGTM !

PR #218, opened 2023-02-20
Upgrade dependencies

r1viollet:
# What does this PR do?

- Allow ubuntu 22 to build
- Upgrade elfutils to 0.188
- Upgrade default alpine to 3.17.2

# Motivation

N/A


app/base-env/llvm.sh
```diff
@@ -10,8 +10,21 @@
 
 set -eux
 
+usage() {
```
nsavoire, line 13: Why not retrieve this file from https://apt.llvm.org/llvm.sh ?

cmake/Findelfutils.cmake
```diff
@@ -9,17 +9,17 @@
 # Lzma / ZLib Force static library by overriding CMAKE_FIND_LIBRARY_SUFFIXES since
 # FindZLIB/FindLIBLZMA don't provide the option
 set(ORIG_CMAKE_FIND_LIBRARY_SUFFIXES ${CMAKE_FIND_LIBRARY_SUFFIXES})
-set(CMAKE_FIND_LIBRARY_SUFFIXES ".a")
+set(CMAKE_FIND_LIBRARY_SUFFIXES ".a" ".so")
```
nsavoire, line 12: Why change this ?

r1viollet: This is a mistake. Thanks for spotting this. There was an issue previously with lzma

app/base-env/llvm.sh
r1viollet: That is another option. Either way I would have to compute the sha. Copying the file is roughly the same overhead.

src/lib/loader.c
```diff
@@ -55,18 +55,21 @@ int timer_create(clockid_t clockid, struct sigevent *sevp, timer_t *timerid)
     __attribute__((weak));
 
 static void *s_libdl_handle = NULL;
+static __typeof(dlerror) *s_dlerror = NULL;
+static __typeof(dlopen) *dlopen_ptr = &dlopen;
```
nsavoire, line 59:
```suggestion
static __typeof(dlopen) *s_dlopen = &dlopen;
```

src/lib/loader.c
```diff
@@ -55,18 +55,21 @@ int timer_create(clockid_t clockid, struct sigevent *sevp, timer_t *timerid)
     __attribute__((weak));
 
 static void *s_libdl_handle = NULL;
+static __typeof(dlerror) *s_dlerror = NULL;
```
nsavoire, line 58:
```suggestion
static __typeof(dlerror) *s_dlerror = &dlerror;
```

Need to add at the top:
```
const char *dlerror(void) __attribute__((weak));
```

src/lib/loader.c
```diff
@@ -262,6 +265,13 @@ static void __attribute__((constructor)) loader() {
   ensure_libpthread_is_loaded();
   ensure_librt_is_loaded();
 
+  // now that we have loaded libdl, we can ensure that we use the real
+  // dlopen function (instead of internal libc function)
+  if (s_libdl_handle) {
```
nsavoire, line 270: Code could be moved to `ensure_libdl_is_loaded`, `s_dlerror` should be updated only if null

nsavoire (approved):
LGTM !

Issue #217, opened 2023-02-09
Currently values of profile tags can only start with letters


tomthomson:
Hi,
When calling the profiler with the tags attribute like this
`./ddprof/bin/ddprof -l warn -T 'someUUID:0dbe19c2-a86b-11ed-afa1-0242ac120002'`
where the value of the tag starts with a number, then the tag is ignored:
`<WARNING>Feb 09 12:03:24.964300 ddprof[124943]: [TAGS] Bad tag value - skip someUUID:0dbe19c2-a86b-11ed-afa1-0242ac120002`

This is due to the fact that after splitting the tags into keys and values the same validation is applied for tags and values:
https://github.com/DataDog/ddprof/blob/6d3665544ea3edc9ee851dd7ff9314a282ae1065/src/tags.cc#L70-L77

I see the ValidateTags function is taken from the DogFood repository, but there the ValidateTags is called on the combined key-value string - not the splitted key and value string
https://github.com/garrettsickles/DogFood/blob/2ff69d92b7c1c63bee3ac8d5657ed8d2f34ef29d/DogFood.hpp#L358-L362

The Datadog docs also only document rules for the entire tag - which is the combined key-value string: https://docs.datadoghq.com/getting_started/tagging/#define-tags

**Solution**
I think ValidateTags should only be called on the combined key-value string

r1viollet:
Thank you for being so accurate in your bug report :bow:

PR #221, opened 2023-03-03
Fix tag validation

r1viollet:
# What does this PR do?

Call tag validation on the key value string.
In the current implementation we were not allowing IDs to be forwarded as tags.

# Motivation

This was reported in https://github.com/DataDog/ddprof/issues/217


nsavoire (approved):
LGTM

PR #223, opened 2023-03-07
Upgrade codeql version

r1viollet:
# What does this PR do?

This also fixes some of the libc issues that were coming up during codeql builds on ubuntu 22


nsavoire (approved):
LGTM !

PR #224, opened 2023-03-12
Feature/leak detection

r1viollet:
# What does this PR do?

Leak analysis tooling.
Also known as live heap. The aim is to highlight the memory that is still alive and relates to RSS usage.

# Motivation

Several requests to investigate leaks.

# Additional Notes

Some follow ups are expected to:
- Push the data to a different type of pprof
- Improve the documentation around the usage of the feature
- Input checks to prevent faulty usage of live mode option `mode=l`

# How to test the change?

If you compile the change, you can request it manually
```
ddprof -l debug --preset cpu_only -e "sALLOC mode=l" -u 30 ./test
```

You can also see a test here (private Datadog repo)
https://github.com/DataDog/prof-correctness/pull/38/files

Deploying this to staging, the overhead numbers seem reasonable.


include/ddprof_context_lib.hpp
```diff
@@ -14,4 +14,6 @@ typedef struct PerfWatcher PerfWatcher;
 DDRes ddprof_context_set(DDProfInput *input, DDProfContext *);
 void ddprof_context_free(DDProfContext *);
 
+void log_watcher(const PerfWatcher *w, int idx);
```
r1viollet, line 17: this should be declared somewhere else

r1viollet:
Missing parts
- PID cleanup and aggregating heap view before cleanup
https://github.com/DataDog/ddprof/pull/203/files
- Avoiding unbounded growth, we can find a better strategy and this can remain an experimental feature
https://github.com/DataDog/ddprof/pull/183/files

nsavoire:
I am not sure how I feel about abstracting the concept of LiveCallgraph to all watchers:
- This currently only works for allocations (what happens if we use the option for other watchers ?)
- The implementation is very specific to allocations, I don't think there is a generic way to define a live callgraph for any watcher

nsavoire:
We should probably create new profile types in backend/UI for live allocations instead of reusing those for allocations.

src/lib/malloc_wrapper.cc
```diff
@@ -0,0 +1,222 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
```
r1viollet, line 1: to be removed

src/ddprof_worker.cc
```diff
@@ -584,14 +767,25 @@ DDRes ddprof_worker_process_event(const perf_event_header *hdr, int watcher_pos,
...
+
         if (sample) {
-          DDRES_CHECK_FWD(ddprof_pr_sample(ctx, sample, watcher_pos));
+          // Handle special profiling types first
+          if (watcher->ddprof_event_type == DDPROF_PWE_tALLOCSYS1 ||
+              watcher->ddprof_event_type == DDPROF_PWE_tALLOCSYS2) {
+            // For now we have a different path for
+            // - mmap/munmap syscalls
+            DDRES_CHECK_FWD(
+                ddprof_pr_sysallocation_tracking(ctx, sample, watcher_pos));
```
r1viollet, line 785: todo: split PR into 2 PRs

src/lib/allocation_tracker.cc
```diff
@@ -184,16 +199,34 @@ void AllocationTracker::track_allocation(uintptr_t, size_t size,
...
-    }
-  } else {
-    if (_state.failure_count.load(std::memory_order_relaxed) > 0) {
-      _state.failure_count = 0;
-    }
+  bool success = IsDDResOK(push_sample(addr, total_size, tl_state));
+  free_on_consecutive_failures(success);
+
+  if (success) { // ensure we track this dealloc if it occurs
+    _address_set.insert(addr);
```
r1viollet, line 206: at 500 000 elements we should clear the state of the set.

r1viollet:
> We should probably create new profile types in backend/UI for live allocations instead of reusing those for allocations.

This part I will do after this PR if that is OK.

include/lib/allocation_tracker.hpp
```diff
@@ -60,13 +63,22 @@ class AllocationTracker {
   static inline bool is_active();
 
 private:
+  using AdressSet = std::unordered_set<uintptr_t>;
+
   struct TrackerState {
+    void init(bool track_alloc, bool track_dealloc) {
```
r1viollet, line 69: not very idiomatic... Should I rely on constructor ? Can we afford to re-init the mutex.

PR #225, opened 2023-03-13
Documentation update - v0.11.0

r1viollet:
# What does this PR do?

- Improved documentation
- Updated changelog

# Motivation

Our documentation was confusing


Readme.md
```diff
@@ -1,37 +1,40 @@
...
-The Datadog Native Profiler for Linux has only been tested on kernel 4.15 above.  It may be supported by older kernels, but your mileage may vary.  One can verify the kernel version by running `uname`:
-
-```bash
-uname -r
-```
+#### Perf event paranoid
 
-In addition, the target machine must have `perf_event_paranoid` set to 2 or lower OR `CAP_SYS_ADMIN` enabled.
+The target machine must have `perf_event_paranoid` set to 2 or lower OR `CAP_SYS_ADMIN` enabled.
+Some distributions have set the default to higher values (up to 4), which will prevent `ddprof` from instrumenting CPU activity through the perf_event_open API.
```
nsavoire, line 37: Maybe add a bash snippet to change the value `perf_event_paranoid` ?

sanchda: I don't know if we need to specify that some distros have higher values.  I feel that `perf_event_paranoid <= 2` covers that and wonder whether overstating it might be confusing for some people.

We should probably mention `CAP_PERFMON` being introduced in Linux 5.8.

sanchda (approved):
LGTM, thank you!  Added one note, but it isn't a blocker.

Readme.md
r1viollet: I thought of this, though it would not cover all cases. I guess the most common way is to have a boot time setting ?

`sudo sh -c 'echo kernel.perf_event_paranoid=2 > /etc/sysctl.d/perf_event_paranoid_2.conf'`

nsavoire: I was more thinking about changing `perf_event_paranoid` temporarily for testing with:
```bash
echo 2 | sudo tee /proc/sys/kernel/perf_event_paranoid
```
But both would be nice

nsavoire (dismissed):
Nice improvement to documentation !

PR #226, opened 2023-03-13
Add jq as a build time dependency

r1viollet:
# What does this PR do?

The scripts that push automatically to GitHub require the jq dependency


nsavoire (approved):
LGTM!

PR #228, opened 2023-03-16
Minor - add a test to backpopulate mechanism

r1viollet:
# What does this PR do?

Add a test for the back populate mechanism with a large proc maps file.

# Motivation

Make sure we don't grow the size of the DSO with several backpopulates.
Have some test data to improve performance.


include/dso_hdr.hpp
```diff
@@ -86,7 +86,7 @@ class DsoHdr {
   using DsoFindRes = std::pair<DsoMapConstIt, bool>;
 
   /******* MAIN APIS **********/
-  explicit DsoHdr(int dd_profiling_fd = -1);
```
r1viollet, line 89: Explicit no longer needed
A comment around the API would be nice.

nsavoire (approved):
LGTM!

include/dso_hdr.hpp
r1viollet: Actually explicit is still needed due to default parameters.

PR #230, opened 2023-03-23
Better Rust demangling

sanchda:
# What does this PR do?

Adds support for additional post-processing for Rust symbols.  The test harness for these additions were provided by libiberty.  This implementation is not 100% correct, but it is better than what we currently have.  Hopefully this will serve until libdatadog has demangling support.

# Motivation

Many of our users enjoy Rust and are sad that their symbol names are unreadable.

# Additional Notes

I tried to make a few performance-conscious decisions when they wouldn't have negatively impacted simplicity (and when they came to mind).  I used https://github.com/sanchda/systems_experiments/blob/main/find_all_syms/findsyms.sh to create a test harness and looked at the results from the native profiler.  Some room for improvement, but added only a small overhead (<10%) to demangling, which is already not a major part of symbolization.

# How to test the change?

Tests are included.


include/demangler/demangler.hpp
```diff
@@ -0,0 +1,16 @@
...
+// under the Apache License Version 2.0. This product includes software
+// developed at Datadog (https://www.datadoghq.com/). Copyright 2021-Present
+// Datadog, Inc.
+
+#pragma once
+
+#include <string>
+#include <string_view>
+
+namespace Demangler {
```
nsavoire, line 11: Why not use `ddprof` namespace ?

CMakeLists.txt
```diff
@@ -139,12 +139,14 @@ list(APPEND DDPROF_INCLUDE_LIST ${CMAKE_SOURCE_DIR}/include ${CMAKE_SOURCE_DIR}/
 # Find the source files
 aux_source_directory(src COMMON_SRC)
 aux_source_directory(src/pprof PPROF_SRC)
+aux_source_directory(src/demangler DEMANGLER_SRC)
```
nsavoire, line 142: Not sure a directory is need for a single file

src/demangler/demangler.cc
```diff
@@ -0,0 +1,155 @@
...
+#include <llvm/Demangle/Demangle.h>
+
+// With some exceptions we don't handle here, v0 Rust symbols can end in a
+// prefix followed by a 16-hexdigit hash, which must be removed
+constexpr std::string_view hash_pre = "::h";
+constexpr std::string_view hash_eg = "0123456789abcdef";
+
+// Minimal check that a string can end, and does end, in a hashlike substring
+// Some tools check for entropy, we do not.
+static inline bool has_hash(const std::string &str) {
```
nsavoire, line 20: Better to use `std::string_view`

src/demangler/demangler.cc
```diff
@@ -0,0 +1,155 @@
...
+  // If the size can't conform, then the string is invalid
+  if (str.size() <= hash_pre.size() + hash_eg.size()) {
+    return false;
+  }
+
+  // Check that the string contains the hash prefix in the right position
+  if (str.compare(str.size() - hash_eg.size() - hash_pre.size(),
```
nsavoire, line 27: With `std::string_view`:
```c++
auto str2 = str.substr(str.size() - hash_eg.size() - hash_pre.size());
return str2.starts_with(hash_pre) && std::all_of(str2.begin() + hash_pre.size(), str2.end(), 
    [](auto c) { return isxdigit(c) && (islower(c) || isdigit(c)); });
```

src/demangler/demangler.cc
```diff
@@ -0,0 +1,155 @@
...
+                  hash_pre.size(), hash_pre)) {
+    return false;
+  }
+
+  // Check that the string ends in lowercase hex digits
+  for (size_t i = str.size() - hash_eg.size(); i < str.size(); ++i) {
+    if (!isxdigit(str[i]) || (!islower(str[i]) && !isdigit(str[i]))) {
```
nsavoire, line 34: Would probably be better to write your own `is_lower_xdigit` function ;)

src/demangler/demangler.cc
```diff
@@ -0,0 +1,155 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0. This product includes software
+// developed at Datadog (https://www.datadoghq.com/). Copyright 2021-Present
+// Datadog, Inc.
+
+#include "demangler/demangler.hpp"
+
+#include <string.h> // isspace, isxdigit, islower
```
nsavoire, line 8: Rather `ctype.h` or `cctype`

src/demangler/demangler.cc
```diff
@@ -0,0 +1,155 @@
...
+    return dig - 'a' + 10;
+  else if (dig >= 'A' && dig <= 'F')
+    return dig - 'A' + 10;
+  else
+    return -1;
+}
+
+// Demangles a Rust string by building a copy piece-by-piece
+static std::string rust_demangle(const std::string &str) {
+  static const std::unordered_map<std::string, std::string> patterns{
```
nsavoire, line 87: Why use an `std::unordered_map` ? 
`patterns` is never used as a map

src/demangler/demangler.cc
```diff
@@ -0,0 +1,155 @@
...
+        }
+      } else if (!replaced) {
+        ret += str[i];
+      }
+    } else {
+      ret += str[i];
+    }
+  }
+
+  ret.shrink_to_fit();
```
nsavoire, line 145: Not sure this is useful, in most cases demangled string will be larger than source string, and in the cases where it's not, losing a few bytes of memory should not be an issue.

sanchda: I agree that this probably isn't _too_ impactful--on the order of ~19 bytes per string, plus a little bit more for strings with a lot of pattern substitutions.

At this point in processing, the input string has undergone C++ demangling (making it bigger than the source string).  This function acts on the resulting string by
1. Dropping a terminating hash (16 + 3 characters)
2. Converting patterns of length 3-5 into single characters (or in one case, two characters)

I guess most of that space could be restored by just not reserving the full size of the input on line 93, but rather `str.size() - hash_eg.size() - hash_pre.size())` or something, which is probably better than shrinking at the end.  In that case, I don't think the few bytes we gain by shrinking in the average case are worth it (ddprof manages its own deduplication for function names, but when strings are transferred to libdatadog I anticipate them to be copied by size rather than capacity).

I'll modify this accordingly--thanks for noticing

nsavoire:
All my comments about this PR are nitpicks ;)

src/demangler/demangler.cc
nsavoire: My bad, please disregard my comment.
I completely missed the fact that the hash at the end was removed !

include/demangler/demangler.hpp
```diff
@@ -0,0 +1,16 @@
...
+
+// Functions
+std::string demangle(const std::string &mangled);
```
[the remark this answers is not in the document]
sanchda: Yup, fixed in 8bb02a9

src/demangler/demangler.cc
sanchda: Ah, right.  Fixed in 8bb02a9

src/demangler/demangler.cc
sanchda: Agree, fixed in 8bb02a9

src/demangler/demangler.cc
sanchda: Yup, you're right.  Changed to a `std::array` of pairs in 8bb02a9.

CMakeLists.txt
sanchda: Yeah, I had mixed feelings about this.  I wanted to keep the implementation pure so that the Java profiler could use it as well, but I think they'll only care about C++ demangling anyway, so this implementation will be slightly unnecessary for them.

If it's OK with you, I'll ignore your feedback for the purpose of this PR, merge (when/if approved), propose to the Java team, and almost definitely spin out a subsequent PR where I do exactly what you suggest here and in the comment about the namespace.  :)

sanchda:
As usual, thank you for the considerate + thoughtful review, @nsavoire!  I took many of your suggestions (and the rest, I'll almost definitely take later)

r1viollet:
This is great! Thanks for wrapping this up!
Not jumping in as @nsavoire already did a great review.

nsavoire (approved):
LGTM !

Issue #232, opened 2023-03-23
Support UDS


r1viollet:
# Description

Allow usage of Unix Domain Sockets to send profiles.

# Motivation

On k8s, these are easier to setup.

r1viollet:
UDS works, though it needs a specific option. The standard behavior should be to detect if a UDS configuration is available.

r1viollet:
Closing as we support UDS. We can reopen if someone wants better automatic configuration.

PR #233, opened 2023-03-23
Enable UDS settings

r1viollet:
# What does this PR do?

Ensure UDS settings are allowed by the exporter

# Motivation

Allow usage of UDS.
https://github.com/DataDog/ddprof/issues/232

# Testing

Testing is ongoing

### DIFF
--- a/include/ddprof_input.hpp
+++ b/include/ddprof_input.hpp
@@ -34,7 +34,6 @@ typedef struct DDProfInput {
   char *worker_period;
   char *internal_stats;
   char *tags;
-  char *url;
   char *socket;
   char *preset;
   char *switch_user;
@@ -77,7 +76,7 @@ typedef struct DDProfInput {
   XX(DD_ENV,                            environment,        E, 'E', 1, input, NULL, "",           exp_input.)  \
   XX(DD_AGENT_HOST,                     host,               H, 'H', 1, input, NULL, "localhost",  exp_input.)  \
   XX(DD_TRACE_AGENT_PORT,               port,               P, 'P', 1, input, NULL, "8126",       exp_input.)  \
-  XX(DD_TRACE_AGENT_URL,                url,                U, 'U', 1, input, NULL, "", )                      \
+  XX(DD_TRACE_AGENT_URL,                url,                U, 'U', 1, input, NULL, "",           exp_input.)  \
   XX(DD_SERVICE,                        service,            S, 'S', 1, input, NULL, "myservice",  exp_input.)  \
   XX(DD_VERSION,                        service_version,    V, 'V', 1, input, NULL, "",           exp_input.)  \
   XX(DD_PROFILING_EXPORT,               do_export,          X, 'X', 1, input, NULL, "yes",        exp_input.)  \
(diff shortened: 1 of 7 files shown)


include/ddprof_input.hpp
```diff
@@ -76,9 +75,8 @@ typedef struct DDProfInput {
   XX(DD_API_KEY,                        api_key,            A, 'A', 1, input, NULL, "",           exp_input.)  \
   XX(DD_ENV,                            environment,        E, 'E', 1, input, NULL, "",           exp_input.)  \
   XX(DD_AGENT_HOST,                     host,               H, 'H', 1, input, NULL, "localhost",  exp_input.)  \
-  XX(DD_SITE,                           site,               I, 'I', 1, input, NULL, "",           exp_input.)  \
   XX(DD_TRACE_AGENT_PORT,               port,               P, 'P', 1, input, NULL, "8126",       exp_input.)  \
-  XX(DD_TRACE_AGENT_URL,                url,                U, 'U', 1, input, NULL, "", )                      \
+  XX(DD_TRACE_AGENT_URL,                url,                U, 'U', 1, input, NULL, "",           exp_input.)  \
```
sanchda, line 79: Ah, makes sense to move this 👍

src/exporter/ddprof_exporter.cc
```diff
@@ -91,20 +115,43 @@ DDRes ddprof_exporter_init(const ExporterInput *exporter_input,
...
-    exporter->_url =
-        alloc_url_agent("http://", exporter_input->host, exporter_input->port);
+    const char *port_str = exporter_input->port;
+
+    if (exporter_input->url) {
+      // uds -> no port
+      if (!strncasecmp(exporter_input->url, "unix", 4)) {
+        port_str = nullptr;
+      }
+      // already port -> no port
```
sanchda, line 125: yup, agreed

src/exporter/ddprof_exporter.cc
```diff
@@ -91,20 +115,43 @@ DDRes ddprof_exporter_init(const ExporterInput *exporter_input,
   }
 
   if (exporter->_agent) {
-    exporter->_url =
-        alloc_url_agent("http://", exporter_input->host, exporter_input->port);
+    const char *port_str = exporter_input->port;
```
nsavoire, line 118: This should be probably be extracted to its own function.
I would simplify the logic by sopping mixing url and port:
* either a url is given in input, in that case it's expected to be a full url (with protocol and port), in that host and port are ignored
* either no url is given in input and url is built from 'http', host and port

That way you can even stop distinguishing between agent and agentless.

sanchda (approved):
I think this looks good.  At some point we should probably change the input stuff to using std::string or something instead of raw C strings, but I don't think this is really the time or the place for that.  :)

Good stuff!  Thank you!  UDS support may not improve performance substantially, since we use async, but it's still a configuration users will want!

nsavoire (approved):

src/exporter/ddprof_exporter.cc
r1viollet: This is a good idea, though we would need to check if endpoint is reachable on startup,
Adding this permissive config might be bad for reverting it later.

r1viollet:
Merging as is for now, I agree there should be a follow up on this.
The way to configure this is not really explicit as of now. The documentation needs some love. FYI:

```
unix:///run/apm.sock
```

PR #234, opened 2023-03-23
Fix typo - set private field - add some missing includes

sjanel:
Remove extra ';'
Set `kInvalidSocket` private, removed `inline` (`static constexpr` implies inline)
Add missing includes


nsavoire (approved):
LGTM !
Thanks for the contribution ;)

PR #240, opened 2023-04-07
Live Heap - Pid cleanup

r1viollet:
- Clear aggregated live information when we remove a PID.
- Re-introduce the watcher dimension in the aggregation of allocations.

### DIFF
--- a/src/ddprof_worker.cc
+++ b/src/ddprof_worker.cc
@@ -16,6 +16,7 @@
 #include "pevent_lib.hpp"
 #include "pprof/ddprof_pprof.hpp"
 #include "procutils.hpp"
+#include "signal_helper.hpp"
 #include "stack_handler.hpp"
 #include "tags.hpp"
 #include "timer.hpp"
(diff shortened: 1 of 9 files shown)


src/ddprof_worker.cc
```diff
@@ -818,10 +908,18 @@ DDRes ddprof_worker_process_event(const perf_event_header *hdr, int watcher_pos,
...
-                             reinterpret_cast<const DeallocationEvent *>(hdr));
+                             reinterpret_cast<const DeallocationEvent *>(hdr),
+                             watcher_pos);
+      break;
     case PERF_CUSTOM_EVENT_CLEAR_LIVE_ALLOCATION:
-      ddprof_pr_clear_live_allocation(
-          ctx, reinterpret_cast<const ClearLiveAllocationEvent *>(hdr));
+      {
+        const ClearLiveAllocationEvent* event = reinterpret_cast<const ClearLiveAllocationEvent *>(hdr);
+#ifndef DDPROF_NATIVE_LIB
```
r1viollet, line 917: this is not great, but I'm removinog the native lib in next PR

src/ddprof_worker.cc
```diff
@@ -442,76 +446,151 @@ void *ddprof_worker_export_thread(void *arg) {
...
 #endif
 
-static void clear_unvisted_pids(DDProfWorkerContext &worker_ctx) {
-  UnwindState *us = worker_ctx.us;
+static DDRes clear_unvisited_pids(DDProfContext *ctx) {
+  UnwindState *us = ctx->worker_ctx.us;
   const std::vector<pid_t> pids_remove = us->dwfl_hdr.get_unvisited();
   for (pid_t el : pids_remove) {
-    unwind_pid_free(us, el);
+    if (!process_is_alive(el)) {
```
r1viollet, line 579: why would we check this ? I'm not sure why this slipped in.

PR #245, opened 2023-04-18
Remove native library mode

r1viollet:
# What does this PR do?

Remove native library and self-unwind test

# Motivation

As we now have a better test framework with prof-correctness, we can remove the native library code paths. We would have to re-think the design if we want to integrate with other profilers.

# Additional Notes

# How to test the change?


sanchda:
Yup, it's a little sad to see this code go, but I do agree with all of your points and I think we'll have to reconsider the implementation + API anyway if this use-case becomes valid again.

PR #249, opened 2023-05-03
Fix: Remove incomplete frames

r1viollet:
# What does this PR do?

In case of incomplete stacks, we add an `[incomplete]` frame. We were adding this frame too often This PR adds new cases where we should consider the stack as complete.

# Motivation

Improve the look of frames by removing some of the `[incomplete]` qualifiers

### DIFF
--- a/src/unwind.cc
+++ b/src/unwind.cc
@@ -53,8 +53,15 @@ static bool is_ld(const std::string &path) {
 }
 
 static bool is_stack_complete(UnwindState *us) {
-  static constexpr std::array s_expected_root_frames{"_start"sv, "__clone"sv,
-                                                     "_exit"sv};
+  static constexpr std::array s_expected_root_frames{
+      "__clone"sv,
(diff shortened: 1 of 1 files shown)


src/unwind.cc
```diff
@@ -53,8 +53,15 @@ static bool is_ld(const std::string &path) {
 }
 
 static bool is_stack_complete(UnwindState *us) {
-  static constexpr std::array s_expected_root_frames{"_start"sv, "__clone"sv,
-                                                     "_exit"sv};
+  static constexpr std::array s_expected_root_frames{
```
nsavoire, line 56: **💭 thought:** ‏ wondering why some of these are needed (eg. `_exit`)

nsavoire (approved):

src/unwind.cc
r1viollet: I'm honestly not sure about that one :-)

PR #250, opened 2023-05-03
Fix management of file descriptors

r1viollet:
# What does this PR do?

In the case of whole host, we continuously get new files. As we were keeping file descriptors to all the files we were unwinding, we were quickly reaching file descriptor limits.

# Motivation

Avoid crashing / hitting the file descriptor limit.

# Additional Notes

Other strategies like cleaning up the file_info vector could be considered. I am happy to discuss this further.


sanchda (approved):
LGTM.  Definitely improves the code.

The `FileDescriptorHolder` class is a little bit weird to me, especially around the semantics of ownership.  It's slightly strange that one object gets to decide whether a separate object has taken ownership of a commonly-typed field after-the-fact.  That said, it's problematic for for cleanup to do otherwise.  I think this is one of those "unfortunate, but necessary" things.

src/ddprof_module_lib.cc
```diff
@@ -15,12 +15,33 @@
...
+      return ddres_warn(DD_WHAT_MODULE);
+    }
+    LG_DBG("[Mod] Success opening %s, ", path.data());
+    return ddres_init();
+  }
+
+  void take_ownership() { _fd = -1; }
+
+  ~FileDescriptorHolder() {
+    if (_fd > 0) {
```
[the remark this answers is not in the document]
r1viollet: Indeed, closing 1, 2, 3 does not make sense. The class could be more generic, though here it is mostly a small wrapper to avoid thinking about exit paths.
The `open` API does return a negative number in case of error. The check was tailored to the file's logic.
We should consider this if we ever want to extract this class.
Also I should probably have put the class in an anonymous namespace.

src/ddprof_module_lib.cc
```diff
@@ -15,12 +15,33 @@
...
+      close(_fd);
```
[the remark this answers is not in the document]
r1viollet: It could have been interesting to differentiate the error paths from the expected paths where `libdwfl` takes ownership of the file descriptors. Keeping this in mind.

src/ddprof_module_lib.cc
```diff
@@ -15,12 +15,33 @@
 #include <fcntl.h>
 #include <gelf.h>
 #include <libelf.h>
-#include <sys/stat.h>
-#include <sys/types.h>
 #include <unistd.h>
 
 namespace ddprof {
 
+class FileDescriptorHolder {
```
nsavoire, line 22: This could be a use case for [unique_resource](https://en.cppreference.com/w/cpp/experimental/unique_resource), but I don't know if it is widely available...

src/ddprof_module_lib.cc
```diff
@@ -15,12 +15,33 @@
...
+public:
+  FileDescriptorHolder() : _fd(-1) {}
+  DDRes open_file(std::string_view path) {
```
nsavoire, line 25: I think it would be best to follow `unique_ptr` API and having the resource creation outside the class

src/ddprof_module_lib.cc
```diff
@@ -15,12 +15,33 @@
...
+    _fd = ::open(path.data(), O_RDONLY);
+    if (_fd < 0) {
+      LG_WRN("[Mod] Couldn't open fd to module (%s)", path.data());
+      return ddres_warn(DD_WHAT_MODULE);
+    }
+    LG_DBG("[Mod] Success opening %s, ", path.data());
+    return ddres_init();
+  }
+
+  void take_ownership() { _fd = -1; }
```
nsavoire, line 35: I would follow `unique_ptr` API and name this `release` instead, because here I would expect for `take_ownership` that `FileDescriptorHolder` takes the ownership of the file descriptor

PR #252, opened 2023-05-11
Mask API key when printing parameters

r1viollet:
# What does this PR do?

When printing parameters in agentless mode, we are printing the API key.

# Motivation

Avoid possibly leaking a key in the logs.

### DIFF
--- a/src/ddprof_input.cc
+++ b/src/ddprof_input.cc
@@ -57,7 +57,7 @@
   "  -" #c ", --" #b ", (envvar: " #a ")",
 
 namespace {
-std::string mask_value(const std::string_view value) {
+std::string api_key_to_dbg_string(const std::string_view value) {
   if (value.size() != k_size_api_key) {
     return "invalid";
   }
@@ -66,7 +66,7 @@ std::string mask_value(const std::string_view value) {
   masked.replace(len - 4, 4, value.substr(len - 4));
   return masked;
 }
-}
+} // namespace
 
 #define X_PRNT(a, b, c, d, e, f, g, h, i)                                      \
   {                                                                            \
@@ -74,7 +74,7 @@ std::string mask_value(const std::string_view value) {
       if (strcmp(#b, "api_key") != 0) {                                        \
         PRINT_NFO("  " #b ": %s", (f)->i b);                                   \
       } else {                                                                 \
-        std::string masked_value = mask_value((f)->i b);                       \
+        std::string masked_value = api_key_to_dbg_string((f)->i b);            \
         PRINT_NFO("  " #b ": %s", masked_value.c_str());                       \
       }                                                                        \
     }                                                                          \
--- a/src/ddprof_input.cc
+++ b/src/ddprof_input.cc
@@ -56,11 +56,27 @@
 #define X_HLPK(a, b, c, d, e, f, g, h, i)                                      \
   "  -" #c ", --" #b ", (envvar: " #a ")",
 
-// Helpers for expanding the OPT_TABLE here
+namespace {
+std::string mask_value(const std::string_view value) {
+  if (value.size() != k_size_api_key) {
+    return "invalid";
+  }
+  size_t len = value.length();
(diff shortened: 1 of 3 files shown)


src/ddprof_input.cc
```diff
@@ -56,11 +56,27 @@
...
 
-// Helpers for expanding the OPT_TABLE here
+namespace {
+std::string mask_value(const std::string_view value) {
+  if (value.size() != k_size_api_key) {
+    return "invalid";
+  }
+  size_t len = value.length();
+  std::string masked(len, '*');
+  masked.replace(len - 4, 4, value.substr(len - 4));
```
sanchda, line 66: Good call; consistent with UI

src/ddprof_input.cc
```diff
@@ -56,11 +56,27 @@
 #define X_HLPK(a, b, c, d, e, f, g, h, i)                                      \
   "  -" #c ", --" #b ", (envvar: " #a ")",
 
-// Helpers for expanding the OPT_TABLE here
+namespace {
+std::string mask_value(const std::string_view value) {
+  if (value.size() != k_size_api_key) {
```
sanchda, line 61: Should we do stricter validation?  https://github.com/DataDog/dd-trace-go/blob/main/profiler/options.go#L188

I'm divided because this isn't really a feature we want to lean on, but there are so many places you can go wrong with copy+pasting things from docs/setup.  :)

src/ddprof_input.cc
```diff
@@ -56,11 +56,27 @@
...
+  masked.replace(len - 4, 4, value.substr(len - 4));
+  return masked;
+}
+}
+
 #define X_PRNT(a, b, c, d, e, f, g, h, i)                                      \
   {                                                                            \
     if ((f)->i b) {                                                            \
-      PRINT_NFO("  " #b ": %s", (f)->i b);                                     \
+      if (strcmp(#b, "api_key") != 0) {                                        \
```
sanchda, line 74: Adding a strcmp precisely once per key isn't so bad, since these get printed just once.

sanchda (dismissed):
LGTM--good call.  If this is something we're going to include, we ought to make sure it is sanitized.

My gut tells me it isn't so useful, but if managing secrets is important
 * Environment variables can be sanitized, specially after we fork off from user process
 * cmdline args can also be sanitized (https://github.com/sanchda/systems_experiments/tree/main/argv)

src/ddprof_input.cc
r1viollet: Yes! This does not make the top of the list, but it is high up there ^^

src/ddprof_input.cc
```diff
@@ -56,11 +56,27 @@
 #define X_HLPK(a, b, c, d, e, f, g, h, i)                                      \
   "  -" #c ", --" #b ", (envvar: " #a ")",
 
-// Helpers for expanding the OPT_TABLE here
+namespace {
+std::string mask_value(const std::string_view value) {
```
r1viollet, line 60: minor, the naming should be updated

src/ddprof_input.cc
r1viollet: I think changing CLI tool would help here. A lot of parameters would have stronger types and out of the box validation.
I'll check on that later.

src/ddprof_input.cc
```diff
@@ -56,11 +56,27 @@
 #define X_HLPK(a, b, c, d, e, f, g, h, i)                                      \
   "  -" #c ", --" #b ", (envvar: " #a ")",
 
-// Helpers for expanding the OPT_TABLE here
+namespace {
+std::string api_key_to_dbg_string(const std::string_view value) {
+  if (value.size() != k_size_api_key) {
+    return "invalid";
```
nsavoire, line 62: **🤓 nitpick:** ‏ why not keep last 4 chars whatever the input string ?

nsavoire (approved):

PR #254, opened 2023-05-15
Upgrade third party CLI11 version

r1viollet:
# What does this PR do?

Third party dependency upgrade

# Motivation

The aim is to use it as our CLI tool (instead of getopt).


nsavoire (dismissed):
LGTM !

r1viollet:
Really thankful that the CI is now failing in Github :+1: (I might not have expected this conversion issue)

test/simple_malloc.cc
```diff
@@ -177,99 +177,96 @@ int main(int argc, char *argv[]) {
...
-    }
-    wrapper(opts, stats[0]);
-    for (auto &t : threads) {
-      t.join();
-    }
-    auto pid = getpid();
-    for (auto &stat : stats) {
-      print_stats(pid, stat);
-    }
-  } catch (...) { return 1; }
```
r1viollet, line 274: @nsavoire I did not see the point of the try / catch (I was actually confused as to why this was failing), are you OK removing this ?

nsavoire: I think it's a good practice to catch all exceptions in main, here what was missing was printing errors I guess.
What was the error ?

nsavoire (approved):
LGTM !

test/simple_malloc.cc
r1viollet: As discussed, leaving exception propagate, makes the crash more explicit.

PR #256, opened 2023-05-15
Add python script to check for unsafe libc functions

nsavoire:
Some libc functions are unsafe to use in injected lib because their ABI is different between musl and glibc.
Add a small python script that check for the use of these functions.


r1viollet (dismissed):
LGTM

r1viollet (approved):
LGTM
Thanks for adding that in a way that works with the build system.

PR #263, opened 2023-05-17
Ensure that temporary process exits normally

nsavoire:
Killing the temporary intermediate process generates false memory leak reports in valgrind because some resources are not cleaned up (eg. local strings on the stack).
The fix is to install a sigterm handler in temp pid, and upon receiving SIGTERM, make temp pid exits normally through main.

### DIFF
--- a/src/daemonize.cc
+++ b/src/daemonize.cc
@@ -5,22 +5,31 @@
 
 #include "daemonize.hpp"
 
+#include <cstdlib>
 #include <fcntl.h>
 #include <sys/wait.h>
 #include <unistd.h>
 
 namespace ddprof {
-DaemonizeResult daemonize(std::function<void()> cleanup_function) {
+
+namespace {
+void handle_signal(int) {}
+DaemonizeResult daemonize_error() {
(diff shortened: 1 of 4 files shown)


nsavoire:
> CI seems to be KO :-), This seems like a cleaner approach!

No idea what was going on with CI. Seems OK after rebase.

src/daemonize.cc
```diff
@@ -29,15 +38,24 @@ DaemonizeResult daemonize(std::function<void()> cleanup_function) {
     temp_pid = getpid();
     if (pid_t child_pid = fork();
         child_pid) { // If I'm the temp PID again, enter branch
-      if (cleanup_function) {
-        cleanup_function();
+
+      struct sigaction sa;
+      if (sigemptyset(&sa.sa_mask) == -1) {
+        exit(1);
```
r1viollet, line 44: Minor: we should define a library logger for these cases.

r1viollet (approved):
LGTM! 
Good thinking, this will make leak analysis easier.

PR #264, opened 2023-05-17
Fix kernel frames configuration

r1viollet:
# What does this PR do?

Kernel frames were removed by default

# Motivation

Kernel frames were not showing up by default.

# How to test the change?

To be defined
The logic is confusing as it is a negation. Testing the generation of the configuration seems valuable.

### DIFF
--- a/src/perf.cc
+++ b/src/perf.cc
@@ -90,9 +90,9 @@ perf_event_attr perf_config_from_watcher(const PerfWatcher *watcher,
   attr.sample_type = watcher->sample_type;
   attr.sample_stack_user = watcher->sample_stack_size;
 
-  // If use_kernel is requested false --> exclude_kernel == true
-  if (watcher->options.use_kernel != PerfWatcherUseKernel::kOff)
-    attr.exclude_kernel = false;
+  // If use_kernel==off means we exclude_kernel
+  attr.exclude_kernel =
+      (watcher->options.use_kernel == PerfWatcherUseKernel::kOff);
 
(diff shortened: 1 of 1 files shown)


src/perf.cc
```diff
@@ -91,8 +91,8 @@ perf_event_attr perf_config_from_watcher(const PerfWatcher *watcher,
   attr.sample_stack_user = watcher->sample_stack_size;
 
   // If use_kernel is requested false --> exclude_kernel == true
-  if (watcher->options.use_kernel == PerfWatcherUseKernel::kTry)
-    attr.exclude_kernel = true;
```
r1viollet, line 95: the kTry config generates an extra attempt while excluding
```
  if (watcher->options.use_kernel == PerfWatcherUseKernel::kTry) {
    // duplicate the config, while excluding kernel
    ret_attr.push_back(ret_attr.back());
    ret_attr.back().exclude_kernel = true;
  }
```

src/perf.cc
```diff
@@ -91,8 +91,8 @@ perf_event_attr perf_config_from_watcher(const PerfWatcher *watcher,
...
+  if (watcher->options.use_kernel != PerfWatcherUseKernel::kOff)
+    attr.exclude_kernel = false;
```
nsavoire, line 95: **🤓 nitpick:** ‏ 
```suggestion
   attr.exclude_kernel = (watcher->options.use_kernel == PerfWatcherUseKernel::kOff);
```

PR #265, opened 2023-05-17
Clarify parser behavior

sanchda:
Parser was unterminated when multiple confs were given; fix by terminating.  This also left a few ambiguities open around whitespace handling in options, so make those explicit as well.  This also could have been fixed by allowing for an empty opt, but I didn't want to think about such an abomination.

# What does this PR do?

Changes the behavior of the parser and adds a small number of clarifying tests.

# Motivation

We had a warning during build.  This also makes the behavior more strict.

# Additional Notes

No additional notes

# How to test the change?

Tests included


sanchda:
Deleting in favor of #266 which is a more comprehensive fix.

PR #266, opened 2023-05-17
Fix small issues with event parser

nsavoire:
* Grammar was allowing multiple events separated by ';'
  but only the last one was taken into account.
  Now `watchers_from_str` return a vector of watchers.
* Make `mode` option less permissive
* Allow for empty configs: "; ; ;;;" is a valid event string
* Simplify stripping by skipping space/tabs in the lexer
* Convert presets to single string since multiple events
  can now be described by a string
* Remove warnings from lexer/bison

### DIFF
--- a/src/event_parser/event_parser.y
+++ b/src/event_parser/event_parser.y
@@ -10,10 +10,12 @@
 %parse-param {void *scanner} // actually yyscan_t
 
 %{
+#include <optional>
 #include <stdio.h>
 #include <stdint.h>
 #include <stdlib.h>
 #include <string>
+#include <vector>
 
 #include "event_config.hpp"
 #include "event_parser.h"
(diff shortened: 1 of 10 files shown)


src/event_parser/event_parser.y
```diff
@@ -46,21 +48,20 @@ EventConfMode mode_from_str(const std::string &str) {
...
-    }
-    if (l_str.find(c) != std::string::npos) {
+    } else if (l_str.find(c) != std::string::npos) {
         mode |= EventConfMode::kLiveCallgraph;
-    }
-    if (a_str.find(c) != std::string::npos) {
+    } else if (a_str.find(c) != std::string::npos) {
       mode |= EventConfMode::kAll;
+    } else {
+      return {};
```
sanchda, line 58: Yeah, this is probably the way to go.

I had originally made this permissive because I wanted to keep ddprof from throwing an error if the user specified a mode which is not compatible with that version of the profiler, but after seeing a few people getting frustrated with the overly-permissive mode specification, I see that was a mistake.

I think we should probably make this more specific in the future, like the user has to specify words instead of just characters.  Like `mode=metric|callgraph` or `mode=m|g`, but I don't think that precise syntax is a good one and I don't think it needs to be done now.

sanchda (approved):
LGTM; sorry for burdening you with these fixups, but thank you for proposing them.

PR #267, opened 2023-05-17
Fix aggregation of live allocated memory

nsavoire:
`pprof_aggregate` was multiplying passed value and count to obtain sample value, changed it to used directly passed value.


src/pprof/ddprof_pprof.cc
```diff
@@ -178,7 +178,7 @@ DDRes pprof_aggregate(const UnwindOutput *uw_output,
   ddog_prof_Profile *profile = pprof->_profile;
 
   int64_t values[DDPROF_PWT_LENGTH] = {};
-  values[watcher->pprof_sample_idx] = value * count;
```
r1viollet, line 181: Oh wow, thanks :facepalm:

r1viollet (approved):
LGTM

PR #269, opened 2023-05-23
Minor - allocation profiling error message

r1viollet:
# What does this PR do?

Improve the error message when process is not alive.

# Motivation

Reduce some of the noise.

# Additional Notes

Should we consider stopping the profiler ? 
I'm not sure if there are cases where process is not alive and we should continue:
- short lived processes where allocation profiling does not work ?

# How to test the change?

```
./ddprof some_bad_command_line
```

### DIFF
--- a/src/exe/main.cc
+++ b/src/exe/main.cc
@@ -14,6 +14,7 @@
 #include "defer.hpp"
 #include "ipc.hpp"
 #include "logger.hpp"
+#include "signal_helper.hpp"
 #include "tempfile.hpp"
 #include "timer.hpp"
 #include "user_override.hpp"
@@ -302,7 +303,6 @@ static int start_profiler_internal(DDProfContext *ctx, bool &is_profiler) {
       defer_parent_socket_close.release();
       return 0;
     }
-
     temp_pid = daemonize_res.temp_pid;
     defer_child_socket_close.release();
     defer_parent_socket_close.reset();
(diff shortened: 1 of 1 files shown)


src/exe/main.cc
```diff
@@ -381,12 +381,24 @@ static int start_profiler_internal(DDProfContext *ctx, bool &is_profiler) {
...
-        LG_WRN("Unable to connect to profiler library (target executable might "
-               "be statically linked and library cannot be preloaded). "
-               "Allocation profiling will be disabled.");
+        if (!process_is_alive(ctx->params.pid)) {
+          // Tell the user that process died
+          // Most of the time this is an invalid command line
+          LG_ERR(
+              "Target process(%d) is not alive. Check command line validity.",
+              ctx->params.pid);
+          // Avoiding returning
```
r1viollet, line 390: I'm not sure if we should stop here.

r1viollet: 99% of the time, the user just did a wrong command line.

r1viollet: @nsavoire do you think we should return -1 in this case ?

nsavoire: I think the first thing to fix is to print an error message when `execvp` fails.
Currently logs [emitted](https://github.com/DataDog/ddprof/blob/main/src/exe/main.cc#L467) in that case are not printed, probably because logger is not configured at this point...

nsavoire: In the case where process is dead and library connection failed, I would not print anything and let the execution continue.

sanchda: If we're here, then `pid` is set to the value of `getpid()` in the parent, so either
1. `kill()` failed because the pid is invalid (meaning the original process died)
2. `kill()` failed for some permissions-related reason???

I agree that the overwhelming commonality is that the user specified something that couldn't be executed, like a shell builtin or an incorrect command.

I suppose that rarely it may also be the case that the target process was very short-lived and the underlying cgroup didn't admit enough scheduler slots for this child process to emit the check before the parent process died.

Personally, I think we should stop here.  It used to make sense to keep going since we only had CPU profiling, but now I don't think there's such a guarantee, right?  Plus we're at the point now where we have the most information about the state of the failure.

On the other hand
1. we should emit a better error message at time of `execve()` failure in the parent
2. if the main application _was_ short-lived and it mattered, hopefully the user emitted their own errors

so maybe it's more useful to say more elsewhere rather than here.

r1viollet: On the small chance that there is a short lived process that forked, I will continue with a warning.
Thanks for the feedback! I'm also adding a message to exec failures.

nsavoire (approved):
LGTM !

PR #272, opened 2023-05-26
Remove unused code for child_fd management

sanchda:
This deletes semantically orphaned code.  It gets executed, but it has no purpose in the current codebase.


nsavoire (approved):
LGTM
Thanks for the cleanup !

Issue #220, opened 2023-03-03
Run ddprof as a sidecar container


r1viollet:
# Description

Allow `ddprof` to be deployed as a side container to monitor the activity of processes running on the host.

# Context

This was a request within the following discussion
https://github.com/DataDog/ddprof/issues/212

Within this ticket we should define the expected user experience

aristotelhs:
Although I mentioned the sidecar container is the previous post, the term sidecar was used to distinguish in ddprof running in the same container vs ddprof running in another container, side by side with the monitoring process. 

However, the ideal scenario for ddprof is the following:
1. Execute ddprof in a "privileged" (not sure if privileged is required of we could just make it work adding  CAP_SYSADMIN, SYSLOG capabilities).
2. Filter which processes ddprof would be able to monitor, using for example kubernetes annotations. 

This deployment has a number of advantages:
1. ddprof is the only process that runs in privileged mode so one can try to secure it better.
2. it is able to choose which pid to monitor using kubernetes mechanisms, so it can start doing continuous profiling as new pods are created.
3. no changes are required from the application container creation and deployment (perhaps some minor tuning so to include debugging symbols; not sure if ddprof supports debuginfo servers).

r1viollet:
Thanks for the feedback, I established an example [here](https://github.com/DataDog/ddprof-examples/tree/main/side-container) to help us iterate on a side-container deployment.

Keep in mind this is not the current way we advocate using `ddprof` and I would need to get more customer requests to work further on this.  Allocation profiling for instance currently does not work with this way of deploying the profiler.

However I'd be happy if you can give it a try and explain what is missing to make this usable for your use case.

aristotelhs:
This is for our first iteration of the deployment of ddprof on our k8s cluster:
1. we created a daemonset that executed ddprof as a container in all our nodes, in order to monitor the process running within the host.
2. ddprof was executed using the command: `ddprof -l notice -g yes` which sampled the whole host.

The profiles seems to work but there seems to be some questions on the unwinding. For example what does this anonymous means? 
<img width="377" alt="image" src="https://user-images.githubusercontent.com/48688196/236138327-ca1bbbc6-3d47-4090-ab7a-ce657e34eb12.png">

The other issue using this mode is that is hard to go from the processes back to the pod names. Is there a way that we can annotate the profiles that we collect with their containerID (at least. Best option is to have the pod name) so that we are able to go from logs/metrics to the corresponding profiles? 
[...]

PR #274, opened 2023-06-06
Label frames with container_id

r1viollet:
# What does this PR do?

Add a label with container_id information
We won't lookup short lived processes.

# Motivation

When viewing data from a host, we want to understand if our process is running within a container.
Refer to following issue: https://github.com/DataDog/ddprof/issues/220

# Additional Notes

This introduces a ProcessHdr object as an attempt to move some of the responsibilities to this object. 

I have a libdatadog PR open : https://github.com/DataDog/libdatadog/pull/169
Though I will not wait for the libdatadog code. Most of the code would be the same.

# How to test the change?

I added some unit tests that make sure we are able to read the container_id information.
An end to end test is still missing.


include/ddprof_process.hpp
```diff
@@ -0,0 +1,66 @@
...
+  uint64_t _sample_counter = {};
+private:
+  static DDRes read_cgroup_id(pid_t pid, std::string_view path_to_proc,
+                              CGroupId_t &cgroup);
+
+  pid_t _pid;
+  CGroupId_t _cgroup;
+};
+
+// todo ref counting
```
r1viollet, line 47: to be discussed, do we need ref counting for so little elements ?

src/ddprof_process.cc
```diff
@@ -0,0 +1,143 @@
...
+    // new process, parse cgroup
+    auto pair = _process_map.try_emplace(pid, pid);
+    if (pair.second) {
+      it = pair.first;
+    } else {
+      return std::nullopt;
+    }
+  }
+  ++(it->second._sample_counter);
+  if(!force && (it->second._sample_counter) < k_nb_samples_container_id_lookup) {
```
r1viollet, line 114: This is something I'll want to talk about.

r1viollet:
> This PR will add a custom slice dimension using the container-id: Example of user experience image

I slightly changed the labels. They now read: 
- None : not running within a container
- Unknown: We did not perform the lookup (as we only do so for long-lived pids)

docs/Build.md
```diff
@@ -34,3 +34,12 @@ MkBuildDir Rel
 RelCMake ../
 make -j 4 .
 ```
+
+### Updating libdatadog
+
+Head over to the libdatadog, do your changes and copy the library back to your vendor directory.
+
+```bash
+cp ${workdir}/libdatadog/headers/include/datadog/common.h ${workdir}/ddprof/vendor_gcc_unknown-linux-2.35_Debug/libdatadog-v2.1.0/include/datadog/common.h
```
r1viollet, line 43: not really relevant to the PR.

sanchda: Still worth mentioning, and much appreciated.  Every time I do this it still feels like a hack 😓

include/unwind_output.hpp
```diff
@@ -24,8 +25,10 @@ struct UnwindOutput {
   void clear() {
     locs.clear();
     is_incomplete = true;
+    container_id = "unknown";
```
sanchda, line 28: Ideally we'd use a default value that conforms to the right part of the Datadog BE.  Ultimately, I think the huge incremental value is going to be relating the sub-profiles to different things we harvest in the product.

That said, the immediate consumer is the profiling BE, and whenever we have a NULL context I think we have a challenge with upgrading it into a distinct type (for filtering/discoverability), so this is good.

src/ddprof_process.cc
```diff
@@ -0,0 +1,143 @@
...
+    return ddres_warn(DD_WHAT_CGROUP);
+  }
+  std::string line;
+  while (std::getline(cgroup_file, line)) {
+    container_id = container_id_from_line(line);
+    if (container_id) {
+      return {};
+    }
+  }
+  container_id = "none";
```
sanchda, line 69: Ah, interesting.  So our default is `"unknown"`, but failure to process this upgrades it to `"none"`.  I think there's value in having a semantic distinction and I think these two values are as good as any other, but in the future we may want to be more descriptive.

sanchda (approved):
Staggering.  This is a tremendous step in the right direction and major context that we were missing.  Very happy to see this in the code!

sanchda (dismissed):
Staggering.  This is a tremendous step in the right direction and major context that we were missing.  Very happy to see this in the code!

include/unwind_output.hpp
r1viollet: Yeah, we might need to revisit this. I'm making it at least more transparent in the next commit.

src/ddprof_process.cc
r1viollet: Thanks for highlighting that this was unclear. I added comments.
`none` means that it is operating outside a container.

r1viollet:
CI was flaky due to a metric push failing.

PR #279, opened 2023-06-28
Handle statsd paths in different format

r1viollet:
# What does this PR do?

Add an environment variable to the internal stat field
Adjust the format if it contains `unix://`
Revert removal of env var for pprof_prefix

# Motivation

Make it easier to set the internal stats field.
Avoid breaking relenv
Fix prof-correctness tests

### DIFF
--- a/include/statsd.hpp
+++ b/include/statsd.hpp
@@ -29,3 +29,5 @@ DDRes statsd_close(int);
 
 /* Private */
 DDRes statsd_listen(std::string_view path, int *fd);
+
+std::string_view adjust_uds_path(std::string_view path);
--- a/include/statsd.hpp
+++ b/include/statsd.hpp
@@ -29,5 +29,3 @@ DDRes statsd_close(int);
 
 /* Private */
 DDRes statsd_listen(std::string_view path, int *fd);
-
-std::string_view adjust_uds_path(std::string_view path);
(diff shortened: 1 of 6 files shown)


sanchda (approved):
Yeah, this is a good idea.  Thank you!

r1viollet:
> Yeah, this is a good idea. Thank you!

Thanks. CI was failing due to documentation updates. I think PR is now ready!

include/statsd.hpp
```diff
@@ -29,3 +29,5 @@ DDRes statsd_close(int);
 
 /* Private */
 DDRes statsd_listen(std::string_view path, int *fd);
+
+std::string_view adjust_uds_path(std::string_view path);
```
nsavoire, line 33: Does this need to be exposed in header ?

src/statsd.cc
```diff
@@ -15,10 +15,18 @@
 #include <sys/un.h>
 #include <unistd.h>
 
+std::string_view adjust_uds_path(std::string_view path) {
+  // Check if the socket_path starts with unix://
+  if (path.substr(0, 7) == "unix://") {
+    path.remove_prefix(7);
+  }
+  return path;
```
nsavoire, line 23:
```suggestion
  static constexpr std::string_view prefix{"unix://"};
  if (path.starts_with(prefix)) {
    path.remove_prefix(prefix.length());
  }
  return path;
```

nsavoire (dismissed):

nsavoire (approved):
LGTM!

Issue #276, opened 2023-06-11
Attempting to read outside of stack


aristotelhs:
Hello, 
we are trying to use ddprof for our continuous profiling. Currently we are deploying ddprof as a sidecar container that is used to profile the main process. The main process is written in Julia and it is using the Julia jit for the execution of the end code. 
Checking the result in the web interface, we can see that some frames are missing. Running ddprof in debug mode we notice a lot of error messages like:
```
<DEBUG>Jun 11 12:07:19.028064 ddprof[143]: Attempting to read outside of stack 0x7ffd39f95258 from 47, (0x7ffd39f8d0d0, 0x7ffd39f950d0)[0x7f418fb376b8, 0x7f418fb3f6b8]
<DEBUG>Jun 11 12:07:19.028067 ddprof[143]: Attempting to read outside of stack 0x7ffd39f95260 from 47, (0x7ffd39f8d0d0, 0x7ffd39f950d0)[0x7f418fb376b8, 0x7f418fb3f6b8]
<DEBUG>Jun 11 12:07:19.028070 ddprof[143]: Attempting to read outside of stack 0x7ffd39f95268 from 47, (0x7ffd39f8d0d0, 0x7ffd39f950d0)[0x7f418fb376b8, 0x7f418fb3f6b8]
<DEBUG>Jun 11 12:07:19.028073 ddprof[143]: Attempting to read outside of stack 0x7ffd39f95278 from 47, (0x7ffd39f8d0d0, 0x7ffd39f950d0)[0x7f418fb376b8, 0x7f418fb3f6b8]
<DEBUG>Jun 11 12:07:19.083331 ddprof[143]: Attempting to read outside of stack 0x7ffd39f92878 from 47, (0x7ffd39f8a730, 0x7ffd39f92730)[0x7f418faff790, 0x7f418fb07790]
<DEBUG>Jun 11 12:07:19.083354 ddprof[143]: Attempting to read outside of stack 0x7ffd39f928a0 from 47, (0x7ffd39f8a730, 0x7ffd39f92730)[0x7f418faff790, 0x7f418fb07790]
<DEBUG>Jun 11 12:07:19.083359 ddprof[143]: Attempting to read outside of stack 0x7ffd39f92880 from 47, (0x7ffd39f8a730, 0x7ffd39f92730)[0x7f418faff790, 0x7f418fb07790]
[...]

PR #280, opened 2023-06-29
JIT Symbols - update symbolization priorities

r1viollet:
# What does this PR do?

Give higher priority to JIT lookup
In the case of jitdump information, we don't need to look for the DSO.

# Motivation

Discussions in : https://github.com/DataDog/ddprof/issues/276
The logs are suggesting we are failing in DSO lookup before checking the JIT information.

# Additional Notes

Not finding the DSO is still strange, especially as the logs highlight a valid executable dso
```
7f0863300000-7f0863400000 r-xp 00000000 00:00 0 
```
I'm assuming that the DSOs are changing at high frequency. 

# How to test the change?


src/unwind_dwfl.cc
```diff
@@ -270,30 +285,38 @@ static DDRes add_dwfl_frame(UnwindState *us, const Dso &dso, ElfAddress_t pc,
...
+                                      std::string_view jitdump_path,
+                                      bool &symbol_found) {
+  assert(!jitdump_path.empty());
+  SymbolHdr &unwind_symbol_hdr = us->symbol_hdr;
+  SymbolTable &symbol_table = unwind_symbol_hdr._symbol_table;
+  RuntimeSymbolLookup &runtime_symbol_lookup =
+      unwind_symbol_hdr._runtime_symbol_lookup;
+  SymbolIdx_t symbol_idx =
+      runtime_symbol_lookup.get_or_insert(us->pid, pc, symbol_table);
+  symbol_found = symbol_idx == -1 ? false : true;
```
sanchda, line 298:
```suggestion
  symbol_found = symbol_idx != -1 ?;
```

src/unwind_dwfl.cc
```diff
@@ -270,30 +285,38 @@ static DDRes add_dwfl_frame(UnwindState *us, const Dso &dso, ElfAddress_t pc,
...
+  if (symbol_found) {
+    return add_frame(symbol_idx, -1, pc, us);
```
sanchda, line 300: I think this is probably right for now.  I can imagine a point in the future where we have some mapping information to pass, but that point isn't now.

sanchda (approved):
LGTM

r1viollet:
After discussion with @nsavoire, I'm closing this PR. We will mostly allow more backpopulates for now and investigate further why we were missing mmap events.

PR #281, opened 2023-06-29
Symbol cache fix off by one error

r1viollet:
# What does this PR do?

- The symbol cache had a off by one error in some cases.
- When libdwfl did not git us the size of a symbol, we would use pc+1 (which was wrong)

# Motivation

Fix for flaky test

### DIFF
--- a/src/dwfl_symbol_lookup.cc
+++ b/src/dwfl_symbol_lookup.cc
@@ -81,14 +81,16 @@ SymbolIdx_t DwflSymbolLookup::get_or_insert(const DDProfMod &ddprof_mod,
     return find_res.first->second.get_symbol_idx();
   }
 
-  return insert(ddprof_mod, table, dso_symbol_lookup, process_pc, dso, map);
+  return insert(ddprof_mod, table, dso_symbol_lookup, process_pc, dso, map,
+                find_res.first);
 }
 
 SymbolIdx_t DwflSymbolLookup::insert(const DDProfMod &ddprof_mod,
                                      SymbolTable &table,
                                      DsoSymbolLookup &dso_symbol_lookup,
                                      ProcessAddress_t process_pc,
-                                     const Dso &dso, SymbolMap &map) {
+                                     const Dso &dso, SymbolMap &map,
+                                     SymbolMap::It &it) {
 
   Symbol symbol;
   GElf_Sym elf_sym;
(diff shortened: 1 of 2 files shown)


src/dwfl_symbol_lookup.cc
```diff
@@ -128,6 +130,23 @@ SymbolIdx_t DwflSymbolLookup::insert(const DDProfMod &ddprof_mod,
     assert(0);
   }
 
+  // previous element was not within bounds, but it still matches
```
r1viollet, line 133: As discussed, it would be better to understand why the size of the symbol is returned as 0

src/dwfl_symbol_lookup.cc
```diff
@@ -81,14 +81,16 @@ SymbolIdx_t DwflSymbolLookup::get_or_insert(const DDProfMod &ddprof_mod,
...
+                find_res.first);
 }
 
 SymbolIdx_t DwflSymbolLookup::insert(const DDProfMod &ddprof_mod,
                                      SymbolTable &table,
                                      DsoSymbolLookup &dso_symbol_lookup,
                                      ProcessAddress_t process_pc,
-                                     const Dso &dso, SymbolMap &map) {
+                                     const Dso &dso, SymbolMap &map,
+                                     SymbolMap::It &it) {
```
r1viollet, line 93: minor: this could be previous symbol or closest symbol to improve the naming

src/dwfl_symbol_lookup.cc
```diff
@@ -128,6 +130,23 @@ SymbolIdx_t DwflSymbolLookup::insert(const DDProfMod &ddprof_mod,
...
+  // previous element was not within bounds, but it still matches
+  // We adjust assuming this is still the same symbol
+  if (it != map.end() &&
+      table[it->second.get_symbol_idx()]._symname == symbol._symname) {
+    if (elf_pc > it->second.get_end()
+        // $x and $d symbols are compiler and linker generated symbols and they
+        // are not part of your original source code. They are added as part of
+        // the binary creation process to handle interworking between ARM and
+        // Thumb code. They can be inserted in different places
+        && (symbol._symname.size() != 2 || symbol._symname[0] == '$')) {
```
r1viollet, line 142: There is a bug here.

src/dwfl_symbol_lookup.cc
r1viollet: Also const it is valid (though the data is not const)

PR #284, opened 2023-07-05
Configure stack size

r1viollet:
# What does this PR do?

Add a parameter to configure the size of samples.

# Motivation

Ensure we can unwind deeper stacks.
Discussions in https://github.com/DataDog/ddprof/issues/276

# Additional Notes

For now the maximum value allowed by linux is the closest value to USHORT_MAX, 8 byte aligned:
`65528`
If this is not enough we should consider a different kernel API.

# How to test the change?

A test with deep stacks was added. I will run this in CI.
https://github.com/DataDog/ddprof-build/pull/101

### DIFF
--- a/include/ddprof_defs.hpp
+++ b/include/ddprof_defs.hpp
@@ -24,7 +24,7 @@ constexpr uint32_t k_default_perf_stack_sample_size = (4096UL * 8);
 // considering sample size, we adjust the size of ring buffers.
 // Following is considered as a minimum number of samples to be fit in the
 // ring buffer.
-constexpr auto k_min_number_samples_per_ring_buffer = 8;
+constexpr auto k_min_number_samples_per_ring_buffer = 7;
 
 constexpr int k_size_api_key = 32;
 
--- a/include/ddprof_defs.hpp
+++ b/include/ddprof_defs.hpp
@@ -19,12 +19,12 @@
 // as `size_stack` might be smaller then the requested size
 // The type is uint32 to be consistent with the perf_event interface
 // Check linux sources for a reference to the sample size check
-constexpr uint32_t k_default_perf_stack_sample_size = (4096UL * 8);
+constexpr uint32_t k_default_perf_stack_sample_size = 32000;
 
 // considering sample size, we adjust the size of ring buffers.
 // Following is considered as a minimum number of samples to be fit in the
 // ring buffer.
-constexpr auto k_min_number_samples_per_ring_buffer = 7;
+constexpr auto k_min_number_samples_per_ring_buffer = 8;
 
 constexpr int k_size_api_key = 32;
 
--- a/include/ddprof_defs.hpp
+++ b/include/ddprof_defs.hpp
@@ -13,6 +13,9 @@
 // Maximum depth for a single stack
 #define DD_MAX_STACK_DEPTH 512
 
+// default stack size for perf_event_open
+constexpr unsigned long k_default_perf_sample_stack_user = (4096UL * 8);
+
 constexpr int k_size_api_key = 32;
 
 // Linux Inode type
--- a/include/ddprof_defs.hpp
+++ b/include/ddprof_defs.hpp
@@ -14,9 +14,17 @@
 #define DD_MAX_STACK_DEPTH 512
 
 // Sample stack size must a multiple of 8 and strictly inferior to 2^16
+// Note that since maximum perf_event_hdr size is 2^16-1 and there are other
+// data/headers in perf_event struct, actual maximum stack sample size returned
+// as `size_stack` might be smaller then the requested size
 // The type is uint32 to be consistent with the perf_event interface
-// Check linux sources for a reference to this check
-constexpr uint32_t k_default_perf_sample_stack_user = (4096UL * 8);
+// Check linux sources for a reference to the sample size check
+constexpr uint32_t k_default_perf_stack_sample_size = (4096UL * 8);
+
+// considering sample size, we adjust the size of ring buffers.
+// Following is considered as a minimum number of samples to be fit in the
+// ring buffer.
+constexpr auto k_min_number_samples_per_ring_buffer = 8;
 
 constexpr int k_size_api_key = 32;
 
--- a/include/ddprof_defs.hpp
+++ b/include/ddprof_defs.hpp
@@ -13,8 +13,10 @@
 // Maximum depth for a single stack
 #define DD_MAX_STACK_DEPTH 512
 
-// default stack size for perf_event_open
-constexpr unsigned long k_default_perf_sample_stack_user = (4096UL * 8);
+// Sample stack size must a multiple of 8 and strictly inferior to 2^16
+// The type is uint32 to be consistent with the perf_event interface
+// Check linux sources for a reference to this check
+constexpr uint32_t k_default_perf_sample_stack_user = (4096UL * 8);
 
 constexpr int k_size_api_key = 32;
 
--- a/include/lib/allocation_event.hpp
+++ b/include/lib/allocation_event.hpp
@@ -13,7 +13,6 @@ namespace ddprof {
 // This represent a sampled allocation.
 // We Keep the same layout as a perf event to unify the code paths.
 struct AllocationEvent {
-  static constexpr unsigned k_data_fake_size = 8; // 8 to keep aligned struct
   perf_event_header hdr;
   struct sample_id sample_id;
   uint64_t addr; /* if PERF_SAMPLE_ADDR */
@@ -27,7 +26,7 @@ struct AllocationEvent {
 // An extra field is added after the end to communicate the dyn_size
 //  uint64_t dyn_size_stack;
 
-static inline size_t SizeOfAllocationEvent(uint32_t stack_sample_user) {
+inline size_t sizeof_allocation_event(uint32_t stack_sample_user) {
   // stack_sample_user is 8 byte aligned
   // (Size of the event) + (stack size - 1) + sizeof(dyn_size field)
   return sizeof(AllocationEvent) + stack_sample_user + sizeof(uint64_t);
--- a/include/lib/allocation_event.hpp
+++ b/include/lib/allocation_event.hpp
@@ -1,0 +1,24 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0. This product includes software
+// developed at Datadog (https://www.datadoghq.com/). Copyright 2023-Present
+// Datadog, Inc.
+
+#pragma once
+
+#include <linux/perf_event.h>
+
+namespace ddprof {
+
+struct AllocationEvent {
+  perf_event_header hdr;
+  struct sample_id sample_id;
+  uint64_t addr; /* if PERF_SAMPLE_ADDR */
+  uint64_t period;
+  uint64_t abi;                   /* if PERF_SAMPLE_REGS_USER */
+  uint64_t regs[PERF_REGS_COUNT]; /* if PERF_SAMPLE_REGS_USER */
+  uint64_t size;                  /* if PERF_SAMPLE_STACK_USER */
+  std::byte data[1]; /* if PERF_SAMPLE_STACK_USER, the actual size will be
+                                     determined at runtime */
+  uint64_t dyn_size; /* if PERF_SAMPLE_STACK_USER && size != 0 */
+};
+} // namespace ddprof
--- a/include/lib/allocation_event.hpp
+++ b/include/lib/allocation_event.hpp
@@ -18,18 +18,18 @@ struct AllocationEvent {
   struct sample_id sample_id;
   uint64_t addr; /* if PERF_SAMPLE_ADDR */
   uint64_t period;
-  uint64_t abi;                   /* if PERF_SAMPLE_REGS_USER */
-  uint64_t regs[PERF_REGS_COUNT]; /* if PERF_SAMPLE_REGS_USER */
-  uint64_t size_stack;                  /* if PERF_SAMPLE_STACK_USER */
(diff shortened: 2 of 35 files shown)


include/lib/allocation_event.hpp
```diff
@@ -0,0 +1,35 @@
...
+  static constexpr unsigned k_data_fake_size = 8; // 8 to keep aligned struct
+  perf_event_header hdr;
+  struct sample_id sample_id;
+  uint64_t addr; /* if PERF_SAMPLE_ADDR */
+  uint64_t period;
+  uint64_t abi;                   /* if PERF_SAMPLE_REGS_USER */
+  uint64_t regs[PERF_REGS_COUNT]; /* if PERF_SAMPLE_REGS_USER */
+  uint64_t size_stack;                  /* if PERF_SAMPLE_STACK_USER */
+  std::byte data[k_data_fake_size]; /* requires PERF_SAMPLE_STACK_USER, dyn size will contain the actual size */
+};
```
r1viollet, line 25: @nsavoire any better ideas for this part ?

nsavoire: I miss the point of `data[k_data_fake_size]` member here.
> // 8 to keep aligned struct

In which way ?

src/pevent_lib.cc
```diff
@@ -55,11 +58,34 @@ static void display_system_config(void) {
   }
 }
 
+int pevent_compute_min_mmap_order(int min_buffer_size_order,
```
r1viollet, line 61: @nsavoire @sanchda would you agree with computing the size of the ring buffer based on the requested sample size ?

include/lib/allocation_event.hpp
r1viollet: fixed

nsavoire:
I find that the naming `sample_stack_user` used throughout the PR is not very good are telling what the parameter really is.
Perhaps something like: `stack_sample_size` or `user_stack_sample_size`

include/ddprof_defs.hpp
```diff
@@ -13,6 +13,11 @@
 // Maximum depth for a single stack
 #define DD_MAX_STACK_DEPTH 512
 
+// Sample stack size must a multiple of 8 and strictly inferior to 2^16
+// The type is uint32 to be consistent with the perf_event interface
+// Check linux sources for a reference to this check
```
nsavoire, line 18: Why not leave the other part of the comment ?
>Note that since maximum perf_event_hdr size is 2^16-1 and there are other
> data/headers in perf_event struct, actual maximum stack sample size returned
> as `size_stack` might be smaller then the requested size

include/lib/allocation_event.hpp
```diff
@@ -0,0 +1,35 @@
...
+
+#include <linux/perf_event.h>
+
+namespace ddprof {
+
+// AllocationEvent
+// This represent a sampled allocation.
+// We Keep the same layout as a perf event to unify the code paths.
+struct AllocationEvent {
+  static constexpr unsigned k_data_fake_size = 8; // 8 to keep aligned struct
```
nsavoire, line 16: useless now

include/lib/allocation_event.hpp
```diff
@@ -0,0 +1,35 @@
...
+  uint64_t abi;                   /* if PERF_SAMPLE_REGS_USER */
+  uint64_t regs[PERF_REGS_COUNT]; /* if PERF_SAMPLE_REGS_USER */
+  uint64_t size_stack;            /* if PERF_SAMPLE_STACK_USER */
+  std::byte data[];               /* requires PERF_SAMPLE_STACK_USER, dyn size
+                                      will contain the actual size */
+};
+// An extra field is added after the end to communicate the dyn_size
+//  uint64_t dyn_size_stack;
+
+static inline size_t SizeOfAllocationEvent(uint32_t stack_sample_user) {
```
nsavoire, line 30: `static` is not needed
`SizeOfAllocationEvent` does not follow the standard naming for free functions

src/ddprof_cli.cc
```diff
@@ -59,6 +59,22 @@ void write_config_file(const CLI::App &app, const std::string &file_path) {
...
+
+using Validator = CLI::Validator;
+struct SampleStackSizeValidator : public Validator {
+  SampleStackSizeValidator() {
+    name_ = "SAMPLE_STACK_SIZE";
+    func_ = [](const std::string &str) {
+      int value = std::stoi(str);
+      if (value >= USHRT_MAX || value % 8 != 0) {
+        return std::string("Invalid sample_stack_user value. Value should be "
+                           "less than USHRT_MAX and a multiple of 8.");
```
nsavoire, line 71: perhaps replace `USHRT_MAX` with the actual value ?

r1viollet:
> I find that the naming `sample_stack_user` used throughout the PR is not very good at telling what the parameter really is. Perhaps something like: `stack_sample_size` or `user_stack_sample_size`

I thought the same, though I wanted the consistency with the perf_event_open API
```
       sample_stack_user (since Linux 3.7)
              This defines the size of the user stack to dump if
              PERF_SAMPLE_STACK_USER is specified.
```
I guess we can forget about consistency and make it more user friendly ?

src/ddprof_cmdline.cc
```diff
@@ -135,19 +135,24 @@ static bool watcher_from_config(EventConf *conf, PerfWatcher *watcher) {
...
+                       uint32_t sample_stack_user) {
   std::vector<EventConf> configs;
   if (EventConf_parse(str, configs) != 0) {
     return false;
   }
 
   for (auto &conf : configs) {
+    // todo{@r1viollet}: ideally here we would want to change the template
+    // value of the event conf. Instead we override the conf value.
+    if (conf.sample_stack_user == k_default_perf_sample_stack_user) {
```
nsavoire, line 148: Logic is strange, what happens if I do `ddprof --sample-stack-user 65528 -e sCPU,s=32768` ?

src/pevent_lib.cc
```diff
@@ -55,11 +58,34 @@ static void display_system_config(void) {
...
+                                  uint32_t sample_stack_user) {
+  int ret_order = min_buffer_size_order;
+  // perf events and allocation events should be roughtly the same size
+  size_t single_event_size =
+      sample_stack_user - 1 + sizeof(ddprof::AllocationEvent);
```
nsavoire, line 66: Why subtracting `1` ?

src/pevent_lib.cc
nsavoire: ok for me

nsavoire, line 61: why not pass the `k_min_number_samples_per_ring_buffer` as argument to the function ?

src/ddprof_cmdline.cc
r1viollet: `s=32768` wins: conf will hold the value from the watcher configuration, so global parameter is ignored. I'm not a fan of this. Alternative ideas are welcome.

src/pevent_lib.cc
r1viollet: This sounds reasonable indeed.

src/pevent_lib.cc
```diff
@@ -55,11 +58,34 @@ static void display_system_config(void) {
...
+  // Ensure we can at least fit 8 samples within one buffer
+  while ((perf_mmap_size(ret_order) / single_event_size) <
```
nsavoire, line 68: `perf_mmap_size` returns `(2^order+1) * PAGE_SIZE`, but one page is used for metadata, hence the usable size is `perf_mmap_size(ret_order)-PAGE_SIZE`.
Note that before this PR, ring buffer could not hold 8 samples: ring buffer usable size was 262144 and 8 samples would be `(4096*8+perf_sample_data)*8`

test/deep_stacks.cc
```diff
@@ -0,0 +1,29 @@
+#include <chrono>
```
nsavoire, line 1: License ?

test/deep_stacks.sh
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env bash
```
nsavoire, line 1: License ?

test/deep_stacks.sh
```diff
@@ -0,0 +1,18 @@
...
+
```
nsavoire, line 2: `set -euo pipefail`

test/deep_stacks.sh
```diff
@@ -0,0 +1,18 @@
...
+CONFIG_DEEP_STACK="../test/configs/deep_stacks.toml"
```
nsavoire, line 3: Is a config file really useful here ?

test/deep_stacks.cc
```diff
@@ -0,0 +1,29 @@
...
+#include <iostream>
+#include <string>
+
+template <int N> std::string compute() {
```
nsavoire, line 5: Compiler will probably fully inline all the function and the recursive calls (and perhaps even figure out directly the final value).

test/deep_stacks.cc
```diff
@@ -0,0 +1,29 @@
...
+  }
+  return ret_arr;
+}
+
+int main() {
+  using namespace std::chrono;
+  auto start = high_resolution_clock::now();
+  auto end = start + seconds(2); // set a time limit of 10 seconds
+  while (high_resolution_clock::now() < end) {
+    volatile auto str = compute<3000>();
```
nsavoire, line 25: `volatile` is almost always a code smell in c++, here you shoud use `ddprof::DoNotOptimize`.

test/deep_stacks.sh
```diff
@@ -0,0 +1,18 @@
...
+
+./ddprof --config ${CONFIG_DEEP_STACK} ./test/deep_stacks | tee ./log_ddprof.txt
+
+# Check for truncated stack traces
+truncated_input=$(awk -F': ' '/datadog.profiling.native.unwind.stack.truncated_input/ { print $NF }' ./log_ddprof.txt)
```
nsavoire, line 8: Why not test that the stacks are correct ?

r1viollet: I would need to open source prof-correctness to avoid duplicating some of the logics for this.

src/ddprof_cmdline.cc
nsavoire: I would indeed expect that event parameter takes precedence, but that's not what is written in this function.
I believe that with the current code, event will have its sample stack size set to 65528.

test/deep_stacks.sh
nsavoire: Yes, but in its current state this test does not test much (and is quite complicated).
In release mode, everything is probably inlined in `main` function.

src/ddprof_cmdline.cc
r1viollet: Indeed, I did not account for the case where user would re-force the value to the default.

src/pevent_lib.cc
r1viollet: Indeed. 8 sounds like such a low number...
I would be in favour of increasing min size to order 7. This could come with more mmap failures though.

nsavoire: Why not slightly decrease the default sample stack size to make sure that 8 samples fit into the buffer ?

r1viollet: I'll adjust the log message to help users with this.

r1viollet:
The failures in CI are related to mmap failures..

test/deep_stacks.sh
r1viollet: I have some amount of trouble with this setting, though I understand the value. I added it.

test/deep_stacks.sh
r1viollet: I'm not sure this is useful ^^

src/pevent_lib.cc
r1viollet: Good point, adjusted.

src/ddprof_cmdline.cc
r1viollet: I fixed this implementing the todo.

r1viollet:
> > I find that the naming `sample_stack_user` used throughout the PR is not very good at telling what the parameter really is. Perhaps something like: `stack_sample_size` or `user_stack_sample_size`
> 
> I thought the same, though I wanted the consistency with the perf_event_open API
> 
> ```
>        sample_stack_user (since Linux 3.7)
>               This defines the size of the user stack to dump if
>               PERF_SAMPLE_STACK_USER is specified.
> ```
> 
> I guess we can forget about consistency and make it more user friendly ?

I adjusted the naming, we are now inconsistent, though the setting is easier to understand.

test/deep_stacks.sh
r1viollet: Trying out release mode, everything looks OK
image

r1viollet:
> The failures in CI are related to mmap failures..

I reverted the increase in buffer size :-(

test/deep_stacks.sh
r1viollet: I did add a "no inline" attribute to make sure. Thanks for mentioning this.

include/ddprof_defs.hpp
```diff
@@ -13,6 +13,19 @@
...
+// data/headers in perf_event struct, actual maximum stack sample size returned
+// as `size_stack` might be smaller then the requested size
+// The type is uint32 to be consistent with the perf_event interface
+// Check linux sources for a reference to the sample size check
+constexpr uint32_t k_default_perf_stack_sample_size = (4096UL * 8);
+
+// considering sample size, we adjust the size of ring buffers.
+// Following is considered as a minimum number of samples to be fit in the
+// ring buffer.
+constexpr auto k_min_number_samples_per_ring_buffer = 7;
```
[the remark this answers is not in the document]
r1viollet: I pushed a change on default sample sizes.

nsavoire (approved):
LGTM !

PR #285, opened 2023-07-07
Backpopulate - allow more useful backpopulates

r1viollet:
# What does this PR do?

If the state of the maps is changing and we are missing events we want to still allow backpopulates.

# Motivation

Discussions in https://github.com/DataDog/ddprof/issues/276
We can see that we are often trying to backpopulate.


nsavoire (approved):
LGTM!

PR #287, opened 2023-07-17
Improve performance of allocation profiling

r1viollet:
# What does this PR do?

- Adjust the pseudo random generator to be thread local
- Change the generator to be smaller
- Remove the lock from the allocation profiling path

Credit for this idea goes to @nsavoire 

# Motivation

Improve performance of allocation profiling. This really brings down the allocation profiling to numbers that are comparable. 
```
Benchmark                                  Time             CPU   Iterations
----------------------------------------------------------------------------
BM_ThreadedAllocations_NoTracking     343085 ns       172913 ns         4525
BM_ThreadedAllocations_Tracking       483263 ns       244706 ns         2886
```

The numbers were 4x worse prior to this change.

After this change I will need to focus on the deallocation code path.

# Additional Notes

This needs extensive testing.
Risks are:
- Accuracy 
- Thread safety

# How to test the change?

The benchmark only tests parts of this. We will need to be careful when deploying this change.

### DIFF
--- a/src/lib/allocation_tracker.cc
+++ b/src/lib/allocation_tracker.cc
@@ -57,7 +57,7 @@ private:
 AllocationTracker *AllocationTracker::_instance;
 thread_local TrackerThreadLocalState AllocationTracker::_tl_state;
 
-AllocationTracker::AllocationTracker() : _gen(std::random_device{}()) {}
+AllocationTracker::AllocationTracker() {}
 
 AllocationTracker *AllocationTracker::create_instance() {
   static AllocationTracker tracker;
@@ -156,10 +156,6 @@ void AllocationTracker::track_allocation(uintptr_t addr, size_t size,
     return;
   }
 
-  // \fixme{nsavoire} reduce the scope of the mutex:
-  // change prng/make it thread local
-  std::lock_guard lock{_state.mutex};
-
   // recheck if profiling is enabled
   if (!_state.track_allocations) {
     return;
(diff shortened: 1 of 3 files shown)


src/lib/allocation_tracker.cc
```diff
@@ -194,6 +190,10 @@ void AllocationTracker::track_allocation(uintptr_t addr, size_t size,
   free_on_consecutive_failures(success);
 
   if (success && _state.track_deallocations) {
+    // \fixme{r1viollet} adjust set to be lock free
```
richardstartin, line 193: this could probably be replaced with a rather simple sparse bitset over the least significant 48 bits of the pointers, which would allow CAS operations on pages of the bitset rather than a lock.

test/allocation_tracker-bench.cc
```diff
@@ -39,7 +44,7 @@ void perform_memory_operations(bool track_allocations,
   std::mt19937 gen(rd());
 
   for (auto _ : state) {
-    state.PauseTiming();
+    //    state.PauseTiming();
```
richardstartin, line 47: why does this need to be commented out?

richardstartin (approved):
The change makes sense to me

sanchda (approved):

test/allocation_tracker-bench.cc
r1viollet: Initially I was measuring the cost of deallocation. In this PR I was optimizing the allocation code path.
I think the correct thing to do is have different benchmarks. I am continuing this work and I'll revisit this.

src/lib/allocation_tracker.cc
r1viollet: I like this, especially as the impact of collisions are easy to manage. I'll probably make it smaller than that.

PR #293, opened 2023-08-02
Julia uses start_task as the top frame

sanchda:
# What does this PR do?

See #292.  This is a clone of that PR that should trigger CI.


sanchda:
Closing because this branch caused the tests to pass against the underlying commit, so #292 became green.

PR #295, opened 2023-08-24
Jitdump relative path fix

r1viollet:
# What does this PR do?

When reading from the proc maps, the path to the jitdump file can be relative. We now will check if it is a relative or absolute path. Depending on the case we will check either the /proc/pid/cwd or the /proc/pid/root directory.

# Motivation

The logs in the following issue were highlighting that the jitdump file was not found:
https://github.com/DataDog/ddprof/issues/276

# Additional Notes

N/A

# How to test the change?

A unit test was added.
I did not reproduce this locally. I will have to check if this fixes the reported issue.


nsavoire (approved):
LGTM

PR #297, opened 2023-08-28
Jitdump fix for wholehost

r1viollet:
# What does this PR do?

In wholehost we were not considering the jitdump files. The format was required to include the PID in the name of the file. However the namespace's PID is not known in wholehost. This fix relaxes the test to just look for jit- and .dump in the name of the file.

# Motivation

A debug session with @aristotelhs

### DIFF
--- a/src/dso.cc
+++ b/src/dso.cc
@@ -83,13 +83,12 @@ Dso::Dso(pid_t pid, ElfAddress_t start, ElfAddress_t end, ElfAddress_t pgoff,
 bool Dso::is_jit_dump_str(std::string_view file_path, pid_t pid) {
   // test if we finish by .dump before creating a string
   if (file_path.ends_with(".dump")) {
-    // llvm uses this format
-    std::string jit_dump_str = string_format("jit-%d.dump", pid);
-    // this is to remove a gcc warning
-    if (jit_dump_str.size() >= PTRDIFF_MAX) {
-      return false;
-    }
-    if (file_path.ends_with(jit_dump_str)) {
+    // The string should end with: "jit-[0-9]+\\.dump"
+    // and the number should be the pid, however, in wholehost mode
+    // we don't have visibility on the namespace's PID value.
+    // So here we just check for the presence of "jit-"
+    size_t pos = file_path.rfind("jit-");
+    if (pos != std::string::npos) {
       return true;
     }
   }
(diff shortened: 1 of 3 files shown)


src/dso.cc
```diff
@@ -83,13 +83,12 @@ Dso::Dso(pid_t pid, ElfAddress_t start, ElfAddress_t end, ElfAddress_t pgoff,
...
-    // this is to remove a gcc warning
-    if (jit_dump_str.size() >= PTRDIFF_MAX) {
-      return false;
-    }
-    if (file_path.ends_with(jit_dump_str)) {
+    // The string should end with: "jit-[0-9]+\\.dump"
+    // and the number should be the pid, however, in wholehost mode
+    // we don't have visibility on the namespace's PID value.
+    // So here we just check for the presence of "jit-"
+    size_t pos = file_path.rfind("jit-");
```
nsavoire, line 90: Why not check for the format given above ?

r1viollet: I added a more restrictive check

src/dso.cc
```diff
@@ -81,16 +81,24 @@ Dso::Dso(pid_t pid, ElfAddress_t start, ElfAddress_t end, ElfAddress_t pgoff,
...
-      return true;
+  // Test if we finish by .dump before creating a string
+  constexpr std::string_view suffix = ".dump";
+  constexpr std::string_view prefix = "jit-";
+  if (file_path.ends_with(suffix)) {
+    // The string should end with: "jit-[0-9]+\\.dump"
+    // and the number should be the pid, however, in wholehost mode
+    // we don't have visibility on the namespace's PID value.
+    // So here we just check for the presence of "jit-"
+    size_t pos = file_path.rfind(prefix);
```
nsavoire, line 92: This will incorrectly match `aaaajit-1234.dump`

src/dso.cc
```diff
@@ -80,20 +80,32 @@ Dso::Dso(pid_t pid, ElfAddress_t start, ElfAddress_t end, ElfAddress_t pgoff,
...
 
-bool Dso::is_jit_dump_str(std::string_view file_path, pid_t pid) {
-  // test if we finish by .dump before creating a string
-  if (file_path.ends_with(".dump")) {
-    // llvm uses this format
-    std::string jit_dump_str = string_format("jit-%d.dump", pid);
-    // this is to remove a gcc warning
-    if (jit_dump_str.size() >= PTRDIFF_MAX) {
+bool is_digits(std::string_view str) {
+  for (char ch : str) {
```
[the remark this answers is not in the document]
r1viollet: Done

PR #298, opened 2023-08-28
Performance live heap profiling - deallocation code path

r1viollet:
# What does this PR do?

Ensure the structure to keep the allocation addresses is lock free.
The proposal is to have a bitset (using atomic integers) instead of a hash map. There is some logic to map the address to a position in the bitset.

# Motivation

Performances:
I get significantly better numbers in the benchmark (when focusing on the allocation code paths)

```
# New numbers are in the same ball park
BM_ShortLived_NoTracking/process_time/real_time     440168 ns       998071 ns         1349
BM_ShortLived_Tracking/process_time/real_time       490663 ns      1160473 ns         1505
BM_LongLived_NoTracking/process_time                345291 ns       218756 ns         3223
BM_LongLived_Tracking/process_time                  342495 ns       358709 ns         1840

# Previous numbers (with same config)
BM_ShortLived_NoTracking/process_time/real_time     448681 ns       946333 ns         1602
BM_ShortLived_Tracking/process_time/real_time      1860407 ns      5558747 ns          377
BM_LongLived_NoTracking/process_time                339814 ns       197875 ns         3370
BM_LongLived_Tracking/process_time                 1316344 ns      4319851 ns          142
```

# Additional Notes

The bitset has a limited number of elements. This impacts the probability of capturing the allocation event.
This only occurs when live allocation is triggered.

# How to test the change?

The improvement in performances can be observed through benchmarks.
The unit test cover the new data structure.
The existing tests already test live heap profiling.

Thanks to @richardstartin for mentioning this data structure.


src/lib/address_bitset.cc
```diff
@@ -0,0 +1,60 @@
...
+}
+
+bool AddressBitset::unset(uintptr_t addr) {
+  size_t hash_addr = std::hash<uintptr_t>().operator()(addr);
+  int significant_bits = hash_addr & _nb_bits_mask;
+  auto res = std::div(significant_bits, 64);
+  unsigned index_array = res.quot;
+  unsigned bit_offset = res.rem;
+  bool success = false;
+  int attempt = 0;
```
r1viollet, line 41: A failed attempt here can have a desync consequence. Is this acceptable ?

r1viollet: I am not seeing this happen in any test case + added an assertion.

src/lib/allocation_tracker.cc
```diff
@@ -229,27 +231,34 @@ void AllocationTracker::track_allocation(uintptr_t addr, size_t size,
...
+          _dealloc_bitset.set(addr);
+        } else {
+          log_once(
+              "Error: %s",
+              "Stop allocation profiling. Unable to clear live allocation \n");
+          free();
+        }
       }
+    } else {
+      // we won't track this allocation as we can't double count in the bitset
```
r1viollet, line 253: This is something we can improve.

r1viollet: It does not happen very often and we can adjust the likeliness of this happening using the bitset size.

src/lib/allocation_tracker.cc
```diff
@@ -229,27 +231,34 @@ void AllocationTracker::track_allocation(uintptr_t addr, size_t size,
...
-            "Stop allocation profiling. Unable to clear live allocation \n");
-        free();
+  if (_state.track_deallocations) {
+    if (_dealloc_bitset.set(addr)) {
+      if (unlikely(_dealloc_bitset.nb_elements() >
+                   ddprof::liveallocation::kMaxTracked)) {
+        // Check if we reached max number of elements
+        // Clear elements if we reach too many
+        // todo: should we just stop new live tracking (like java) ?
+        if (IsDDResOK(push_clear_live_allocation(tl_state))) {
```
r1viollet, line 241: I don't want to adjust the strategy in this pr. However the current behaviour can be troubling for users.

src/lib/allocation_tracker.cc
r1viollet: A simple improvement could be to avoid tracking this within the live heap state.

src/lib/address_bitset.cc
```diff
@@ -0,0 +1,65 @@
...
+  do {
+    uint64_t old_value = _address_bitset[index_array].load();
+    if (old_value & (static_cast<uint64_t>(1) << bit_offset)) {
+      // element is already set (collisions are expected)
+      return false;
+    }
+    uint64_t new_value = old_value;
+    new_value |= static_cast<uint64_t>(1) << bit_offset;
+    success = _address_bitset[index_array].compare_exchange_weak(old_value,
```
nsavoire, line 29: Better to use `std::atomic::fetch_or` instead of CAS.

src/lib/address_bitset.cc
```diff
@@ -0,0 +1,65 @@
...
+  int attempt = 0;
+  do {
+    uint64_t old_value = _address_bitset[index_array].load();
+    if (!(old_value & (static_cast<uint64_t>(1) << bit_offset))) {
+      // element is not already set (unexpected?)
+      break;
+    }
+    uint64_t new_value = old_value;
+    new_value ^= static_cast<uint64_t>(1) << bit_offset;
+    success = _address_bitset[index_array].compare_exchange_weak(old_value,
```
nsavoire, line 54: Better to use `std::atomic::fetch_xor` instead of CAS.

include/lib/lib_logger.hpp
```diff
@@ -12,7 +12,7 @@
 namespace ddprof {
 template <typename... Args>
 void log_once(char const *const format, Args... args) {
-#ifndef DEBUG
+#ifdef NDEBUG
   static std::once_flag flag;
```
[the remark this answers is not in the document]
r1viollet: I agree :-), though I could not think of something better for now. 
I still prefer we don't log too much.

src/lib/address_bitset.cc
r1viollet: Then we have a possible race between checking the value and actually setting it.
I'll make the change and you can tell me if you are OK with that possible race condition.

src/lib/address_bitset.cc
```diff
@@ -0,0 +1,90 @@
...
+  _nb_bits = max_addresses;
+  _k_nb_elements = (_nb_bits) / (_nb_bits_per_elt);
+  if (_nb_bits) {
+    _nb_bits_mask = _nb_bits - 1;
+    _address_bitset = std::make_unique<std::atomic<uint64_t>[]>(_k_nb_elements);
+  }
+}
+
+bool AddressBitset::set(uintptr_t addr) {
+  uint64_t hash_addr = remove_lower_bits(addr);
```
r1viollet, line 43: I think using a standard hash has some merit.

r1viollet: This strategy is also interesting considering we don't expect a lot of interesting consecutive addresses for live heap.

src/lib/allocation_tracker.cc
```diff
@@ -140,6 +140,11 @@ DDRes AllocationTracker::init(uint64_t mem_profile_interval,
       static_cast<int>(RingBufferType::kMPSCRingBuffer)) {
     return ddres_error(DD_WHAT_PERFRB);
   }
+  if (track_deallocations) {
+    // 16 times as we want to probability of collision to be low enough
+    _allocated_address_set.init(_sampling_interval,
```
r1viollet, line 145: We only init with a larger size if we activate the tracking of deallocations

src/lib/allocation_tracker.cc
```diff
@@ -163,6 +162,9 @@ void AllocationTracker::allocation_tracking_free() {
   if (!instance) {
     return;
   }
+  if (instance->_state.track_deallocations) {
+    instance->_dealloc_bitset.clear();
```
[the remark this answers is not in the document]
r1viollet: I added an init step to ensure we only create the bitmap if needed.

include/lib/address_bitset.hpp
```diff
@@ -0,0 +1,61 @@
...
+  // We remove the lower bits (as the alignment constraints makes them useless)
+  // We fold the address
+  // Then we only keep the bits that matter for the order in the bitmap
+  uint32_t remove_lower_bits(uintptr_t h1) {
+    uint64_t intermediate = h1 >> _lower_bits_ignored;
+    uint32_t high = (uint32_t)(intermediate >> 32);
+    uint32_t low = (uint32_t)intermediate;
+    uint32_t res = high ^ low;
+    return res & _nb_bits_mask;
+  }
```
richardstartin, line 59: I think you could get away with exactness here:
* you can strip the top 16 bits out anyway, so you only need to actually store 48 bits, on top of the up to 4 you can remove for alignment.
* if you only need to store `48 - alignment` bits, you can consider a paginated structure, where you have up to `2^(32 - alignment)` pages of 16 bit bitsets. This means you have an enormous array of null pointers, occasionally pointing to an 8KiB bitset whenever an address in the range [i, i + 2^16) has been encountered. 
* `2^(32 - alignment)` is going to be too large (at least 2GiB with the max alignment), but allocation patterns will be sparse, especially in the high order bits, so what about a 3 level structure with up to 2^16 pages for the top 16 bits, each (there will be very few of these populated) indexing paginated bitsets of `2^(16 - alignment)` pages of 16 bit bitsets. This pushes the alignment saving into the middle 12-16 bits which are likely to have higher cardinalities than the top 16. 
   * The top level costs 512 KiB, most of it will be null pointers
   * Each populated range in the top level will cost 32-512 KiB according to alignment (32 for 16 byte alignment, 512 for no alignment)
   * Each range of 2^16 addresses after alignment costs 8KiB for the leaf level bitset.
   * E.g. for an 8GiB address space and 8 byte alignment, you end up with a fixed 512KiB, 2 * 64 KiB for the 2 populated middle pages, and 8KiB for any range of 2^16 consecutive addresses where you sample a pointer.

richardstartin: And you can trim the 512KiB a lot if you know the max address

src/lib/address_bitset.cc
```diff
@@ -0,0 +1,75 @@
...
+  uint64_t bit_in_element = (1UL << bit_offset);
+  uint64_t old_value = _address_bitset[index_array].load();
+  if (old_value & bit_in_element) {
+    // element is already set (collisions are expected)
+    return false;
+  }
+  // there is a possible race between checking the value
+  // and setting it
+  _address_bitset[index_array].fetch_or(bit_in_element);
+  ++_nb_addresses;
```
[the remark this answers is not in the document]
r1viollet: I actually can do the same with the return value of the fetch or. thanks.

include/lib/address_bitset.hpp
r1viollet: Thanks for exposing the math here.

Exactness can have value, especially for cases where the user puts a lower sampling interval and just wants to debug his program. I will delay trying this out as it will take me down a new cycles of experiments and measures.
On average the current approach is good to reduce memory impact. We do skew slightly the sampling intervals when a collision occurs.

r1viollet:
I checked that CI is fine (a deploy job could not start https://gitlab.ddbuild.io/DataDog/apm-reliability/ddprof-build/-/jobs/325597595)

src/lib/address_bitset.cc
```diff
@@ -0,0 +1,70 @@
...
+    ++_nb_addresses;
+    return true;
+  }
+  // Collision, element was already set
+  return false;
+}
+
+bool AddressBitset::unset(uintptr_t addr) {
+  uint64_t hash_addr = remove_lower_bits(addr);
+  int significant_bits = hash_addr & _nb_bits_mask;
```
nsavoire, line 50: masking is already done in `remove_lower_bits` (contrary to what the name suggests ;) )

src/lib/address_bitset.cc
```diff
@@ -0,0 +1,70 @@
...
+  unsigned index_array = significant_bits / 64;
+  unsigned bit_offset = significant_bits % 64;
+  uint64_t bit_in_element = (1UL << bit_offset);
+  if ((_address_bitset[index_array].fetch_xor(bit_in_element) &
+       bit_in_element) &&
+      likely(_nb_addresses.load(std::memory_order_relaxed) >= 0)) {
```
nsavoire, line 56: `_nb_addresses.load` followed by `--nb_addresses` seems wrong

src/lib/allocation_tracker.cc
```diff
@@ -229,27 +233,34 @@ void AllocationTracker::track_allocation(uintptr_t addr, size_t size,
...
-        free();
+  if (_state.track_deallocations) {
+    if (_allocated_address_set.set(addr)) {
+      if (unlikely(_allocated_address_set.nb_addresses() >
+                   ddprof::liveallocation::kMaxTracked)) {
+        // Check if we reached max number of elements
+        // Clear elements if we reach too many
+        // todo: should we just stop new live tracking (like java) ?
+        if (IsDDResOK(push_clear_live_allocation(tl_state))) {
+          _allocated_address_set.clear();
```
nsavoire, line 244: `clear` is not thread safe

include/lib/address_bitset.hpp
```diff
@@ -0,0 +1,61 @@
...
+  // returns true if the element was inserted
+  bool set(uintptr_t addr);
+  // returns true if the element was removed
+  bool unset(uintptr_t addr);
+  void clear();
+  int nb_addresses() const { return _nb_addresses; }
+
+private:
+  static constexpr unsigned _k_max_bits_ignored = 4;
+  unsigned _lower_bits_ignored;
```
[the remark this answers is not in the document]
r1viollet: I suspect this might change and be a function of alignment.

src/lib/allocation_tracker.cc
```diff
@@ -140,6 +140,10 @@ DDRes AllocationTracker::init(uint64_t mem_profile_interval,
       static_cast<int>(RingBufferType::kMPSCRingBuffer)) {
     return ddres_error(DD_WHAT_PERFRB);
   }
+  if (track_deallocations) {
+    // 16 times as we want to probability of collision to be low enough
+    _allocated_address_set.init(liveallocation::kMaxTracked * 16);
```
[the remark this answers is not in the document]
r1viollet: Just to make sure we are on same page: the probability of occurring is `k(k+1)/2n` (at startup, but k^2/n is a good generalization when we are running), though probability of individual events is still `k/n`.
What are we optimizing for ? I don't think we want this never to happen.
I would be more interested in keeping memory impact as low as possible for now. The impact of skipping an allocation for live is OK I feel. Skipping it for allocation profiling is not ideal, though we can fix this.

src/lib/address_bitset.cc
```diff
@@ -0,0 +1,70 @@
...
+  // Due to memory alignment, on 64 bits we can assume that the first 4
+  // bits can be ignored
+  _lower_bits_ignored = _k_max_bits_ignored;
+  if (_address_bitset) {
+    _address_bitset.reset();
+  }
+  _nb_bits = max_addresses;
+  _k_nb_elements = (_nb_bits) / (_nb_bits_per_elt);
+  if (_nb_bits) {
+    _nb_bits_mask = _nb_bits - 1;
```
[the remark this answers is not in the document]
r1viollet: Done, good idea.

src/lib/allocation_tracker.cc
r1viollet: Can you clarify what you expect the clear function to look like ?

include/lib/address_bitset.hpp
```diff
@@ -0,0 +1,61 @@
...
+  // so lets say you have 1111, you can store 4 addresses.
+  // We hash the address to a number (to have an equal probability of using
+  // all bits). Then we use the mask to position this address in our bitset.
+  // Addr -> Hash -> Mask (to get useable bits) -> Position in the bitset
+  // Note: the hashing step might be bad for cache locality.
+public:
+  // Publish 1 Meg as default
+  constexpr static unsigned _k_default_max_addresses = 8 * 1024 * 1024;
+  AddressBitset(unsigned max_addresses = 0) { init(max_addresses); }
+  void init(unsigned max_addresses);
```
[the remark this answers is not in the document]
r1viollet: I guess the intent to make the object immutable and to just exchange with a different object when we want to adjust the values ?

include/lib/lib_logger.hpp
```diff
@@ -10,13 +10,24 @@
...
-  fprintf(stderr, format, args...);
+template <typename Func> void log_once_helper(std::once_flag &, Func &&func) {
+  func();
 #endif
 }
+
+// create a once flag for the line and file where this is called:
+#define LOG_ONCE(format, ...)                                                  \
+  do {                                                                         \
+    static std::once_flag UNIQUE_ONCE_FLAG_##__LINE__##__FILE__;               \
```
r1viollet, line 27: @nsavoire I am pushing the once flag to the place of the caller, if this is OK.

src/lib/address_bitset.cc
```diff
@@ -0,0 +1,102 @@
...
+  return false;
+}
+
+unsigned int AddressBitset::count_set_bits(Word_t w) {
+  unsigned int set_bits = 0;
+  while (w) {
+    set_bits += w & 1;
+    w >>= 1;
+  }
+  return set_bits;
```
richardstartin, line 88: You can use `__builtin_popcount` to do all this in a single instruction

src/lib/address_bitset.cc
```diff
@@ -0,0 +1,102 @@
...
+  if ((num & (num - 1)) == 0) {
+    return num;
+  }
+  // not a power of two
+  unsigned count = 0;
+  while (num) {
+    num >>= 1;
+    count++;
+  }
+  return 1 << count;
```
richardstartin, line 28: This can be replaced by 
```
1 << (32 - _builtin_clz(x - 1));
```

src/lib/address_bitset.cc
richardstartin: (hopefully the compiler knows this, but it might not.)

include/lib/address_bitset.hpp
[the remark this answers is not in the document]
r1viollet: Clarified: the pattern is cleaner.

include/lib/address_bitset.hpp
```diff
@@ -0,0 +1,63 @@
...
+  // We have one address per individual bit).
+  // so lets say you have 1111, you can store 4 addresses.
+  // We hash the address to a number (to have an equal probability of using
+  // all bits). Then we use the mask to position this address in our bitset.
+  // Addr -> Hash -> Mask (to get useable bits) -> Position in the bitset
+  // Note: the hashing step might be bad for cache locality.
+public:
+  // Publish 1 Meg as default
+  constexpr static unsigned _k_default_bitset_size = 8 * 1024 * 1024;
+  explicit AddressBitset(unsigned max_addresses = 0) { init(max_addresses); }
```
r1viollet, line 24: todo: naming update (max_addresses is confusing)

include/lib/lib_logger.hpp
r1viollet: Counter is the more standard way of giving a unique naming

src/lib/address_bitset.cc
r1viollet: Thanks I had looked at this, though missed it was accessible in c++ 20.

src/lib/allocation_tracker.cc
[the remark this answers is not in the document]
r1viollet: Talking more about this, the aim would be to have a stable probability based on maxed tracked.
That means the size would be a function of k^2 (which blows up really fast).
I won't spend too much time on this as I think it might be better to do what @richardstartin was suggesting (make it work for all addresses).

src/lib/address_bitset.cc
r1viollet: Thanks, good to know. I won't use the builtin as it is an init flow (so it does not really need the performance).

src/lib/allocation_tracker.cc
r1viollet: As discussed the count does not have to be 100% accurate.

include/lib/address_bitset.hpp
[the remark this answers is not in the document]
r1viollet: Done

include/lib/address_bitset.hpp
r1viollet: Added a ticket to internal board for now: https://datadoghq.atlassian.net/browse/PROF-8222

src/lib/address_bitset.cc
```diff
@@ -0,0 +1,70 @@
...
+    _address_bitset = std::make_unique<std::atomic<uint64_t>[]>(_k_nb_elements);
+  }
+}
+
+bool AddressBitset::set(uintptr_t addr) {
+  uint32_t significant_bits = remove_lower_bits(addr);
+  // As per nsavoire's comment, it is better to use separate operators
+  // than to use the div instruction which generates an extra function call
+  // Also, the usage of a power of two value allows for bit operations
+  unsigned index_array = significant_bits / 64;
```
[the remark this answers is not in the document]
r1viollet: This was not equivalent ;-), though no consequences.

include/lib/address_bitset.hpp
r1viollet: A first draft around this topic: https://github.com/DataDog/ddprof/pull/307/files
Perhaps we need to have some quick access path. It looks like the current performances are not yet good enough.

include/lib/allocation_tracker.hpp
```diff
@@ -116,7 +118,8 @@ class AllocationTracker {
   uint32_t _stack_sample_size;
   PEvent _pevent;
   bool _deterministic_sampling;
-  AdressSet _address_set;
+
+  AddressBitset _allocated_address_set;
```
r1viollet, line 122: minor: naming can be adjusted

r1viollet: nvm, this is good naming

nsavoire (approved):
👍

PR #299, opened 2023-08-30
Improve ELF segment matching

nsavoire:
# What does this PR do?
To compute elf bias, we need to match a Dso (ie. a mapping in process address space) with the coresponding LOAD segment in elf file.
Before this change, we would use the Dso containing the address we want to unwind, and make the assumption that an elf file contains only one LOAD executable segment, and therefore that the matching segment would be the unique executable LOAD segment of the elf file.
This assumption turns out to be wrong: an elf file can contain multiple LOAD executable segments.
With this change, we use all the executable mappings of the elf file and try to match them with LOAD segments by using the aligned offset and the permissions of the mappings.
We consider only executable mappings LOAD segments (because we configure perf in such a way that we receive only executable mmap events).
The matching is done by comparing mapping offset with aligned file offset (file offset may not be aligned, in particular with lld, but when segment is mmapped, mmap offset needs to be page aligned, file offset is rounded down to satisfy segment p_align alignment) and memory protection of segment.
Several segments can have the same aligned offsets and protections though, thus making the matching ambiguous.
Any mapping can be used for matching, but some mappings might be missing because of lost perf events. We start with the last mapping in address order, because it is the one that has the most constraints: if we observed n executable mappings, then we know that the last mapping is preceded by at least n-1 other exec mappings and therefore cannot match the first n-1 LOAD exec segments. If last mapping leads to an ambiguous match, we try again with the second to last mapping and so on...

## Other changes:
* add an implementation of `std::unique_resource`/`std::scope_exit` (only available with gcc 13 currently) from [here](https://github.com/PeterSommerlad/SC22WG21_Papers/tree/master/workspace/P0052_scope_exit)
* reimplement defer on top of `scope_exit`

### DIFF
--- a/include/ddprof_module_lib.hpp
+++ b/include/ddprof_module_lib.hpp
@@ -9,12 +9,53 @@
 #include "ddprof_module.hpp"
 #include "ddres_def.hpp"
 #include "dso.hpp"
+#include "dso_hdr.hpp"
 #include "dwfl_internals.hpp"
 
 namespace ddprof {
+
+// Convert elf flags to mmap prot flags
+inline uint32_t elf_flags_to_prot(Elf64_Word flags) {
+  return ((flags & PF_X) ? PROT_EXEC : 0) | ((flags & PF_R) ? PROT_READ : 0) |
+      ((flags & PF_W) ? PROT_WRITE : 0);
+}
+
+// Structure to represent both a mapping and a loadable elf segment
+struct Segment {
+  ElfAddress_t addr;
+  Offset_t offset;
+  uint64_t alignment;
+  uint32_t prot;
+};
+
+struct Mapping {
+  ElfAddress_t addr;
+  Offset_t offset;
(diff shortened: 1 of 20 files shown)


r1viollet (dismissed):
Thanks for looking deeper into this.
LGTM

r1viollet:
Where should we put the test case you created ?

nsavoire:
> Where should we put the test case you created ?

I thought about this, but I did not find yet a good solution:
* should I add the binaries directly to the repository ?
* should I build them (but some cases require lld) ?
* how can I simulate some missing perf events ?

r1viollet:
> > Where should we put the test case you created ?
> ...
The build steps have a lot of value to tune other test cases. If you can commit them to ddprof that would be nice.

r1viollet (approved):
LGTM

r1viollet (approved):
LGTM

PR #302, opened 2023-09-11
Pprof name - live heap

r1viollet:
# What does this PR do?

Manage several types of aggregation methods per event types.

# Motivation

Ensure the correct pprof name is generated with live mode.
Ensure other types of events can have different output methods.

### DIFF
--- a/include/perf_watcher.hpp
+++ b/include/perf_watcher.hpp
@@ -28,11 +28,11 @@ struct PerfWatcherOptions {
       k_default_perf_stack_sample_size}; // size of the user stack to capture
 };
 
-enum PprofPos_t {
-  kCallgraph = 0,
-  kLive,
-  kNbPprofPos,
-};
+// enum PprofPos_t {
+//   kCallgraph = 0,
+//   kLive,
+//   kNbPprofPos,
+// };
 
 struct PProfIndices {
   int pprof_index = -1;
@@ -66,10 +66,9 @@ struct PerfWatcher {
   // Other configs
   bool suppress_pid;
   bool suppress_tid;
-  PProfIndices pprof_indices[kNbPprofPos]; // std and live
-  bool instrument_self;                    // do my own perfopen, etc
-  // Defalut output is callgraph
-  EventConfMode output_mode = EventConfMode::kCallgraph;
+  PProfIndices pprof_indices[kNbEventValueModes]; // std and live
+  bool instrument_self;                           // do my own perfopen, etc
+  EventValueMode output_mode{EventValueMode::kOccurence};
 };
 
 // The Datadog backend only understands pre-configured event types.  Those
--- a/include/perf_watcher.hpp
+++ b/include/perf_watcher.hpp
@@ -28,12 +28,6 @@ struct PerfWatcherOptions {
       k_default_perf_stack_sample_size}; // size of the user stack to capture
 };
 
-// enum PprofPos_t {
-//   kCallgraph = 0,
-//   kLive,
-//   kNbPprofPos,
-// };
-
 struct PProfIndices {
   int pprof_index = -1;
   int pprof_count_index = -1;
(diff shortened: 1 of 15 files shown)


test/simple_malloc-ut.sh
```diff
@@ -49,10 +49,10 @@ check() {
     fi
     if [[ "${expected_pids}" -ne 0 ]]; then
 
-        counted_pids_alloc=$(count "${log_file}" "alloc-samples" "pid")
```
r1viollet, line 52: I'm not sure why we were putting the countable name in the sample in this test.

src/ddprof_worker.cc
```diff
@@ -245,10 +246,8 @@ static DDRes ddprof_unwind_sample(DDProfContext &ctx, perf_event_sample *sample,
   if (watcher->config == PERF_COUNT_SW_TASK_CLOCK)
     ddprof_stats_add(STATS_TARGET_CPU_USAGE, sample->period, NULL);
 
-  // Attempt to fully unwind if the watcher has a callgraph type
-  DDRes res = {};
-  if (AnyCallgraph(watcher->output_mode))
-    res = unwindstate__unwind(us);
+  // Attempt to fully unwind for this watcher
+  DDRes res = unwindstate__unwind(us);
```
r1viollet, line 250: Metrics are not yet a thing and need more work. I'm just removing the branch for now.

src/event_parser/event_parser.y
```diff
@@ -228,7 +223,8 @@ opt:
                delete $3;
                VAL_ERROR();
              }
-             g_accum_event_conf.mode |= *mode;
+             // override mode if present
+             g_accum_event_conf.mode = *mode;
```
r1viollet, line 227: I don't think we should allow several modes (though we don't check that afaik...)

src/ddprof_worker.cc
r1viollet: Metrics belong in a different field.
They don't change the value computation, they change the way we report the data.

include/perf_watcher.hpp
```diff
@@ -55,25 +60,27 @@ struct PerfWatcher {
   // Other configs
   bool suppress_pid;
   bool suppress_tid;
-  int pprof_sample_idx;       // index into the SampleType in the pprof
-  int pprof_count_sample_idx; // index into the pprof for the count
-  bool instrument_self;       // do my own perfopen, etc
-  EventConfMode output_mode;  // defines how sample data is aggregated
+  PProfIndices pprof_indices[kNbEventValueModes]; // std and live
+  bool instrument_self;                           // do my own perfopen, etc
+  EventValueMode output_mode{EventValueMode::kOccurence};
```
r1viollet, line 65: this is a value_mode (not output_mode)

include/perf_watcher.hpp
```diff
@@ -55,25 +60,27 @@ struct PerfWatcher {
...
 };
 
 // The Datadog backend only understands pre-configured event types.  Those
 // types are defined here, and then referenced in the watcher
 // The last column is a dependent type which is always aggregated as a count
 // whenever the main type is aggregated.
+//  type,    pprof,     unit, live-pprof,  sample_type,
```
r1viollet, line 72: naming: sample-type (not pprof)

src/ddprof_worker.cc
```diff
@@ -306,23 +304,20 @@ DDRes ddprof_pr_sample(DDProfContext &ctx, perf_event_sample *sample,
...
     struct UnwindState *us = ctx.worker_ctx.us;
-    if (Any(EventConfMode::kLiveCallgraph & watcher->output_mode)) {
-      // Live callgraph mode
-      // for now we hard code the live aggregation mode
+    if (Any(EventValueMode::kLiveUsage & watcher->output_mode)) {
       ctx.worker_ctx.live_allocation.register_allocation(
           us->output, sample->addr, sample->period, watcher_pos, sample->pid);
-    } else if (Any(EventConfMode::kCallgraph & watcher->output_mode)) {
+    }
+    if (Any(EventValueMode::kOccurrence & watcher->output_mode)) {
```
r1viollet, line 311: notice that this is no longer an else condition

r1viollet:
Closing in favour of : https://github.com/DataDog/ddprof/pull/316

PR #303, opened 2023-09-13
Replace tcb/span with std::span

nsavoire:
# What does this PR do?

Replace tcb/span with std::span.

# Motivation

Remove dependency on tcb/span.


r1viollet (approved):
LGTM
Thanks for the cleanup!

PR #306, opened 2023-09-14
Fix loaded libs check

nsavoire:
# What does this PR do?

* Make loaded libs check initial delay and interval configurable
* Add simple malloc test for loaded libs check
* Reinstall loaded libs check timer after fork


r1viollet (approved):
LGTM

PR #308, opened 2023-09-19
Add support for dynamically linked jemalloc

nsavoire:
* Override jemalloc non-standard API (mallocx,rallocx,xallocx,...)
* Override all variants of operator new/delete:
This was not necessary when using glibc malloc because implementation of these operators in libstdc++ are calling malloc/free. But since jemalloc provides its own versions of these operators that call directly into some jemalloc internal functions, why we need to intercept these allocations at the operator new/delete level or otherwise we will miss all c++ allocations.
* Override sized free functions added by C23 (free_sized/free_aligned_sized)
* Add tests for all (de)allocation functions

### DIFF
--- a/cmake/ExtendBuildTypes.cmake
+++ b/cmake/ExtendBuildTypes.cmake
@@ -19,7 +19,7 @@ if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
     add_compile_options(-Wshadow=local)
   endif()
 else()
-  add_compile_options(-Wshadow -Wno-c99-designator)
+  add_compile_options(-Wshadow -Wno-c99-designator -fsized-deallocation)
 endif()
 
 set(SAN_FLAGS "-fsanitize=undefined -fsanitize=float-divide-by-zero -fno-sanitize-recover")
@@ -42,12 +42,6 @@ set(CMAKE_CXX_FLAGS_SANITIZEDDEBUG
 set(CMAKE_C_FLAGS_SANITIZEDDEBUG
     "${SAN_FLAGS} ${ASAN_FLAGS} ${STACK_FLAGS}"
     CACHE STRING "Flags used by the C compiler during sanitized builds." FORCE)
-set(CMAKE_EXE_LINKER_FLAGS_SANITIZEDDEBUG
-    ""
-    CACHE STRING "Flags used for linking binaries during sanitized builds." FORCE)
-set(CMAKE_SHARED_LINKER_FLAGS_SANITIZEDDEBUG
-    ""
-    CACHE STRING "Flags used by the shared libraries linker during sanitized builds." FORCE)
 
 # Add flags for thread-sanized debu
 set(CMAKE_CXX_FLAGS_THREADSANITIZEDDEBUG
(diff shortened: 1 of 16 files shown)


cmake/ExtendBuildTypes.cmake
```diff
@@ -19,7 +19,7 @@ if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
     add_compile_options(-Wshadow=local)
   endif()
 else()
-  add_compile_options(-Wshadow -Wno-c99-designator)
+  add_compile_options(-Wshadow -Wno-c99-designator -fsized-deallocation)
```
r1viollet, line 22: I can't remember what the defaults are across compilers. Thanks for adding this in. I guess we could also just add it for all versions / compilers ?

cmake/Jemalloc.cmake
```diff
@@ -45,4 +51,17 @@ mark_as_advanced(
     JEMALLOC_LIBRARIES
     JEMALLOC_INCLUDE_DIR
 )
+if(JeMalloc_FOUND AND NOT (TARGET JeMalloc::JeMalloc))
```
r1viollet, line 54: minor: maybe a comment to explain the places where this is not the case ?

include/lib/reentry_guard.hpp
```diff
@@ -59,7 +59,6 @@ class TLReentryGuard {
 
 private:
   ThreadEntries &_entries;
-  pid_t _tid;
```
r1viollet, line 62: This was indeed unused.

src/lib/symbol_overrides.cc
```diff
@@ -31,6 +31,19 @@ __attribute__((weak)) void *reallocarray(void *ptr, size_t nmemb,
 __attribute__((weak)) void *pvalloc(size_t size) NOEXCEPT;
 // NOLINTNEXTLINE cert-dcl51-cpp
 __attribute__((weak)) int __libc_allocate_rtsig(int high) NOEXCEPT;
+
+// sized free functions (C23, not yet available in glibc)
```
r1viollet, line 35: New ones ! Thanks for keeping an eye out for these. How do we find / refresh this list of APIs ?

src/lib/symbol_overrides.cc
```diff
@@ -93,33 +252,259 @@ struct free {
...
     ref(ptr);
   }
 };
 
+struct free_sized {
+  static constexpr auto name = "free_sized";
+  static inline auto ref = &::free_sized;
+  static inline bool ref_checked = false;
+
+  static void hook(void *ptr, size_t size) noexcept {
```
r1viollet, line 270: Is the size given always the same as the size from the allocation ?

r1viollet: This is not relevant to the PR, I'm mostly worried about live heap's accuracy.

src/lib/symbol_overrides.cc
```diff
@@ -67,22 +80,168 @@ void check_libraries() {
...
   static inline auto ref = &::malloc;
   static inline bool ref_checked = false;
 
   static void *hook(size_t size) noexcept {
-    check_libraries();
-    ddprof::TrackerThreadLocalState *tl_state =
-        ddprof::AllocationTracker::get_tl_state();
-    ddprof::ReentryGuard guard(tl_state ? &(tl_state->double_tracking_guard)
-                                        : nullptr);
+    AllocTrackerHelper helper;
```
r1viollet, line 125: so the assumption is that we always keep track of the first function that is tracked (as that is the one that keeps the guard)
This is good for 
```
malloc -> mmap
free -> munmap
```
Are we worried about things that would happen in a reverse order ?

cmake/ExtendBuildTypes.cmake
nsavoire: this is has been the default for gcc for a long time, only necessary for clang.

cmake/Jemalloc.cmake
nsavoire: Just copied it from https://fossies.org/linux/ceph/cmake/modules/FindJeMalloc.cmake
Not sure that the `if(NOT TARGET` is necessary.

src/lib/symbol_overrides.cc
nsavoire: it's the responsibility of the caller to give the same size as the one passed to the allocation function.

src/lib/symbol_overrides.cc
nsavoire: Are you worried about `mmap` calling `malloc` ? I don't see how that could happen (mmap is a wrapper around a syscall).

r1viollet (approved):
LGTM
Thanks for figuring out where the story and coming up with better testing!

cmake/Jemalloc.cmake
nsavoire: I removed the second part of the if.

r1viollet (approved):
LGTM

PR #310, opened 2023-09-22
Allow to update CI status

nsavoire:
# What does this PR do?

When downstream pipeline failed, there was not way to update parent CI status.
Now when downstream pipeline is retried and succeeds report_success job is run and updates EXIT_STATUS. Then rerunning report_gitlab_CI_status allows to update parent CI status.


r1viollet (approved):
LGTM

PR #314, opened 2023-09-28
Crash fix - Fix segment matching

r1viollet:
# What does this PR do?

Ensure we consider the case where there are more loaded segments than elf segments

# Motivation

Fix a crash when running in whole host

# Additional Notes

This was introduced by recent changes to the way we match segments

# How to test the change?

This library was loaded by appgate on linux.

### DIFF
--- a/src/ddprof_module_lib.cc
+++ b/src/ddprof_module_lib.cc
@@ -10,6 +10,7 @@
 #include "defer.hpp"
 #include "logger.hpp"
 
+#include <algorithm>
 #include <cstring>
 #include <fcntl.h>
 #include <gelf.h>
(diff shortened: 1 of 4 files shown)


src/ddprof_module_lib.cc
```diff
@@ -104,14 +105,16 @@ find_matching_segment(std::span<const Segment> segments,
 
 MatchResult find_match(std::span<const Mapping> executable_mappings,
                        std::span<const Segment> elf_load_segments) {
-  int nb_mappings = executable_mappings.size();
+  unsigned nb_mappings = executable_mappings.size();
```
nsavoire, line 108: you should stick to int

nsavoire, line 108:
```suggestion
  int nb_mappings = std::min(executable_mappings.size(), elf_load_segments.size());
```

r1viollet: Done, though I am not sure I see the point considering sizes are unsigned.

nsavoire (approved):

PR #315, opened 2023-09-28
Duplicate mappings upon fork

nsavoire:
# What does this PR do?

Upon fork duplicate parent pid mappings for child pid.


r1viollet (approved):
LGTM
I think this was our original strategy, though we encountered some issues along the way.
Since then we added logics to detect inconsistencies which should make this strategy safe.
Let's see how this behaves.

r1viollet:
Some tests on relenv should be responsive to such changes, please consider following up on metrics.

PR #318, opened 2023-10-04
Fix elf segment matching (again)

nsavoire:
# What does this PR do?

*  Use first LOAD segment for bias computation if available
    If all mappings are available (ie. they come from proc maps), then  use the first LOAD segment (executable or not) to compute the bias.
    This will ensure that bias computation is possible even for processes that punch holes into existing mappings  (eg.dotnet).

Other changes:
*  Fix typos and improve some naming
*  Use an enum class for DsoType and remove dso namespace
* Add DsoOrigin to Dso class
    DsoOrigin indicates if Dso information was obtained from a perf mmap event or from parsing proc maps.
*Only adjust Dso end if smaller
* Improve mapping insertion
    When a new mapping is inserted, if it intersects other mappings, keep  the remaining parts of the existing mappings if mapped file is the same or if new mapping is anonymous and it truncates the end of an existing  mapping.
* Avoid throwing exception if parsing of a proc map line fails
    Also explicits why expected return value of sscanf does not match number of specifiers.
* Rename dso_from_procline to dso_from_proc_line
* Improve const-correctness of dso_from_proc_line


test/dso-ut.cc
```diff
@@ -458,4 +460,124 @@ TEST(DSOTest, large_backpopulate) {
...
+  DsoHdr dso_hdr;
+  Dso dso1{5, 0x1000, 0x4fff, 0, "libfoo.so.1", 0, PROT_READ};
+  // map whole file
+  dso_hdr.insert_erase_overlap(Dso{dso1});
+
+  // map second segment
+  Dso dso2{5, 0x2000, 0x4fff, 0x1000, "libfoo.so.1", 0, PROT_READ | PROT_EXEC};
+  dso_hdr.insert_erase_overlap(Dso{dso2});
+
+  ASSERT_EQ(dso_hdr.get_nb_dso(), 2);
```
r1viollet, line 474: thanks for this test.

r1viollet (approved):
LGTM
Thanks for thoroughly looking into tricky load behaviours

PR #319, opened 2023-10-05
Log TSC calibration method

nsavoire:
# What does this PR do?

Add a log message (notice level) with the calibration method used for TSC.

# Motivation

I was surprised to see that clock_monotonic_raw was used on DD workspaces instead of perf.


r1viollet (approved):
LGTM, thanks for adding a way to check this.
Surprising that it is not always perf

PR #321, opened 2023-10-05
Simplify DSO intersection computation

nsavoire:
# What does this PR do?

Simplify DSO intersection computation using std::map upper_bound / lower_bound.


r1viollet (approved):
LGTM, Thanks
I had seen this commit at one point and was actually curious why we had not pushed it earlier ;-)

PR #322, opened 2023-10-05
Add backpopulate benchmark

nsavoire:
# What does this PR do?

* Add benchmark for backpopulate.
* Remove useless std::string temporaries in `Dso` constructor.


r1viollet (approved):
LGTM
Good catch
I will think about logging these benchmarks in CI to avoid reintroducing string creations

PR #323, opened 2023-10-06
Ringbuffer function - Minor refactor

r1viollet:
# What does this PR do?

Make the pointer arithmetic easier to understand.

# Motivation

Static analysis tools were flagging this as an error. By making it more explicit, it should be easier to avoid a false positive.
Reference to the report: https://github.com/DataDog/ddprof/security/code-scanning/210
Although this is a false positive, we can make the check more explicit.

# Additional Notes

This should not change anything.

# How to test the change?

This is already tested


nsavoire:
Commit message could be improved.
This is probably useless: cf. [previous change](https://github.com/DataDog/ddprof/commit/27c66038d20f2fdb992d53bf23b4b753dd4c72d1)

r1viollet:
OK, I dismissed the alert.

PR #324, opened 2023-10-09
Reduce the flakyiess of simple_malloc

r1viollet:
# What does this PR do?

Change niceness level of the simple malloc process
This ensure that ddprof is scheduled with a higher priority.

# Motivation

Less flaky tests.

### DIFF
--- a/test/simple_malloc.cc
+++ b/test/simple_malloc.cc
@@ -17,6 +17,7 @@
 #include <sstream>
 #include <thread>
 #include <unistd.h>
+#include <sys/resource.h>
 
 #include "ddprof_base.hpp"
 #include "enum_flags.hpp"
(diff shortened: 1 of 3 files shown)


test/simple_malloc.cc
```diff
@@ -327,6 +329,16 @@ int main(int argc, char *argv[]) {
     app.add_option("--initial-delay", opts.initial_delay, "Initial delay (ms)")
         ->default_val(0)
         ->check(CLI::NonNegativeNumber);
+    app.add_option("--nice", opts.nice, "Linux niceness setting")
+        ->default_val(19)
```
r1viollet, line 333: Proposal is to set the default to be as nice as possible this way ddprof is scheduled in priority.

nsavoire: I think we should not change the default niceness for simple_malloc, since it's not an "expected" behaviour. I would rather:
* explicitly change its niceness in simple_malloc-ut.sh
* or decrease niceness of ddprof in simple_malloc-ut.sh, which should have a similar effect and not require changing simple_malloc

r1viollet: As ddprof is a wrapper, I don't see how you can do that from the shell part of things.

nsavoire: I was proposing to either:
* by default don't change the niceness of simple_malloc and call `simple_malloc` in simple_malloc-ut.sh explicitly with `--nice 19` option (add it to `${opts}`)
*  instead of increasing the niceness of simple_malloc, decrease the niceness of ddprof with `--nice` option.

nsavoire (approved):

nsavoire (approved):

PR #327, opened 2023-10-11
Enable clang-tidy on headers

nsavoire:
# What does this PR do?

Previously clang-tidy was inadvertently disabled on headers.

I tuned down misc-non-private-member-variables-in-classes check by allowing classes with all members public because it was too much work to fix, but I think it would good to folow the guidelines of this warning and only have:
* pure data struct with public members but no methods (except perhaps a constructor)
* class objects with methods and only private members


include/build_id.hpp
```diff
@@ -2,6 +2,7 @@
 // under the Apache License Version 2.0. This product includes software
 // developed at Datadog (https://www.datadoghq.com/). Copyright 2021-Present
 // Datadog, Inc.
+#pragma once
```
r1viollet, line 5: interesting bug caught

include/ddprof_context.hpp
```diff
@@ -43,12 +40,5 @@ struct DDProfContext {
   std::vector<PerfWatcher> watchers;
   ExporterInput exp_input;
   DDProfWorkerContext worker_ctx;
-
-private:
```
r1viollet, line 47: why remove the deletes ?

include/ipc.hpp
```diff
@@ -26,16 +26,9 @@ static constexpr auto kDefaultSocketTimeout = std::chrono::seconds{2};
...
 public:
   using socket_t = int;
-  UnixSocket() noexcept = default;
 
   explicit UnixSocket(socket_t handle) noexcept : _handle(handle) {}
 
-  UnixSocket(UnixSocket &&socket) noexcept = default;
-  UnixSocket(const UnixSocket &) = delete;
-
-  UnixSocket &operator=(const UnixSocket &) = delete;
```
r1viollet, line 36: Same question here: these had a purpose ?

include/lib/allocation_tracker.hpp
```diff
@@ -63,8 +63,9 @@ class AllocationTracker {
   static TrackerThreadLocalState *get_tl_state();
 
 private:
-  using AdressSet = std::unordered_set<uintptr_t>;
+  using AddressSet = std::unordered_set<uintptr_t>;
```
r1viollet, line 66: apologies for the typo, this is sadly going to conflict with #298 where I redefined it as `AddressBitset _allocated_address_set;`

include/ddprof_context.hpp
nsavoire: Several reasons:
- they should not be private, this was the old way of preventing copy
- if there is no user defined destructor, then there is no need for them:
  - either the copy is allowed by members, so why prevent it ?
  - either some member is not copiable, and therefore already makes the class non-copyable

include/logger.hpp
```diff
@@ -63,33 +64,36 @@ typedef enum LOG_FACILITY {
...
-} LOGF_FACILITY;
+};
 
 // Allow for compile-time argument type checking for printf-like functions
 #define printflike(x, y) __attribute__((format(printf, x, y)))
 
 // Manage the logging backend
 bool LOG_syslog_open();
 void LOG_close();
-bool LOG_open(int, const char *);
```
r1viollet, line 74: Is this a lint or a general convention we should follow ?

include/ipc.hpp
nsavoire: Same as above, no destructor -> no need to delete copy constructors.
Here UniqueFd member already prevents copy.

include/logger.hpp
nsavoire: https://clang.llvm.org/extra/clang-tidy/checks/readability/named-parameter.html
Having named parameters (with names consistent between declaration and definition) helps readability.

r1viollet (dismissed):
LGTM
Thanks for adjusting all the files to these new guidelines.
At least one bug found in the process :+1:

r1viollet (approved):
LGTM

PR #328, opened 2023-10-11
Fix lost sample reporting

nsavoire:
# What does this PR do?

Take into account period or frequency when reporting lost samples.

### DIFF
--- a/src/ddprof_worker.cc
+++ b/src/ddprof_worker.cc
@@ -67,9 +67,15 @@ DDRes report_lost_events(DDProfContext &ctx) {
       us->output.clear();
       add_common_frame(us, SymbolErrors::lost_event);
 
-      auto value = watcher->sample_period * nb_lost;
-      LG_WRN("Reporting #%lu -> [%lu] lost samples for watcher #%d", nb_lost,
-             value, watcher_idx);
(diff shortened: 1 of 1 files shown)


src/ddprof_worker.cc
```diff
@@ -67,9 +67,15 @@ DDRes report_lost_events(DDProfContext &ctx) {
...
-      auto value = watcher->sample_period * nb_lost;
-      LG_WRN("Reporting #%lu -> [%lu] lost samples for watcher #%d", nb_lost,
-             value, watcher_idx);
+      auto period = (watcher->options.is_freq)
+          ? std::chrono::nanoseconds(std::chrono::seconds{1}).count() /
+              watcher->sample_frequency
+          : watcher->sample_period;
+
+      auto value = period * nb_lost;
+      LG_WRN("Reporting %lu lost samples (cumulated lost value: %lu) for "
```
r1viollet, line 76: that makes more sense ^^

r1viollet (approved):
LGTM :+1:

src/ddprof_worker.cc
nsavoire: yeah, I was always confused by the output

PR #331, opened 2023-10-19
Remove unreachable return

sanchda:
Remove unreachable return.

# What does this PR do?

Removes an unreachable return.

# Motivation

Returns should be reachable.  A return which is unreachable cannot be reached, so it can be confusing when people try to reach it.

# Additional Notes

When a return is unreachable, nobody can reach it.

# How to test the change?

There is NO WAY to test this change, do not try.


nsavoire (approved):
David is back ❤️

PR #332, opened 2023-10-20
Add timestamp to allocation using PerfClock

nsavoire:
# What does this PR do?

Add timestamp information to allocation events using perf clock.=

Perf clock parameters are transferred to the library through metadata page.

### DIFF
--- a/src/ringbuffer_utils.cc
+++ b/src/ringbuffer_utils.cc
@@ -7,13 +7,29 @@
 
 #include "ddres_helpers.hpp"
 #include "ipc.hpp"
+#include "perf_clock.hpp"
 #include "pevent.hpp"
 #include "pevent_lib.hpp"
 #include "syscalls.hpp"
+#include "tsc_clock.hpp"
 
 #include <sys/eventfd.h>
 
 namespace ddprof {
+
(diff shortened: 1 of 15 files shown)


src/ringbuffer_utils.cc
```diff
@@ -59,6 +75,14 @@ DDRes ring_buffer_create(size_t buffer_size_page_order,
   pevent->custom_event = custom_event;
   pevent->ring_buffer_type = ring_buffer_type;
   pevent->ring_buffer_size = buffer_size;
+
+  if (ring_buffer_type == RingBufferType::kMPSCRingBuffer) {
```
r1viollet, line 79: The workflow is slightly strange as we `mmap` and `munmap` the buffer within the init.
Though as discussed we do not have the mmap within the creation workflow.

r1viollet: Perhaps we can add a comment to justify the logic

nsavoire: Done

r1viollet (approved):
LGTM!
Very cool mechanism to have consistent time information across different event types.

PR #333, opened 2023-10-20
Add libdatadog5 and timestamps

sanchda:
# What does this PR do?

Adds timestamps and upgrades to libdatadog5

Screenshot 2023-10-20 at 8 05 51 PM

# Motivation

Some fullhost analyses would benefit from the presence of timestamps

# Additional Notes

# How to test the change?

### DIFF
--- a/include/scope.hpp
+++ b/include/scope.hpp
@@ -79,8 +79,8 @@ template <typename T> class box {
 
 public:
   template <typename TT, typename GG>
-  requires std::is_constructible_v<T, TT>
-  explicit box(TT &&t, GG &&guard) noexcept(noexcept(box((T &&) t)))
+    requires std::is_constructible_v<T, TT>
+  explicit box(TT &&t, GG &&guard) noexcept(noexcept(box((T &&)t)))
       : box(std::forward<TT>(t)) {
     guard.release();
   }
@@ -101,8 +101,9 @@ template <typename T> class box<T &> {
 
 public:
   template <typename TT, typename GG>
-  requires std::is_convertible_v<TT, T &> box(TT &&t, GG &&guard)
-  noexcept(noexcept(static_cast<T &>((TT &&) t))) : value(static_cast<T &>(t)) {
+    requires std::is_convertible_v<TT, T &>
+  box(TT &&t, GG &&guard) noexcept(noexcept(static_cast<T &>((TT &&)t)))
+      : value(static_cast<T &>(t)) {
     guard.release();
   }
   T &get() const noexcept { return value.get(); }
@@ -212,8 +213,8 @@ class [[nodiscard]] basic_scope_exit : Policy {
 
 public:
   template <typename EFP>
-  requires _ctor_from<EFP>::value explicit basic_scope_exit(EFP &&ef) noexcept(
-      _noexcept_ctor_from<EFP>::value)
+    requires _ctor_from<EFP>::value
+  explicit basic_scope_exit(EFP &&ef) noexcept(_noexcept_ctor_from<EFP>::value)
       : exit_function(std::forward<EFP>(ef),
                       _make_failsafe(_noexcept_ctor_from<EFP>{}, &ef)) {}
   basic_scope_exit(basic_scope_exit &&that) noexcept(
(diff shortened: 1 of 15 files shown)


include/scope.hpp
```diff
@@ -259,14 +258,14 @@ template <typename R, typename D> class unique_resource {
...
-                 std::is_constructible_v<detail::box<D>, DD,
-                                         detail::_empty_scope_exit>
-  unique_resource(RR &&r, DD &&d, bool should_run) noexcept(
-      noexcept(detail::box<R>(std::forward<RR>(r), detail::_empty_scope_exit{}))
-          && noexcept(detail::box<D>(std::forward<DD>(d),
-                                     detail::_empty_scope_exit{})))
+  requires std::is_constructible_v<detail::box<R>, RR,
+                                   detail::_empty_scope_exit> &&
+      std::is_constructible_v<detail::box<D>, DD, detail::_empty_scope_exit>
+      unique_resource(RR &&r, DD &&d, bool should_run)
```
r1viollet, line 264: you might want to upgrade your llvm version (because clang format changed, and will block you in CI)

src/ddprof_cli.cc
```diff
@@ -233,6 +233,9 @@ int DDProfCLI::parse(int argc, const char *argv[]) {
   app.add_flag("--show_samples", show_samples,
                "Display captured samples as logs.\n")
       ->group("Debug options");
+  app.add_flag("--use_timestamps", use_timestamps,
+               "Put timestamps on sampled events.\n")
```
r1viollet, line 237: This feature could mention timeline (as this will make more sense for a customer)
When timeline is supported, I guess you would expect this to just be a `--timeline` option ?

src/ddprof_worker.cc
```diff
@@ -236,9 +237,12 @@ DDRes aggregate_livealloc_stack(
     const LiveAllocation::PprofStacks::value_type &alloc_info,
     DDProfContext &ctx, const PerfWatcher *watcher, DDProfPProf *pprof,
     const SymbolHdr &symbol_hdr) {
-  DDRES_CHECK_FWD(
-      pprof_aggregate(&alloc_info.first, symbol_hdr, alloc_info.second._value,
-                      alloc_info.second._count, watcher, kLiveSumPos, pprof));
+  DDProfValuePack pack{
+      alloc_info.second._value,
+      static_cast<uint64_t>(std::max<int64_t>(0, alloc_info.second._count)), 0};
```
r1viollet, line 242: As discussed, the time here could be tied to the time of allocation.
Though this would not have immediate value for the front end (only diagnostic value for the cases where we open the pprofs).

r1viollet (dismissed):
LGTM
Thanks for this PR ! :+1: 
Exciting work!

include/scope.hpp
sanchda: Good catch

src/ddprof_worker.cc
```diff
@@ -410,8 +414,10 @@ DDRes ddprof_pr_sample(DDProfContext &ctx, perf_event_sample *sample,
       // in lib mode we don't aggregate (protect to avoid link failures)
       int const i_export = ctx.worker_ctx.i_current_pprof;
       DDProfPProf *pprof = ctx.worker_ctx.pprof[i_export];
-      DDRES_CHECK_FWD(pprof_aggregate(&us->output, us->symbol_hdr, sample_val,
-                                      1, watcher, kSumPos, pprof));
+      const DDProfValuePack pack{static_cast<int64_t>(sample_val), 1,
+                                 ctx.params.timeline ? sample->time : 0};
```
nsavoire, line 418: Currently timestamps have an unspecified origin, could this be an issue ?

nsavoire:
Nice work, timeline is coming to native !

src/ddprof_worker.cc
sanchda: Indeed; I think this is a big problem for numerous reasons.

* Timeline currently has no means to extract the underlying clock source.  It is my belief that timeline should be able to comprehend this in order to determine whether two events on the same host can be ordered.  My primary interest in opening this PR was to add timestamps to fullhost in order to augment some work other teams are doing around sizing and cost analysis.  I think this will have enormous consequences for various products.
* Different events may have different underlying clock sources (in full generality) and there may be good reason to do so, in fact (e.g., if it's a high-frequency event or if we're bumping up against the kernel's event-taking overhead limit).
* As a matter of general principal, I think clocks are tricky and leaving the origin underspecified can only bring us trouble

At the same time, I wasn't so concerned about these topics that I felt the need to fix them in this PR--hopefully we can figure out a general strategy sometime, but it probably doesn't affect what we want to do next.  :)

src/exporter/ddprof_exporter.cc
```diff
@@ -292,7 +292,7 @@ DDRes ddprof_exporter_export(ddog_prof_Profile *profile,
                              DDProfExporter *exporter) {
   DDRes res = ddres_init();
   ddog_prof_Profile_SerializeResult serialized_result =
-      ddog_prof_Profile_serialize(profile, nullptr, nullptr);
```
r1viollet, line 295: minor, this will create a lz4 file, I can change the naming later.

r1viollet (approved):
LGTM :rocket:

src/ddprof_worker.cc
r1viollet: We will postpone the timeline allocation story for now. Other runtimes have not specified this.

PR #335, opened 2023-10-23
Sanchda/timeline to realtime

sanchda:
# What does this PR do?

Adjusts the timestamps we send to be epoch-adjusted.  This is a requirement of the Timeline and makes it much easier for end-users to compare outside of ddprof.

# Motivation

Make timestamps great again

# Additional Notes

This implementation presumes that timestamps only need to be adjusted once per upload cycle.  From the perspective of NTP adjustments, this is probably a good tradeoff between an overly fine-grained consumption of clocks versus fidelity.

# How to test the change?

Gotta test.

### DIFF
--- a/include/ddprof_worker_context.hpp
+++ b/include/ddprof_worker_context.hpp
@@ -41,7 +41,7 @@ struct DDProfWorkerContext {
   uint32_t count_worker{0}; // exports since last cache clear
   std::array<uint64_t, kMaxTypeWatcher> lost_events_per_watcher{};
   LiveAllocation live_allocation;
-  std::chrono::sys_time perf_clock_epoch{};
+  int64_t perfclock_offset;
 };
 
 } // namespace ddprof
--- a/include/ddprof_worker_context.hpp
+++ b/include/ddprof_worker_context.hpp
@@ -41,6 +41,7 @@ struct DDProfWorkerContext {
   uint32_t count_worker{0}; // exports since last cache clear
   std::array<uint64_t, kMaxTypeWatcher> lost_events_per_watcher{};
   LiveAllocation live_allocation;
+  int64_t mono_to_realtime_offset{0};
 };
 
 } // namespace ddprof
--- a/include/ddprof_worker_context.hpp
+++ b/include/ddprof_worker_context.hpp
@@ -41,7 +41,7 @@ struct DDProfWorkerContext {
   uint32_t count_worker{0}; // exports since last cache clear
   std::array<uint64_t, kMaxTypeWatcher> lost_events_per_watcher{};
   LiveAllocation live_allocation;
-  int64_t mono_to_realtime_offset{0};
+  std::chrono::sys_time perf_clock_epoch{};
 };
 
 } // namespace ddprof
--- a/src/ddprof_worker.cc
+++ b/src/ddprof_worker.cc
@@ -414,10 +414,8 @@ DDRes ddprof_pr_sample(DDProfContext &ctx, perf_event_sample *sample,
       // in lib mode we don't aggregate (protect to avoid link failures)
       int const i_export = ctx.worker_ctx.i_current_pprof;
       DDProfPProf *pprof = ctx.worker_ctx.pprof[i_export];
-      DDProfValuePack pack{static_cast<int64_t>(sample_val), 1, 0};
-      if (ctx.params.use_timestamps) {
-        pack.timestamp = sample->time;
-      }
+      DDProfValuePack pack{static_cast<int64_t>(sample_val), 1,
+                           ctx.params.timeline ? sample->time : 0};
       DDRES_CHECK_FWD(pprof_aggregate(&us->output, us->symbol_hdr, pack,
                                       watcher, kSumPos, pprof));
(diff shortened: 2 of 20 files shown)


include/ddprof_worker_context.hpp
```diff
@@ -41,6 +41,7 @@ struct DDProfWorkerContext {
   uint32_t count_worker{0}; // exports since last cache clear
   std::array<uint64_t, kMaxTypeWatcher> lost_events_per_watcher{};
   LiveAllocation live_allocation;
+  int64_t mono_to_realtime_offset{0};
```
nsavoire, line 44:
```suggestion
  std::chrono::sys_time perf_clock_epoch{};
```

src/ddprof_worker.cc
```diff
@@ -88,9 +88,26 @@ DDRes report_lost_events(DDProfContext &ctx) {
   return {};
 }
 
+inline int64_t get_mono_to_realtime_offset(
+    const std::chrono::steady_clock::time_point &steady_now) {
```
nsavoire, line 92: I think the logic is wrong here:
std::chrono::steady_clock is the same `CLOCK_MONOTONIC` so you are basically computing the offset between `CLOCK_MONOTONIC_RAW` and `CLOCK_MONOTONIC`.
You should use `std::chrono::system_clock`.
Also you should not use `CLOCK_MONOTONIC_RAW` but `PerfClock` instead: `PerfClock` is defined as the clock used for events from perf.
You could represent the offset as the epoch of PerfClock in the `system_clock` referential:
```
return std::chrono::system_clock::now() - PerfClock::now().time_since_epoch();
```

src/ddprof_worker.cc
```diff
@@ -88,9 +88,26 @@ DDRes report_lost_events(DDProfContext &ctx) {
...
+  auto ns_realtime = steady_now.time_since_epoch().count();
+  auto ns_monotonic_raw = timespec_to_duration(ts_monotonic_raw).count();
+
+  return ns_realtime - ns_monotonic_raw;
+}
+
 inline void export_time_set(DDProfContext &ctx) {
-  ctx.worker_ctx.send_time =
-      std::chrono::steady_clock::now() + ctx.params.upload_period;
+  auto steady_now = std::chrono::steady_clock::now();
```
nsavoire, line 106: Not sure it's a good idea to reuse `std::chrono::steady_clock::now()` value, just let `get_mono_to_realtime_offset` call again `std::chrono::steady_clock::now()` (it should be `std::chrono::system_time::now()` anyway), and cost does not matter since it's called once per minute.

src/ddprof_worker.cc
```diff
@@ -88,9 +88,26 @@ DDRes report_lost_events(DDProfContext &ctx) {
   return {};
 }
 
+inline int64_t get_mono_to_realtime_offset(
```
nsavoire, line 91: Why `inline` ?

src/ddprof_worker.cc
sanchda: Ah, good point.  Indeed, this seems fundamentally wrong.  Will review your notes.  Strong agree that `PerfClock` is better to use, now that I see it, since at the very least there's a stronger guarantee (thanks to your hard work!!!) that this is the correct underlying clock for the moment.

src/ddprof_worker.cc
sanchda: Oops, fixed.  I was thinking something else at the time.

src/ddprof_worker.cc
sanchda: Agreed, fixed

src/ddprof_worker.cc
sanchda: OK, this was a good suggestion and I changed the strategy to use the PerfClock.  Confirmed that the offsets are better.  Thank you!

sanchda: Fixed in 1a39a1c

src/ddprof_worker.cc
```diff
@@ -88,9 +88,22 @@ DDRes report_lost_events(DDProfContext &ctx) {
...
 }
 
+int64_t get_perfclock_offset() {
+  auto system_now = std::chrono::system_clock::now();
+  auto perf_now = ddprof::PerfClock::now();
+
+  auto system_ns = std::chrono::duration_cast<std::chrono::nanoseconds>(
+                       system_now.time_since_epoch())
+                       .count();
+  auto perf_ns = perf_now.time_since_epoch().count(); // assumes offset is in ns
```
nsavoire, line 98: You don't have to assume, `perf_now.time_since_epoch()` returns a duration and you can use c++ type system to ensure conversions are correct:
```
auto perf_clock_epoch = std::chrono::system_clock::now() - ddprof::PerfClock::now().time_since_epoch();
return std::chrono::nanoseconds(perf_clock_epoch.time_since_epoch()).count();
```

You could return a typed duration (std::chrono::nanoseconds) or even better a `std::chrono::system_clock::time_point` which would represent the epoch of `PerfClock` in `system_clock` referential.

nsavoire (approved):
Thanks !

PR #336, opened 2023-10-24
Remove some dependencies on ddprof::Parser

nsavoire:
Split ddprof_cmdline into 2 files
This allows to removing dependency on ddprof::Parser for profiling lib and some unit tests.


r1viollet (approved):
LGTM
Good idea.

PR #337, opened 2023-10-24
Add function_ref

nsavoire:
# Based on PR #336

Add https://github.com/zhihaoy/nontype_functional as third party, this provides a [function_ref](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2022/p0792r10.html) implementation (non-owning cheap alternative to `std::function`).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -213,7 +213,6 @@ set(DD_PROFILING_SOURCES
     src/lib/saveregisters.cc
     src/lib/symbol_overrides.cc
     src/logger.cc
-    src/logger_setup.cc
     src/perf.cc
     src/perf_clock.cc
     src/perf_ringbuffer.cc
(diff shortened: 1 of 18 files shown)


r1viollet:
Where do we need this ?

nsavoire:
> Where do we need this ?

In an earlier version of my PR to reduce allocation event size, I was passing a callback to `save_context` to reserve size in ring buffer. `function_ref` allowed to pass a capturing lambda without the extra cost associated to `std::function`.
Since I changed my implementation, I do not need this anymore, I changed the PR status to draft.

r1viollet:
> > Where do we need this ?
> 
> In an earlier version of my PR to reduce allocation event size, I was passing a callback to `save_context` to reserve size in ring buffer. `function_ref` allowed to pass a capturing lambda without the extra cost associated to `std::function`. Since I changed my implementation, I do not need this anymore, I changed the PR status to draft.

Thanks for leaving as draft. I was actually considering std::functions recently and was not sure of the cost.

PR #338, opened 2023-10-24
Reduce size of allocation events

nsavoire:
# What does this PR do?

Instead of using fixed sized allocation events, estimate the required stack size (with some margin) before calling reserve on ring buffer.


r1viollet:
That makes a lot of sense :+1: . I will wait for CI to stop complaining to approve

r1viollet (approved):
LGTM

PR #339, opened 2023-10-24
Tracepoints - add a check for debug fs using pid 1

r1viollet:
# What does this PR do?

- Tracepoints - add a check for debug fs using pid 1
- Minor fix on pprof naming - Add lz4 suffix

# Motivation

When using tracepoints on full host, this allows us to find the tracepoint IDs.
When capturing pprofs, the lack of extension was misleading.

# Additional Notes

N.A.

# How to test the change?

TBD

### DIFF
--- a/src/exporter/ddprof_exporter.cc
+++ b/src/exporter/ddprof_exporter.cc
@@ -52,7 +52,7 @@ DDRes create_pprof_file(ddog_Timespec start, const char *dbg_pprof_prefix,
   strftime(time_start, std::size(time_start), "%Y%m%dT%H%M%SZ", tm_start);
 
   char filename[PATH_MAX];
-  snprintf(filename, std::size(filename), "%s%s.pprof", dbg_pprof_prefix,
+  snprintf(filename, std::size(filename), "%s%s.pprof.lz4", dbg_pprof_prefix,
            time_start);
   LG_NTC("[EXPORTER] Writing pprof to file %s", filename);
   constexpr int read_write_user_only = 0600;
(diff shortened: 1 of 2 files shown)


src/exporter/ddprof_exporter.cc
```diff
@@ -52,7 +52,7 @@ DDRes create_pprof_file(ddog_Timespec start, const char *dbg_pprof_prefix,
   strftime(time_start, std::size(time_start), "%Y%m%dT%H%M%SZ", tm_start);
 
   char filename[PATH_MAX];
-  snprintf(filename, std::size(filename), "%s%s.pprof", dbg_pprof_prefix,
+  snprintf(filename, std::size(filename), "%s%s.pprof.lz4", dbg_pprof_prefix,
```
sanchda, line 55: Good catch

src/tracepoint_config.cc
```diff
@@ -23,6 +23,9 @@ int64_t tracepoint_get_id(std::string_view global_name,
     fs_path = "/sys/kernel/tracing/events";
   } else if (stat("/sys/kernel/debug/tracing/events", &sb) == 0) {
     fs_path = "/sys/kernel/debug/tracing/events";
+  }
+  if (stat("/proc/1/root/sys/kernel/debug/tracing/events", &sb) == 0) {
+    fs_path = "/proc/1/root/sys/kernel/debug/tracing/events";
```
sanchda, line 28:
```suggestion
  } else if (stat("/proc/1/root/sys/kernel/debug/tracing/events", &sb) == 0) {
    fs_path = "/proc/1/root/sys/kernel/debug/tracing/events";
```

sanchda: Precise binding order TBD.  I weakly suspect that preferring "official" tracefs or debugfs is preferable, but either way should probably not over-write a previous selection at the end.

sanchda (changes requested):

sanchda (approved):
Good stuff!  The PID 1 override was unexpected, but having seen it in action I'm not too surprised.  I suspect some orgs may benefit from that.

We still ought to handle this differently in the `perf_event_open` setup, but that may require some head-scratching.  :)

PR #341, opened 2023-11-06
Reduce check_clock_vdso duration

nsavoire:
# What does this PR do?

Reduce `check_clock_vdso` duration down to 10ms, while checking that at least 5ms of CPU time was spent.
This might fail to detect non-vdso timer calls in some rare cases, but this seems acceptable.


r1viollet (approved):
LGTM

r1viollet:
The benchmarks are surprisingly accurate on this change

PR #342, opened 2023-11-10
Make allocation profiling work after exec

nsavoire:
# What does this PR do?

This PR add the ability to profile allocations even after a process has exec'd into a new binary image.

# How do this work ?

ddprof now keeps a unix socket open, unix socket path can be overridden with `--socket` input option.
Worker accepts connections on this socket in a separate thread and sends ring buffer information upon request.
When a new process is exec'd, library looks up `DD_PROFILING_NATIVE_LIB_SOCKET` env variable to get socket path and if it exists, connects to it otherwise it spawns a ddprof process, then uses a pipe to synchronize with ddprof start and get socket path, and then connects to it.
By default ddprof creates an abstract socket, that does not exist on the filesystem: `\0/tmp/ddprof-<pid>.sock`.
Exec following can be disabled by setting env variable `DD_PROFILING_NATIVE_ALLOCATION_PROFILING_FOLLOW_EXECS` to 0/off/no.

### DIFF
--- a/include/ipc.hpp
+++ b/include/ipc.hpp
@@ -80,13 +80,12 @@ struct RingBufferInfo {
 };
 
 struct ReplyMessage {
-  enum { kLiveCallgraph = 0 };
+  enum { kLiveSum = 0x1 };
   // reply with the request flags from the request
   uint32_t request = 0;
   // profiler pid
   int32_t pid = -1;
   int64_t allocation_profiling_rate = 0;
-  // RingBufferInfo is returned if request & kRingBuffer
   // cppcheck-suppress unusedStructMember
   RingBufferInfo ring_buffer;
   uint32_t initial_loaded_libs_check_delay_ms = 0;
--- a/include/ipc.hpp
+++ b/include/ipc.hpp
@@ -11,8 +11,10 @@
 
 #include <chrono>
 #include <functional>
+#include <latch>
 #include <span>
 #include <system_error>
+#include <thread>
 
 namespace ddprof {
 
@@ -28,6 +30,8 @@ public:
   using socket_t = int;
 
   explicit UnixSocket(socket_t handle) noexcept : _handle(handle) {}
+  explicit UnixSocket(UniqueFd &&handle) noexcept
+      : _handle(std::move(handle)) {}
 
   void close(std::error_code &ec) noexcept;
 
@@ -65,6 +69,7 @@ struct RequestMessage {
   enum { kProfilerInfo = 0x1 };
   // request is bit mask of request flags
   uint32_t request = 0;
+  pid_t pid = -1;
 };
 
 struct RingBufferInfo {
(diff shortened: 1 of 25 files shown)


include/ipc.hpp
```diff
@@ -90,27 +94,34 @@ struct ReplyMessage {
...
-  UnixSocket _socket;
+  friend WorkerServer start_worker_server(int socket, const ReplyMessage &msg);
+
+  WorkerServer(int socket, const ReplyMessage &msg);
+  void event_loop();
+
+  int _socket;
+  std::latch _latch;
+  ReplyMessage _msg;
+  std::jthread _loop_thread;
```
r1viollet, line 122: good idea to favour jthreads

test/simple_malloc-ut.sh
```diff
@@ -132,6 +132,12 @@ check "./ddprof --initial-loaded-libs-check-delay 500 ./test/simple_malloc --use
 # Test slow profiler startup
 check "env DD_PROFILING_NATIVE_STARTUP_WAIT_MS=200 ./ddprof ./test/simple_malloc ${opts}" 1
 
+# Test follow execs
+check "./ddprof bash -c \"./test/simple_malloc ${opts}\"" 1
+
+# Test disabling follow execs
+check "env DD_PROFILING_NATIVE_ALLOCATION_PROFILING_FOLLOW_EXECS=0 DD_PROFILING_NATIVE_STARTUP_WAIT_MS=200 ./ddprof bash -c \"./test/simple_malloc ${opts}\"" 0 0 "inuse-space"
```
r1viollet, line 139: Interesting to test the removal of the feature. For overhead, there might be a use case where users do not want to follow some execs. Good thinking.

src/ddprof_cli.cc
```diff
@@ -297,10 +297,14 @@ int DDProfCLI::parse(int argc, const char *argv[]) {
       app.add_flag("--help_extended", help_extended, "Show extended options")
           ->group(""));
   extended_options.push_back(
-      app.add_option("--socket", socket,
-                     "Profiler's IPC socket, as a file descriptor")
+      app.add_option("--socket", socket_path, "Profiler socket path")
```
r1viollet, line 300: Minor: we could give a more detailed explanation of why we establish this socket.

src/lib/dd_profiling.cc
```diff
@@ -277,38 +287,63 @@ void notify_fork() {
...
-    //    for a request and replies
-    //  * both library and profiler close their socket and continue
-    int sockfds[2];
-    if (socketpair(AF_UNIX, SOCK_SEQPACKET, 0, sockfds) == -1) {
-      return -1;
-    }
 
-    UniqueFd parent_socket{sockfds[0]};
-    UniqueFd child_socket{sockfds[1]};
+  // Library communicates with profiler through a unix socket.
```
r1viollet, line 295: this is great

src/ddprof_cli.cc
r1viollet: As a user I would be confused with what I can do with this.
The overview is great.

r1viollet: I guess these are more developer options. We do not use them apart from testing use cases. Would something like:
`Override the automatically created socket with a specific path`
Be better ?

src/exe/main.cc
```diff
@@ -171,7 +171,7 @@ DDRes get_library_path(TempFileHolder &libdd_profiling_path,
 
 DDRes check_incompatible_options(const DDProfContext &ctx) {
   if (context_allocation_profiling_watcher_idx(ctx) != -1 && ctx.params.pid &&
```
r1viollet, line 173: CLI has something to manage incompatibilities. The only part we can not check is if the allocation profiling is not active.
Which makes my comment only half useful ^^

src/perf_mainloop.cc
```diff
@@ -109,6 +111,45 @@ DDRes spawn_workers(PersistentWorkerState *persistent_worker_state,
   return {};
 }
 
+ReplyMessage create_reply_message(const DDProfContext &ctx) {
```
r1viollet, line 114: minor: it feels slightly weird to have this function defined here.

src/perf_mainloop.cc
```diff
@@ -109,6 +111,45 @@ DDRes spawn_workers(PersistentWorkerState *persistent_worker_state,
...
+      reply.stack_sample_size =
+          ctx.watchers[alloc_watcher_idx].options.stack_sample_size;
+      reply.initial_loaded_libs_check_delay_ms =
+          ctx.params.initial_loaded_libs_check_delay.count();
+      reply.loaded_libs_check_interval_ms =
+          ctx.params.loaded_libs_check_interval.count();
+
+      if (ctx.watchers[alloc_watcher_idx].aggregation_mode ==
+          EventAggregationMode::kLiveSum) {
+        reply.allocation_flags |= (1 << ReplyMessage::kLiveCallgraph);
```
r1viollet, line 145: it seems someone forgot to rename the options all the way down :sweat_smile:

src/perf_mainloop.cc
```diff
@@ -205,13 +246,16 @@ DDRes worker_loop(DDProfContext &ctx, const WorkerAttr *attr,
...
   // one extra slot in pfd to accommodate for signal fd
-  struct pollfd poll_fds[k_max_nb_perf_event_open + 1];
+  pollfd poll_fds[k_max_nb_perf_event_open + 1];
   int pfd_len = 0;
   pollfd_setup(&ctx.worker_ctx.pevent_hdr, poll_fds, &pfd_len);
 
   DDRES_CHECK_FWD(signalfd_setup(&poll_fds[pfd_len]));
   int const signal_pos = pfd_len++;
 
+  WorkerServer const server =
```
r1viollet, line 256: Somehow I had not thought about making this a separate thread. I thought we would make it live within the mainloop.
I think it is better this way.

r1viollet (approved):
LGTM
A very impressive capability.
I am eager to start playing with this.

r1viollet (approved):
LGTM

PR #343, opened 2023-11-14
Logger rate limiting

nsavoire:
# What does this PR do?

Implement log rate limiting when log level is above debug. Rate limit is set at 100 log/s.

### DIFF
--- a/include/logger_setup.hpp
+++ b/include/logger_setup.hpp
@@ -5,6 +5,13 @@
 
 #pragma once
 
+#include <cstdint>
+
 namespace ddprof {
-void setup_logger(const char *log_mode, const char *log_level);
-}
+inline constexpr auto kMaxLogPerSecForNonDebug = 100;
+
+void setup_logger(
+    const char *log_mode, const char *log_level,
+    uint64_t max_log_per_sec_for_non_debug = kMaxLogPerSecForNonDebug);
+
+} // namespace ddprof
--- a/include/ratelimiter.hpp
+++ b/include/ratelimiter.hpp
@@ -1,0 +1,60 @@
+/*
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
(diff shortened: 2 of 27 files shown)


include/ratelimiter.hpp
```diff
@@ -0,0 +1,60 @@
...
+  IntervalRateLimiter(uint64_t max_count_per_interval,
+                      std::chrono::nanoseconds interval) noexcept
+      : _max_count_per_interval(max_count_per_interval), _interval(interval) {}
+
+  bool check() {
+    auto old_count = _count.fetch_add(1, std::memory_order_acq_rel);
+    if (old_count < _max_count_per_interval) {
+      return true;
+    }
+    return check_slow();
```
r1viollet, line 45: should we not change slightly this pattern to avoid checking at every log ?

r1viollet (approved):
LGTM
A good addition
I think we could still make it more efficient in heavy logging cases, though this is probably only useful in pathological cases that never occur.

include/logger_setup.hpp
```diff
@@ -5,6 +5,13 @@
 
 #pragma once
 
+#include <cstdint>
+
 namespace ddprof {
-void setup_logger(const char *log_mode, const char *log_level);
-}
+inline constexpr auto kMaxLogPerSecForNonDebug = 100;
```
r1viollet, line 11: ok for 100 :+1:

r1viollet (approved):
LGTM

include/ratelimiter.hpp
r1viollet: Never mind, this is a pathological case, we should just avoid logging so much.

r1viollet (approved):
LGTM

sanchda (dismissed):
LGTM--thank you for the improvement!

sanchda (approved):

PR #344, opened 2023-11-16
Refactor symbol overrides

nsavoire:
# What does this PR do?

 * Instead of doing a pass over all relocations of all librariesfor each symbol to override, do a single pass for all symbols.
* Replace timer + signal with a time check when pushing a sample to trigger new loaded libs check.
* Loaded libs check now uses `dl_phdr_info.dlpi_adds` member to quickly get number of loaded libs and check if it has changed.

# Motivation

* Reduce overhead of symbol overriding on process startup.
* Remove clunky timer + signal code.

### DIFF
--- a/include/functionref.hpp
+++ b/include/functionref.hpp
@@ -1,0 +1,236 @@
+// Copyright 2019 The Abseil Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// -----------------------------------------------------------------------------
+// File: function_ref.hpp
+// -----------------------------------------------------------------------------
+//
+// This header file defines the `absl::FunctionRef` type for holding a
+// non-owning reference to an object of any invocable type. This function
+// reference is typically most useful as a type-erased argument type for
+// accepting function types that neither take ownership nor copy the type; using
+// the reference type in this case avoids a copy and an allocation. Best
+// practices of other non-owning reference-like objects (such as
+// `absl::string_view`) apply here.
+//
+//  An `absl::FunctionRef` is similar in usage to a `std::function` but has the
+//  following differences:
+//
+//  * It doesn't own the underlying object.
+//  * It doesn't have a null or empty state.
+//  * It never performs deep copies or allocations.
+//  * It's much faster and cheaper to construct.
+//  * It's trivially copyable and destructable.
+//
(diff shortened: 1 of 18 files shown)


include/functionref.hpp
```diff
@@ -0,0 +1,238 @@
+// Copyright 2019 The Abseil Authors.
```
r1viollet, line 1: Minor: Should we put this in third party ?
Ideally we can identify easily what possibly needs to be upgraded / comes from external projects.

include/runtime_symbol_lookup.hpp
```diff
@@ -51,25 +52,8 @@ class RuntimeSymbolLookup {
   }
 
 private:
-  struct hash_string {
```
r1viollet, line 54: thanks for the refactor

src/lib/allocation_tracker.cc
```diff
@@ -304,14 +314,30 @@ DDRes AllocationTracker::push_lost_sample(MPSCRingBufferWriter &writer,
...
+  event->header.type = PERF_RECORD_LOST;
+  auto now = PerfClock::now();
+  event->sample_id.time = now.time_since_epoch().count();
+  if (_state.pid == 0) {
+    _state.pid = getpid();
+  }
+  if (tl_state.tid == 0) {
+    tl_state.tid = ddprof::gettid();
+  }
+  event->sample_id.pid = _state.pid;
```
r1viollet, line 329: TID and PID info are somewhat questionable considering the lost samples can come from previously lost events ?
Any reason to add these ?

src/lib/allocation_tracker.cc
```diff
@@ -447,13 +484,14 @@ DDRes AllocationTracker::push_alloc_sample(uintptr_t addr,
...
 
   if (buffer.empty()) {
-    // ring buffer is full, increase lost count
-    _state.lost_count.fetch_add(1, std::memory_order_acq_rel);
-
     if (timeout) {
       DDRES_RETURN_ERROR_LOG(DD_WHAT_PERFRB,
                              "Unable to get write lock on ring buffer");
     }
+
```
r1viollet, line 491: why would we not account for timeouts ?

include/functionref.hpp
sanchda: I think you're right.  We may have other sources that are not properly mentioned in the third party licenses doc.

nsavoire: Done, cf. last commit

src/lib/allocation_tracker.cc
nsavoire: Accounting of lost events is done per pid, why not report it ?
tid is indeed more questionable, but since I was adding pid, I also added tid...

src/lib/allocation_tracker.cc
nsavoire: Indeed, I reverted this part.

src/lib/allocation_tracker.cc
r1viollet: maybe this can be interesting. We could also in time add a table of lost events per TID.

r1viollet (approved):
LGTM
Thanks for improving instrumentation performances!

PR #347, opened 2023-11-23
Improve string operations

nsavoire:


src/ddprof_process.cc
```diff
@@ -43,10 +42,10 @@ DDRes Process::read_cgroup_ns(pid_t pid, std::string_view path_to_proc,
                               CGroupId_t &cgroup) {
   cgroup = Process::kCGroupNsError;
   std::string const path =
-      string_format("%s/proc/%d/ns/cgroup", path_to_proc.data(), pid);
-  char buf[k_max_buf_cgroup_link];
-  ssize_t const len = readlink(path.c_str(), buf, k_max_buf_cgroup_link - 1);
-  if (len == -1) {
+      absl::Substitute("$0/proc/$1/ns/cgroup", path_to_proc, pid);
```
r1viollet, line 45: absl cat is usually the preferred API ?
From documentation
```
Note however, that absl::Substitute(), because it requires parsing a format string at run-time, is slower than absl::StrCat(). Choose Substitute() over StrCat() only when code clarity is more important than speed.
```

r1viollet (dismissed):
LGTM
Some minor comments that I can revisit later
Thanks for removing that old format API!

src/ddprof_process.cc
nsavoire: If you want maximum performance, yes, `StrCat` should be preferred to `Substitute`, but it is a lot less readable.
The decision tree is something like:
- do I need formatting ?
  - yes: use `absl::StrFormat`
  - no: do I want maximum performance (at the expense of readability) ?
     - yes: use `absl:StrCat`
     - no: use `absl::Substitute`

nsavoire: Replaced this occurence with `absl::StrCat`

r1viollet (approved):
LGTM
Feels like the future!

src/exe/main.cc
```diff
@@ -64,8 +64,7 @@ void maybe_slowdown_startup() {
   if (const char *s = getenv(k_startup_wait_ms_env_variable); s != nullptr) {
     std::string_view const sv{s};
     int wait_ms = 0;
-    auto [ptr, ec] = std::from_chars(sv.begin(), sv.end(), wait_ms);
-    if (ec == std::errc() && ptr == sv.end() && wait_ms > 0) {
+    if (absl::SimpleAtoi(s, &wait_ms) && wait_ms > 0) {
```
sjanel, line 67: Just for my information, what is the benefit of `absl::SimpleAtoi` against `std::from_chars` ?

nsavoire: Well, `std::from_chars` is more powerful (it can be used to parse a number when the end of the number is not known since it returned the found end as a pointer) but it makes checking for success more difficult in simple cases when we expect the input string to a single number. And moreover it does not accept a `string`/ `string_view` as input.

sjanel: Got it, thx ! For instance with `"123 4"` `std::from_chars` will return `123` without error.

nsavoire: `std::from_chars` would also return success for `123qwerty`.
 The other difference is that `absl::SimpleAtoi` will ignore leading whitespaces whereas those would cause `std::from_chars` to fail.

PR #348, opened 2023-11-24
Do not override weak undefined symbols

nsavoire:
# What does this PR do?

Overriding a weak undefined symbol will change its value from null to a non-null value and might change the behaviour of programs checking the value of the symbol, leading to crashes when program then tries to call the overridden symbol.


r1viollet (approved):
LGTM
Thanks for a very fast fix!

PR #353, opened 2024-01-11
Remove TLReentryGuard from init_tl_state

nsavoire:
# What does this PR do?

Threal local reentry guard in init_tl_state can be safely removed: 
init_tl_state cannot be re-entered because it is only called from allocation_tracking_init and notify_thread_start.


r1viollet (approved):
LGTM
We can come back to this solution if the attachment to running threads is a priority
Right now we saw this had performance impacts

PR #354, opened 2024-01-11
Remove call to retrieve_stack_bounds

nsavoire:
# What does this PR do?

Call to retrieve_stack_bounds in push_alloc_sample is dangerous because retrieve_stack_bounds may allocate memory.
Make sure that retrieve_stack_bounds is called when intializing tl_state for thread.

# Motivation

Avoid calling `retrieve_stack_bounds` (that might allocate) in hooks.

### DIFF
--- a/src/lib/allocation_tracker.cc
+++ b/src/lib/allocation_tracker.cc
@@ -48,6 +48,7 @@ TrackerThreadLocalState *AllocationTracker::init_tl_state() {
   // allocation.
   auto tl_state = std::make_unique<TrackerThreadLocalState>();
   tl_state->tid = ddprof::gettid();
+  tl_state->stack_bounds = retrieve_stack_bounds();
 
   if (int res = pthread_setspecific(_tl_state_key, tl_state.get()); res != 0) {
     // should return 0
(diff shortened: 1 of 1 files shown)


src/lib/allocation_tracker.cc
```diff
@@ -48,6 +48,7 @@ TrackerThreadLocalState *AllocationTracker::init_tl_state() {
   // allocation.
   auto tl_state = std::make_unique<TrackerThreadLocalState>();
   tl_state->tid = ddprof::gettid();
+  tl_state->stack_bounds = retrieve_stack_bounds();
```
r1viollet, line 51: init flow is safe to allocations :+1:

src/lib/allocation_tracker.cc
```diff
@@ -443,14 +444,6 @@ DDRes AllocationTracker::push_alloc_sample(uintptr_t addr,
     push_lost_sample(writer, tl_state, notify_consumer);
   }
 
-  if (tl_state.stack_bounds.empty()) {
-    // This call should only occur on main thread
```
r1viollet, line 447: so this could crash for sure

src/lib/allocation_tracker.cc
```diff
@@ -582,10 +575,6 @@ void AllocationTracker::notify_thread_start() {
       return;
     }
   }
-
-  ReentryGuard const guard(&tl_state->reentry_guard);
-  tl_state->stack_bounds = retrieve_stack_bounds();
```
r1viollet, line 587: why are we removing this one ? This is the safe thread flow. This was good ?

r1viollet (dismissed):
LGTM

r1viollet (approved):
LGTM

PR #355, opened 2024-01-11
Remove useless malloc/free calls

nsavoire:
# What does this PR do?

Since malloc wrappers are not used anymore, these call are not needed anymore.


r1viollet (approved):
LGTM

PR #356, opened 2024-01-11
Improve pid/tid caching

nsavoire:
# What does this PR do?

* Avoid having to check pid/tid caching during sampling by initializing pid in TrackerState::init and updating pid/tid in notify_fork
* Make initialization of all TrackerThreadLocalState members explicit (previously they were zero-initialized by value initialization)
* Rename `_gen` to `gen` since the member is public

### DIFF
--- a/include/lib/allocation_tracker_tls.hpp
+++ b/include/lib/allocation_tracker_tls.hpp
@@ -8,21 +8,21 @@
 namespace ddprof {
 
 struct TrackerThreadLocalState {
-  int64_t remaining_bytes; // remaining allocation bytes until next sample
-  bool remaining_bytes_initialized; // false if remaining_bytes is not
-                                    // initialized
+  int64_t remaining_bytes{0}; // remaining allocation bytes until next sample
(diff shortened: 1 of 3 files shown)


include/lib/allocation_tracker_tls.hpp
```diff
@@ -8,21 +8,21 @@
...
+                             // and double counting of allocations (eg. when new
+                             // calls malloc, or malloc calls mmap internally)
 
-  bool reentry_guard; // prevent reentry in AllocationTracker (eg. when
-                      // allocation are done inside AllocationTracker) and
-                      // double counting of allocations (eg. when new calls
-                      // malloc, or malloc calls mmap internally)
+  // In the choice of random generators, this one is smaller
+  // - smaller than mt19937 (8 vs 5K)
+  std::minstd_rand gen{std::random_device{}()};
```
r1viollet, line 25: I was not aware of the naming rule :ok_hand:

r1viollet (approved):
LGTM

PR #357, opened 2024-01-11
Remove useless functionref.hpp header

nsavoire:
# What does this PR do?

FunctionRef is already provided by Abseil, so we don't need to provide it ourselves.


r1viollet (approved):
LGTM
Thanks for the cleanup

PR #358, opened 2024-01-11
Avoid doing allocations in m(un)map hooks

nsavoire:
# What does this PR do?

m(un)map hooks may be called inside an allocation function (eg. malloc) that is not reentrant, hence we must avoid allocations in these hooks.

### DIFF
--- a/src/lib/symbol_overrides.cc
+++ b/src/lib/symbol_overrides.cc
@@ -80,6 +80,16 @@ public:
     }
   }
 
+  // disallow allocation during tracking
+  void track_no_alloc(void *ptr, size_t size) {
+    if (_guard) {
+      tl_state()->allocation_allowed = false;
+      ddprof::AllocationTracker::track_allocation_s(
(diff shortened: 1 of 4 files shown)


src/lib/symbol_overrides.cc
```diff
@@ -103,6 +113,16 @@ class DeallocTrackerHelper {
     }
   }
 
+  // disallow allocation during tracking
+  void track_no_alloc(void *ptr) {
+    if (_guard) {
+      tl_state()->allocation_allowed = false;
```
r1viollet, line 119: is the reason we do this with a tl_state to help with testing ? 
Set the flag and check if any allocs are happening in the function ?

r1viollet (approved):
LGTM
My initial instinct was to create separate tracking flows.
I don't mind either solution

src/lib/symbol_overrides.cc
nsavoire: which alternative had you in mind ?

r1viollet (approved):
LGTM

PR #360, opened 2024-01-11
Add -fno-semantic-interposition

nsavoire:
# Motivation

"-fno-semantic-interposition" allows the compiler to inline code within a compilation unit even with "-fPIC".
Also make all symbols hidden by default amd use "-Bsymbolic-functions". cf. https://maskray.me/blog/2021-05-16-elf-interposition-and-bsymbolic cf. https://maskray.me/blog/2021-05-09-fno-semantic-interposition


r1viollet (approved):
LGTM

r1viollet:
@nsavoire feel free to merge this

PR #361, opened 2024-01-15
Fix file lookup (inode check removal)

r1viollet:
# What does this PR do?

Remove the need to match the inode when looking up files.

# Motivation

A customer was facing an issue due to the checks on inodes.

# Additional Notes

It is not clear to me why there is an inconsistency between proc maps and the disk lookup. I suspect a self modifying binary.
The risk could be that we missmatch a `libc` library for instance (between cgroups). However the alternative today seems to be a failure. 

# How to test the change?

The CI covers this change.

### DIFF
--- a/src/dso_hdr.cc
+++ b/src/dso_hdr.cc
@@ -518,12 +518,18 @@ FileInfo DsoHdr::find_file_info(const Dso &dso) {
   //   or      /host/proc/<pid>/root/usr/local/bin/exe_file
   std::string const proc_path = _path_to_proc + "/proc/" +
       std::to_string(dso._pid) + "/root" + dso._filename;
-  if (get_file_inode(proc_path.c_str(), &inode, &size) && inode == dso._inode) {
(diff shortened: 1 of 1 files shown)


sanchda (approved):
Makes sense.  Let's try it and see what happens!

mingmxu:
Verified working in my test, let's merge it. Btw when is the next release?

r1viollet:
> Verified working in my test, let's merge it. Btw when is the next release?

Thanks for the feedback! Would you have additional knowledge on why the inode is changing ? I was not able to build a test for this and it would be nice to have a representative test.

r1viollet:
> Verified working in my test, let's merge it. Btw when is the next release?

Do you need a patch release or is the custom build OK for now ? We can cut a release if this unblocks you.

mingmxu:
@r1viollet any timeline for the new release?

r1viollet:
> @r1viollet any timeline for the new release?

I can do a release during the week.

PR #366, opened 2024-01-23
Consider timestamps in comm and fork events

r1viollet:
# What does this PR do?

To prevent clearing mmap data when it is not useful, we consider timestamps of comm and fork events.

# Motivation

We encountered an issue where we gather mmap data just prior to a comm event. The comm event clears the data. 
Then we are unable to compute the offset from an elf file.

# Additional Notes

NA

# How to test the change?

I added a unit test.
I will check with user if this fixes the issue he encountered or if we need to go further.


src/unwind.cc
```diff
@@ -123,11 +123,16 @@ DDRes unwindstate_unwind(UnwindState *us) {
...
-  us->dso_hdr.pid_free(pid);
-  us->dwfl_hdr.clear_pid(pid);
-  us->symbol_hdr.clear(pid);
-  us->process_hdr.clear(pid);
+void unwind_pid_free(UnwindState *us, pid_t pid,
+                     PerfClock::time_point timestamp) {
+  if (!(us->dso_hdr.pid_free(pid, timestamp))) {
+    LG_DBG("(PID Free)%d -> avoid free of mappings (%ld)", pid,
+           timestamp.time_since_epoch().count());
+  } else {
```
r1viollet, line 131: if the mappings are good. Does it really mean that the underlying dwfl modules are OK ?
Today we do not really tie the lifetime of the mappings to the lifetimes of the unwinding caches.
So I am not sure about this change. I think it requires more careful thinking.
@nsavoire fyi

src/ddprof_worker.cc
```diff
@@ -584,21 +587,21 @@ void ddprof_pr_lost(DDProfContext &ctx, const perf_event_lost *lost,
...
 
 DDRes ddprof_pr_fork(DDProfContext &ctx, const perf_event_fork *frk,
-                     int watcher_pos) {
+                     int watcher_pos, PerfClock::time_point timestamp) {
   LG_DBG("<%d>(FORK)%d -> %d/%d", watcher_pos, frk->ppid, frk->pid, frk->tid);
   if (frk->ppid != frk->pid) {
     // Clear everything and populate at next error or with coming samples
-    DDRES_CHECK_FWD(worker_pid_free(ctx, frk->pid));
+    DDRES_CHECK_FWD(worker_pid_free(ctx, frk->pid, timestamp));
     ctx.worker_ctx.us->dso_hdr.pid_fork(frk->pid, frk->ppid);
```
nsavoire, line 605: It would be better to only call `pid_fork` when pid mappings have actually been freed.

src/ddprof_worker.cc
```diff
@@ -231,7 +232,7 @@ DDRes ddprof_unwind_sample(DDProfContext &ctx, perf_event_sample *sample,
   if (us->_dwfl_wrapper && us->_dwfl_wrapper->_inconsistent) {
     // Loaded modules were inconsistent, assume we should flush everything.
     LG_WRN("(Inconsistent DWFL/DSOs)%d - Free associated objects", us->pid);
-    DDRES_CHECK_FWD(worker_pid_free(ctx, us->pid));
```
nsavoire, line 234: It weird to have to call `PerfClock::now()` here, perhaps having two versions of `worker_pid_free` (with and without timestamp) or having `PerfClock::time_point::max()` as default value for `timestamp` in `worker_pid_free` would be more natural.

src/unwind.cc
nsavoire: I agree

PR #367, opened 2024-01-23
Align segment offset down to page size

nsavoire:
# What does this PR do?

Segment offset should not be aligned to `p_align` but to page size. `p_align` is only used by kernel/dynamic loader to find a suitable base address to load the elf file.

### DIFF
--- a/include/ddprof_module_lib.hpp
+++ b/include/ddprof_module_lib.hpp
@@ -24,12 +24,11 @@ inline uint32_t elf_flags_to_prot(Elf64_Word flags) {
 struct Segment {
   ElfAddress_t addr;
   Offset_t offset;
-  uint64_t alignment;
   uint32_t prot;
 };
 
 struct Mapping {
-  ElfAddress_t addr;
+  ProcessAddress_t addr;
   Offset_t offset;
   uint32_t prot;
 };
--- a/src/ddprof_module_lib.cc
+++ b/src/ddprof_module_lib.cc
@@ -9,6 +9,7 @@
 #include "ddres.hpp"
 #include "defer.hpp"
 #include "logger.hpp"
+#include "perf.hpp"
 #include "unique_fd.hpp"
 
 #include <algorithm>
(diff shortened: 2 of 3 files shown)


sanchda (dismissed):
Oh wow, interesting proposal.

So let's make sure I'm understanding this correctly.

`p_align` is a value defined in the ELF header, which is generated at build-time for the target machine.  It's almost always going to be 4096 on x86_64 and arm64, unless the build system is configured in a suitable way.  However, in practice segment alignment happens at load-time and must obey the limitations of the system (in particular, `mmap()` regions are strictly page-aligned).  This means that the precise value of `p_align` may be ignored or overridden.

I reviewed segment.c in the libdwfl code to validate my reasoning and it does appear that they have a `dwfl->segment_align` parameter that overrides `p_align` in various places.

I think you're right.  LGTM, good catch.

r1viollet (approved):
I believe you based on the bugs we noticed ;-)
This was not how I read elf specification.
LGTM

nsavoire:
> Oh wow, interesting proposal.
> 
> So let's make sure I'm understanding this correctly.
> 
> `p_align` is a value defined in the ELF header, which is generated at build-time for the target machine. It's almost always going to be 4096 on x86_64 and arm64, unless the build system is configured in a suitable way. However, in practice segment alignment happens at load-time and must obey the limitations of the system (in particular, `mmap()` regions are strictly page-aligned). This means that the precise value of `p_align` may be ignored or overridden.
> 
> I reviewed segment.c in the libdwfl code to validate my reasoning and it does appear that they have a `dwfl->segment_align` parameter that overrides `p_align` in various places.
> 
> I think you're right. LGTM, good catch.

I need to better document this, and I must admit that I don't fully understand what is the goal of `p_align` and how it interacts with the loading of elf file.
But I checked the kernel elf loader and there maximum `p_align` value is only used to [align the base address](https://github.com/torvalds/linux/blob/master/fs/binfmt_elf.c#L1090-L1092). [Mapping of segments](https://github.com/torvalds/linux/blob/master/fs/binfmt_elf.c#L1132-L1133) rounds down segment offset only to [page size](https://github.com/torvalds/linux/blob/master/fs/binfmt_elf.c#L362).

PR #369, opened 2024-01-25
Fix on backpopulate logics - init flow

r1viollet:
# What does this PR do?

- Allow a retry when attaching a mod with a backpopulate when attaching to a PID in the init flow
- Retry with a backpopulate regardless of the failure in the attachement.

# Motivation

Ensure a consistent behaviour between flows.

# Additional Notes

This should not occur often. Only in case of failures to attach.
Risks: In full host, our backpopulate rates are already very high. Could this make things worse ?

# How to test the change?

I do not have a specific test to cover this flow.
This would happen when mappings are ambiguous.

### DIFF
--- a/src/unwind_dwfl.cc
+++ b/src/unwind_dwfl.cc
@@ -247,83 +247,78 @@ DDRes add_runtime_symbol_frame(UnwindState *us, const Dso &dso, ElfAddress_t pc,
 
   return add_frame(symbol_idx, map_idx, pc, us);
 }
+
+bool try_attach_module(UnwindState *us, const DsoHdr::DsoMap &map,
+                       bool backpopulate_done) {
+  bool success = false;
+  for (auto it = map.cbegin(); it != map.cend(); ++it) {
+    const Dso &dso = it->second;
+    if (!dso.is_executable() || dso._type != DsoType::kStandard) {
+      continue;
+    }
+    const FileInfoId_t file_info_id = us->dso_hdr.get_or_insert_file_info(dso);
+    if (file_info_id <= k_file_info_error) {
+      LG_DBG("Unable to find file for DSO %s", dso.to_string().c_str());
+      continue;
+    }
+    const FileInfoValue &file_info_value =
+        us->dso_hdr.get_file_info_value(file_info_id);
+    DDProfMod *ddprof_mod = nullptr;
+    auto res = us->_dwfl_wrapper->register_mod(us->current_ip,
+                                               DsoHdr::get_elf_range(map, it),
+                                               file_info_value, &ddprof_mod);
+    if (IsDDResOK(res)) {
+      success = true;
+      break;
+    }
+    if (!backpopulate_done) {
+      // we have a chance after the next backpopulate to succeed in registering
+      file_info_value.reset_errored();
+    }
+  }
+  return success;
+}
 } // namespace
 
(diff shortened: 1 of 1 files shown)


src/unwind_dwfl.cc
```diff
@@ -248,66 +247,77 @@ DDRes add_runtime_symbol_frame(UnwindState *us, const Dso &dso, ElfAddress_t pc,
...
-
-    DsoHdr::DsoMap const &map = us->dso_hdr.get_pid_mapping(us->pid)._map;
-    if (map.empty()) {
-      int nb_elts;
-      us->dso_hdr.pid_backpopulate(us->pid, nb_elts);
-    }
+  if (us->_dwfl_wrapper->_attached) {
+    return {};
+  }
+  bool backpopulate_attempt = false;
```
nsavoire, line 290: I would rename `backpopulate_attempt` to `backpopulate_done` to make it clearer.

src/unwind_dwfl.cc
```diff
@@ -248,66 +247,77 @@ DDRes add_runtime_symbol_frame(UnwindState *us, const Dso &dso, ElfAddress_t pc,
 
   return add_frame(symbol_idx, map_idx, pc, us);
 }
+
+bool try_attach_module(UnwindState *us, const DsoHdr::DsoMap &map,
+                       bool &backpopulate_attempt) {
```
nsavoire, line 252: Why pass `backpopulate_attempt` by reference ?

src/unwind_dwfl.cc
```diff
@@ -248,66 +247,77 @@ DDRes add_runtime_symbol_frame(UnwindState *us, const Dso &dso, ElfAddress_t pc,
...
+        us->dso_hdr.get_file_info_value(file_info_id);
+    DDProfMod *ddprof_mod = nullptr;
+    auto res = us->_dwfl_wrapper->register_mod(us->current_ip,
+                                               DsoHdr::get_elf_range(map, it),
+                                               file_info_value, &ddprof_mod);
+    if (IsDDResOK(res)) {
+      success = true;
+      break;
+    }
+    if (!backpopulate_attempt) {
```
nsavoire, line 274: Maybe add a comment to explain the logic there

src/unwind_dwfl.cc
```diff
@@ -248,66 +247,77 @@ DDRes add_runtime_symbol_frame(UnwindState *us, const Dso &dso, ElfAddress_t pc,
...
-    }
-    if (!success) {
-      LG_DBG("Unable to attach a mod for PID%d", us->pid);
-      return ddres_warn(DD_WHAT_UW_ERROR);
+  bool success = try_attach_module(us, map, backpopulate_attempt);
+  if (!success && !backpopulate_attempt) {
+    int nb_elts_added;
+    if (us->dso_hdr.pid_backpopulate(us->pid, nb_elts_added) &&
+        nb_elts_added > 0) {
+      success = try_attach_module(us, map, backpopulate_attempt);
```
nsavoire, line 303: `backpopulate_attempt = true` probably missing or call `try_attach_module` directly with `true`:
```
success = try_attach_module(us, map, true);
```

r1viollet:
@nsavoire Thanks for the quick review and spotting the issue. I had initially slightly rushed that commit.
Things should look better.

nsavoire (approved):
LGTM !

PR #370, opened 2024-01-25
Improve DsoHdr::erase_range

nsavoire:
* Add DDPROF_CHECK_FATAL / DDPROF_DCHECK_FATAL assertions
* Add function to check DsoHdr invariants
* Change pid_fork to do nothing if child pid mappings already exist
    `pid_fork` directly copies mappings from parent to child pid,
    if child pid already has mappoings this can lead to overlapping mappings.
    Currently this cannot happens, because child mappings are cleared before
    `pid_fork`. This check adds a safegard by returning early if child pid
    has mappings.
* Add DsoHdr invariant check in debug mode after inserting a Dso
* Make erase_range function cleaner
    * Add check when re-inserting truncated last mapping
    * Return early in the case new mapping truncates end of a single mapping
      to avoid using range.first ierator that is invalidated by map::extract
      (the node pointed by first is extracted and then re-inserted hence
       the iterator should remain valid, but better be careful)
* Allow anonymous mapping to truncate existing file mapping
    Allow new mapping to truncate existing mappings if previous mapping is
    a file and new mapping is anonymous because a LOAD segment can have
    filesize < memsize as a size optimization when the end of segment is all
    zeros (eg. bss). In that case elf loader / dlopen will map the whole
    file + some extra space to cover the difference between memsize and file
    size and then map an anonymous mapping over the part not in the file.
* Avoid out-of-bound array access when printing unwinding error in debug logs

### DIFF
--- a/include/ddres_helpers.hpp
+++ b/include/ddres_helpers.hpp
@@ -8,7 +8,9 @@
 #include "ddres_def.hpp"
 #include "ddres_list.hpp"
 #include "logger.hpp"
+#include "unlikely.hpp"
 
+#include <cstdlib>
 #include <cstring>
 #include <system_error>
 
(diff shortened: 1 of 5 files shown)


r1viollet (approved):
LGTM
We had to do a dedicated session to review some of the changes here.
The erase_range logics are not easy to understand and require a careful understanding of linker behaviours.

PR #371, opened 2024-01-26
Store elf addresses inside the pprof locations

r1viollet:
# What does this PR do?

Store elf addresses inside the pprof locations

# Motivation

Remote symbolization

# Additional Notes

NA

# How to test the change?

Let's think of a test


r1viollet:
Thinking more about this, it is a bad idea. It is better to just send the virtual offset (not the file offset like it is done today).

PR #372, opened 2024-01-29
Remove check_invariants from hot code path

r1viollet:
# What does this PR do?

Move the check on invariants to a code path where it is called once per cycle (60 seconds)

# Motivation

Debug mode is not usable currently.

# Additional Notes

N/A

# How to test the change?

Already tested

### DIFF
--- a/src/ddprof_worker.cc
+++ b/src/ddprof_worker.cc
@@ -543,6 +543,10 @@ DDRes ddprof_worker_cycle(DDProfContext &ctx,
   // Increase the counts of exports
   ctx.worker_ctx.count_worker += 1;
 
+  // In debug mode, check for possible issues in loaded segments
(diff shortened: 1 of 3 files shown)


src/ddprof_worker.cc
```diff
@@ -543,6 +543,10 @@ DDRes ddprof_worker_cycle(DDProfContext &ctx,
   // Increase the counts of exports
   ctx.worker_ctx.count_worker += 1;
 
+  // In debug mode, check for possible issues in loaded segments
+  DDPROF_DCHECK_FATAL(ctx.worker_ctx.us->dso_hdr.check_invariants(),
```
nsavoire, line 547: Since this is run only once every 60s, we could probably make this a release check: `DDPROF_CHECK_FATAL`.

nsavoire (approved):

PR #373, opened 2024-01-30
Fix infinite loop when retry succeeds

nsavoire:
# What does this PR do?

Fix infinite loop when retry succeeds.
This loop is really hard to understand and should be simplified, but I could not easily extract it to a new function since it has many input/outputs and some direct returns.


sanchda (dismissed):

r1viollet (approved):
LGTM

PR #375, opened 2024-02-05
Add an hidden option to enable remote symbolization

nsavoire:
Enabling remote symbolization currently has 2 effects
* use elf addresses instead of process addresses in exported pprof
* add a `remote_symbols:yes` tag


r1viollet (approved):
LGTM

peterg17 (commented):
thanks!

PR #376, opened 2024-02-09
Symbolize through Blazesym

r1viollet:
# What does this PR do?

Remove elfutils and use blazesym as a symbolization API.
This removes the ddprof caches for symbols. Instead it just relies on blazesym caches.

# Motivation

- Removing the dependency on elfutils
- Better performances
- One flag away from inlining
- Better source code locations

# Additional Notes

For now we still have two separate flows for symbols. 
- The internal symbols we store in a symbol table (example: for jitdump)
- Blazesym symbols, which we populate during pprof aggregation

We loose split debug :-( . However this will be fixed at one point!

Related build pr : https://github.com/DataDog/ddprof-build/pull/131
Do not look at commits, they are a mess (sorry). I will squash merge everything.

# How to test the change?

I adjusted all unit tests to work with the new symbolization API.


src/pprof/ddprof_pprof.cc
```diff
@@ -56,6 +43,89 @@ struct ActiveIdsResult {
   PerfWatcher *default_watcher = nullptr;
 };
 
+bool is_ld(const std::string_view path) {
+  // path is expected to not contain slashes
+  assert(path.rfind('/') == std::string::npos);
```
r1viollet, line 48: I'm not 100% clear on why this is true.

include/ddog_profiling_utils.hpp
```diff
@@ -5,12 +5,122 @@
...
+    ProcessAddress_t ip,
+    ddprof::HeterogeneousLookupStringMap<std::string> &demangled_names,
+    const MapInfo &mapinfo, const blaze_sym *blaze_sym, unsigned &cur_loc,
+    std::span<ddog_prof_Location> locations_buff) {
+  if (cur_loc >= locations_buff.size()) {
+    return ddres_warn(DD_WHAT_UW_MAX_DEPTH);
+  }
+  constexpr std::string_view undef{};
+  constexpr std::string_view undef_inlined = undef;
+  for (unsigned i = 0; i < blaze_sym->inlined_cnt && cur_loc < kMaxStackDepth;
```
r1viollet, line 88: The Inlining can not be triggered. As I did not want to create too many changes to measure the operational impact of the PR.
Though we are just one flag away from having it.

include/pprof/ddprof_pprof.hpp
```diff
@@ -41,8 +46,10 @@ DDRes pprof_create_profile(DDProfPProf *pprof, DDProfContext &ctx);
  * @param pprof
  */
 DDRes pprof_aggregate(const UnwindOutput *uw_output,
-                      const SymbolHdr &symbol_hdr, const DDProfValuePack &pack,
-                      const PerfWatcher *watcher,
+                      const SymbolHdr &symbol_hdr,
+                      ddprof::Symbolizer *symbolizer,
```
r1viollet, line 50: minor: order of parameters could be better (symbolizer is not const)

include/unwind_output.hpp
```diff
@@ -14,30 +14,29 @@
...
 
+#include "ddprof_file_info-i.hpp"
+
 namespace ddprof {
 
 struct FunLoc {
-  ProcessAddress_t ip;   // process address
-  ElfAddress_t elf_addr; // elf address (in elf file virtual memory space)
+  ProcessAddress_t ip;
+  ElfAddress_t _elf_addr;
```
r1viollet, line 23: minor: naming convention

src/pprof/ddprof_pprof.cc
```diff
@@ -294,25 +364,85 @@ DDRes pprof_aggregate(const UnwindOutput *uw_output,
...
+      write_location(
+          &loc, mapinfo_table[loc._map_info_idx], symbol_table[loc._symbol_idx],
+          &locations_buff[write_index++], pprof->use_process_adresses);
+      ++index;
+      continue;
+    }
+
+    if (locs[index]._file_info_id <= k_file_info_error) {
+      // We should either have an index or a file to write the symbol
+      assert(0);
```
r1viollet, line 404: we maybe able to do better than assert thanks to recent work on assertions

r1viollet:
Closing in favour of : https://github.com/DataDog/ddprof/pull/390

PR #377, opened 2024-02-09
Fix self profiling

nsavoire:
# What does this PR do?

* Allow allocation profiler to profile itself
  Initialize all `<symbol>Hook::ref` members to `nullptr`. This is necessary because otherwise these will appear as relocations, and as such might be overridden by another instance of the profiler. `<symbol>Hook::ref` is filled by a lookup of the symbol.

* Avoid overriding symbols in profiling library

* Clear revert info after restoring overrides

### DIFF
--- a/src/lib/elfutils.cc
+++ b/src/lib/elfutils.cc
@@ -192,7 +192,6 @@ std::optional<DynamicInfo> retrieve_dynamic_info(const dl_phdr_info &info,
     if (phdr->p_type == PT_DYNAMIC) {
       phdr_dynamic = phdr;
     }
-    // Exclude this DSO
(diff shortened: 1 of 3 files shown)


src/lib/elfutils.cc
```diff
@@ -433,64 +434,89 @@ void override_symbol(std::string_view symbol_name, uintptr_t new_symbol,
...
-  }
-
   return _overrides
       .try_emplace(std::string{symbol_name}, ref_symbol, new_symbol,
                    do_not_override_this_symbol)
       .second;
 }
 
 void SymbolOverrides::apply_overrides() {
+  // Static / global variables bound to an extern function symbol as:
```
r1viollet, line 471: amazing comment

r1viollet (approved):
LGTM

PR #378, opened 2024-02-09
Only look for runtime symbols in executable mappings

nsavoire:
# What does this PR do?

Avoid looking for jit dump info when `pc` belongs to a non-executable mapping.


r1viollet (approved):
LGTM!
Thanks for looking into this

PR #380, opened 2024-02-16
Add option to disable symbolization

nsavoire:
# Motivation

Allow to completely disable symbolization so as to easily measure its memory / cpu overhead.


r1viollet (approved):
LGTM
Please improve motivation section before merging.
Thank you :bow:

PR #381, opened 2024-02-19
Compilation fixes for Linux Alpine 3.19

nsavoire:
# What does this PR do?

Small fixes for Alpine 3.19:
* missing include
* define `_LARGEFILE64_SOURCE` to get `mmap64` declaration from musl


r1viollet (approved):
LGTM

PR #384, opened 2024-03-06
Use Go build id as fallback for gnu build id

nsavoire:
# What does this PR do?

If the gnu build id is not available, use the Go build id if present.


r1viollet (dismissed):
LGTM

r1viollet (approved):
LGTM

PR #386, opened 2024-03-11
Fix dependency check

nsavoire:
# What does this PR do?

* Make sure script fails when dependency file is missing
* Alpine ddprof should not have any dependency


r1viollet (approved):
LGTM
Thanks for improving the testing and covering new alpine versions

PR #387, opened 2024-03-15
Add option to reorder events by timestamp across ring buffers

nsavoire:
# What does this PR do?

Reorder events from ring buffers before processing them.
Events in each perf ring buffer are already ordered by timestamp. For MPSC ring buffers, there is no such guarantee.
The strategy is to dequeue events from ring buffers and push them in a priority queue, ordered by timestamp. And then process this priority queue.
When a ring buffer is empty, we cannot be sure that a new event with a timestamp less than a previously enqueued event will not be added to the ring buffer in the future.
That's why we arbitrarily define a maximum latency `kMaxSampleLatency` between the timestamp of an event and the time it appears in the ring buffer, and we process events with timestamps up to (now - kMaxSampleLatency).

For MPSC ring buffers, we dequeue events and push them into the priority queue until we reach an event with a timestamp greater than (now - kMaxSampleLatency) or the ring buffer is empty. Note that when an event with a timestamp greater than (now - kMaxSampleLatency) is dequeued, it is still pushed in the priority_queue.

For perf ring buffers, we ensure that at anytime at most one event from each ring buffer is in the priority queue. This ensures that events with identical timestamps in a ring buffer are processed in the same order as in
the ring buffer (priority queue is not stable, therefore pushing events with identical timestamps might permute them) and this also makes advancing the reader cursor position in the ring buffer easier since know that only one event has been read, we can just bump the reader cursor to the last read position.

When an event from a perf ring buffer is dequeued fron the priority queue, we advance the reader cursor position in the ring buffer to free the slot for the writer, and attempt to read the next event from the ring buffer if not empty.

When an event from a MPSC ring buffer is dequeued from the priority queue, we try to advance the reader cursor position in the ring buffer to free the slot for the writer. Since events might be out of order for this ring buffer, advancing is done by marking processed events as discarded and bumping the reader position until empty or we reach the first non-discarded event.

# Motivation

Avoid inconsistencies when processing events out of order.

### DIFF
--- a/src/ddprof_cli.cc
+++ b/src/ddprof_cli.cc
@@ -354,6 +354,12 @@ int DDProfCLI::parse(int argc, const char *argv[]) {
                                  ->envname("DD_PROFILING_DISABLE_SYMBOLIZATION")
                                  ->group(""));
 
+  extended_options.push_back(
+      app.add_flag("--reorder-events,!--no-reorder-events", reorder_events,
+                   "Reorder perf events by timestamp")
+          ->default_val(false)
+          ->envname("DD_PROFILING_REORDER_EVENTS")
+          ->group(""));
   // Parse
   CLI11_PARSE(app, argc, argv);
 
--- a/src/perf_mainloop.cc
+++ b/src/perf_mainloop.cc
@@ -163,6 +163,13 @@ void pollfd_setup(std::span<PEvent> pes, struct pollfd *pfd) {
   }
 }
 
+// EventWrapper holds a reference to a perf_event_header with its associated
+// timestamp.
+// It is used to order events in a std::priority_queue without copying events.
+// perf_event_header is not owned by EventWrapper, it points on ring buffer
+// memory, and must remain valid during the lifetime of EventWrapper.
+// Consequently, care must be taken to advance reader cursor position in ring
+// buffer only after processing the event.
 struct EventWrapper {
   const perf_event_header *event;
   PerfClock::time_point timestamp;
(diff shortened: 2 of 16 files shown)


src/ddprof_cli.cc
```diff
@@ -354,6 +354,12 @@ int DDProfCLI::parse(int argc, const char *argv[]) {
                                  ->envname("DD_PROFILING_DISABLE_SYMBOLIZATION")
                                  ->group(""));
 
+  extended_options.push_back(
+      app.add_flag("--reorder_events,!--no-reorder-events", reorder_events,
```
r1viollet, line 358: Nice that we have an option

src/perf_mainloop.cc
```diff
@@ -72,31 +74,34 @@ void modify_sigprocmask(int how) {
...
   // child immediately exits the while() and returns from this function, whereas
   // the parent stays here forever, spawning workers.
-  while (!g_termination_requested && (child_pid = fork())) {
-    g_child_pid = child_pid;
-    {
-      LG_NTC("Created child %d", child_pid);
-      // unblock signals, we can now forward signals to child
-      modify_sigprocmask(SIG_UNBLOCK);
-      waitpid(g_child_pid, nullptr, 0);
+  while (!g_termination_requested.load(std::memory_order::relaxed)) {
```
r1viollet, line 83: Nice making this more correct.

src/perf_mainloop.cc
```diff
@@ -150,57 +154,133 @@ ReplyMessage create_reply_message(const DDProfContext &ctx) {
...
+  for (size_t i = 0; i < pes.size(); ++i) {
     // NOTE: if fd is negative, it will be ignored
     pfd[i].fd = pes[i].fd;
     pfd[i].events = POLLIN;
   }
 }
 
-DDRes signalfd_setup(pollfd *pfd) {
-  sigset_t mask;
+struct EventWrapper {
```
r1viollet, line 166: Comment proposal
```
This structure holds a reference to the actual events with the associated timestamp.
This allows us to order the events without copying the underlying memory.
```

r1viollet:
I am finding some crashes for now with full host.

nsavoire:
> I am finding some crashes for now with full host.

Pushed a fix

r1viollet:
Build is broken. 
This is great. I could not find overhead from the sorting process and crashes are fixed.

r1viollet (dismissed):
LGTM

PR #388, opened 2024-03-18
Fix live heap profiling when allocation profiling is also enabled

nsavoire:
# What does this PR do?

When both allocation and live heap profiling were enabled, tracking of deallocations was not requested not the library 😞


r1viollet (approved):
LGTM

PR #389, opened 2024-03-18
[PROF-9408] Fix pid mode to attach to all process threads

nsavoire:
# What does this PR do?

In pid mode, attach to all process threads by doing a perf_event_open for each (tid,cpu) combination. For a given cpu, only fd from the first tid will be mmapped, output for fds from other tids will be redirected to the first one.

### DIFF
--- a/include/pevent.hpp
+++ b/include/pevent.hpp
@@ -26,6 +26,9 @@ struct PEvent {
   bool custom_event; // true if custom event (not handled by perf, eg. memory
                      // allocations)
   RingBuffer rb;     // metadata and buffers for processing perf ringbuffer
+  std::vector<int>
+      sub_fds; // perf FDs of other events outputting to the same ring buffer
+               // (eg. perf events for other process threads in PID mode)
 };
 
 struct PEventHdr {
--- a/src/ddprof.cc
+++ b/src/ddprof.cc
@@ -5,13 +5,6 @@
 
 #include "ddprof.hpp"
 
-#include <cerrno>
-
-#include <csignal>
-#include <cstdio>
-#include <sys/resource.h>
-#include <unistd.h>
-
 #include "cap_display.hpp"
 #include "ddprof_cmdline.hpp"
 #include "ddprof_context.hpp"
@@ -28,6 +21,15 @@
 #ifdef __GLIBC__
 #  include <execinfo.h>
 #endif
+
+#include <absl/strings/numbers.h>
+#include <absl/strings/str_format.h>
+#include <cerrno>
+#include <csignal>
+#include <cstdio>
+#include <filesystem>
+#include <sys/resource.h>
+#include <unistd.h>
 
 namespace ddprof {
 namespace {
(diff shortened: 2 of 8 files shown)


include/pevent.hpp
```diff
@@ -26,6 +26,7 @@ struct PEvent {
   bool custom_event; // true if custom event (not handled by perf, eg. memory
                      // allocations)
   RingBuffer rb;     // metadata and buffers for processing perf ringbuffer
+  std::vector<int> sub_fds;
```
r1viollet, line 29: Proposed comment (it might need some improvement):
```
Associated perf event file descriptors that will output to the same buffer
```

src/ddprof.cc
```diff
@@ -65,6 +67,27 @@ void display_system_info() {
     LG_WRN("Unable to access perf_event_paranoid setting");
   }
 }
+
+DDRes get_process_threads(pid_t pid, std::vector<pid_t> &threads) {
+  threads.clear();
+
+  // Open the /proc/pid/task directory
+  std::string task_dir = absl::StrFormat("/proc/%d/task", pid);
```
r1viollet, line 75: Associated issues from CI
```
/go/src/github.com/DataDog/apm-reliability/ddprof-build/ddprof/src/ddprof.cc:75:3: error: variable 'task_dir' of type 'std::string' (aka 'basic_string<char>') can be declared 'const' [misc-const-correctness,-warnings-as-errors]
   75 |   std::string task_dir = absl::StrFormat("/proc/%d/task", pid);
      |   ^
      |               const 
/go/src/github.com/DataDog/apm-reliability/ddprof-build/ddprof/src/ddprof.cc:81:5: error: ignoring return value of function declared with 'warn_unused_result' attribute [clang-diagnostic-unused-result,-warnings-as-errors]
   81 |     absl::SimpleAtoi(dir_entry.path().filename().string(), &val);
      |     ^~~~~~~~~~~~~~~~ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
117094 warnings generated.
```

src/ddprof.cc
```diff
@@ -65,6 +67,31 @@ void display_system_info() {
     LG_WRN("Unable to access perf_event_paranoid setting");
   }
 }
+
+DDRes get_process_threads(pid_t pid, std::vector<pid_t> &threads) {
+  threads.clear();
+
+  // Open the /proc/pid/task directory
+  const std::string task_dir = absl::StrFormat("/proc/%d/task", pid);
```
r1viollet, line 75: The path to /proc lies in different places
I think it is sometimes mounted in `/host`. I added some (imperfect) logic around this for full host.

r1viollet:
Only one minor concern around the location of `/proc` LGTM otherwise!

r1viollet (approved):
LGTM

src/ddprof.cc
r1viollet: We talked about this.
The theory is that if the PID needs to be consistent with our PID namespace. Hence we think that checking in alternative paths is not relevant. We can probably do some associated cleanup for this logic.

r1viollet (approved):
LGTM

PR #392, opened 2024-03-25
[PROF-9088] Force backpopulate upon worker start in PID mode

nsavoire:
# What does this PR do?

Force backpopulate upon worker start in PID mode.

# Motivation

In PID mode (or in library mode, since it is implemented with PID mode), profiler will not receive mmap events since process is already started when profiler starts. This can cause issues for short lived processes that may have already  exited when profiler will attempt a backpopulate upon receiving first event.
By forcing a backpopulate upon start, we ensure that profiler has mapping information as soon as possible.

### DIFF
--- a/include/ddprof_context.hpp
+++ b/include/ddprof_context.hpp
@@ -45,6 +45,7 @@ struct DDProfContext {
   } params;
 
   ddprof::UniqueFd socket_fd;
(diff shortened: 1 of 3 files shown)


r1viollet (approved):
LGTM

r1viollet:
Thanks for addressing this so fast :bow:

PR #393, opened 2024-03-26
[PROF-9429] Search profiler mount namespace first

nsavoire:
# What does this PR do?

When looking for file corresponding to a mapping, first look in profiler mount namespace and if absent then look in `/proc/<pid>/root`.
In the case of short lived processes, path in profiler mount namespace will still be accessible after process exit.


r1viollet (approved):
LGTM
I agree! stable paths > unstable paths

PR #394, opened 2024-03-27
Use binary path for filename when symbolization fails

nsavoire:
# What does this PR do?

Do not leave function filename empty when symbolization fails.


r1viollet (approved):
LGTM

PR #395, opened 2024-03-27
[PROF-9088] Fix symbolization for binaries with compressed debug information

r1viollet:
# What does this PR do?

Ensure we retry without using debug information when symbolization fails

# Motivation

Ensure we symbolize Go binaries

# Additional Notes

This is pending an upstream fix in blazesym

# How to test the change?

The change is tested through simple_malloc


include/symbolizer.hpp
```diff
@@ -103,24 +103,34 @@ class Symbolizer {
...
                                    .auto_reload = false,
                                    .code_info = true,
                                    .inlined_fns = inlined_fns,
                                    .demangle = false,
                                    .reserved = {}};
     }
-
-    explicit BlazeSymbolizerWrapper(std::string elf_src, bool inlined_fns)
-        : opts(create_opts(inlined_fns)),
+    static blaze_symbolizer_opts create_opts_no_debug() {
```
nsavoire, line 114: I don't think you need to change `blaze_symbolizer_opts`, calling `blaze_symbolize_elf_virt_offsets` with `blaze_symbolize_src_elf.debug_syms` set to false should be enough.

include/symbolizer.hpp
```diff
@@ -103,24 +103,34 @@ class Symbolizer {
...
+      return blaze_symbolizer_opts{.type_size = sizeof(blaze_symbolizer_opts),
+                                   .auto_reload = false,
+                                   .code_info = false,
+                                   .inlined_fns = false,
+                                   .demangle = false,
+                                   .reserved = {}};
+    }
+    explicit BlazeSymbolizerWrapper(std::string elf_src, bool inlined_fns,
```
nsavoire, line 122: consecutive bool arguments are error prone

include/symbolizer.hpp
```diff
@@ -103,24 +103,34 @@ class Symbolizer {
...
+                         : create_opts_no_debug()),
           _symbolizer(blaze_symbolizer_new_opts(&opts)),
-          _elf_src(std::move(elf_src)) {}
+          _elf_src(std::move(elf_src)), _use_debug(use_debug) {}
     blaze_symbolizer_opts opts;
     std::unique_ptr<blaze_symbolizer, BlazeSymbolizerDeleter> _symbolizer;
     ddprof::HeterogeneousLookupStringMap<std::string> _demangled_names;
     std::string _elf_src;
     bool _visited{true};
+    bool _use_debug;
```
nsavoire, line 133: public members don't need an underscore prefix

src/symbolizer.cc
```diff
@@ -54,29 +54,34 @@ DDRes Symbolizer::symbolize_pprof(std::span<ElfAddress_t> elf_addrs,
...
-    DDPROF_DCHECK_FATAL(pair.second, "Unable to insert symbolizer object");
-    symbolizer = pair.first->second._symbolizer.get();
-    demangled_names = &(pair.first->second._demangled_names);
-  }
+  bool retry_symbolization = true;
+  bool use_debug = true;
   const blaze_result *blaze_res = nullptr;
-  if (!_disable_symbolization) {
+  while (retry_symbolization && !_disable_symbolization) {
+    const auto it = _symbolizer_map.find(file_id);
```
nsavoire, line 61: the inner loop could be refactored into another function

nsavoire (approved):

r1viollet:
@nsavoire I 100% agree with your comments, I will fix this in the evening. I'll merge just to get new numbers and look out for other issues.

PR #396, opened 2024-03-29
Simplify symbolize_pprof

nsavoire:
# What does this PR do?

Simplify a bit `symbolize_pprof` function.

### DIFF
--- a/src/symbolizer.cc
+++ b/src/symbolizer.cc
@@ -25,7 +25,7 @@ int Symbolizer::remove_unvisited() {
   // Remove all unvisited blaze_symbolizer instances from the map
   const auto count = std::erase_if(_symbolizer_map, [](const auto &item) {
     auto const &[key, blaze_symbolizer_wrapper] = item;
-    return !blaze_symbolizer_wrapper._visited;
+    return !blaze_symbolizer_wrapper.visited;
   });
   return count;
 }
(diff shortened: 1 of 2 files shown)


src/symbolizer.cc
```diff
@@ -45,49 +59,38 @@ DDRes Symbolizer::symbolize_pprof(std::span<ElfAddress_t> elf_addrs,
...
     // Symbolize the addresses
-    blaze_res = blaze_symbolize_elf_virt_offsets(
-        symbolizer, &src_elf, elf_addrs.data(), elf_addrs.size());
+    const auto *blaze_res = blaze_symbolize_elf_virt_offsets(
+        symbolizer_wrapper.symbolizer.get(), &src_elf, elf_addrs.data(),
+        elf_addrs.size());
+    if (!blaze_res && symbolizer_wrapper.use_debug) {
+      // Symbolization failed, retry without using debug symbols
+      // blazesym curently does not support compressed debug sections:
+      // cf. https://github.com/libbpf/blazesym/issues/581
```
r1viollet, line 85: ah yeah, nice adding the reference

src/symbolizer.cc
```diff
@@ -45,49 +59,38 @@ DDRes Symbolizer::symbolize_pprof(std::span<ElfAddress_t> elf_addrs,
...
         .reserved = {},
     };
 
     // Symbolize the addresses
-    blaze_res = blaze_symbolize_elf_virt_offsets(
-        symbolizer, &src_elf, elf_addrs.data(), elf_addrs.size());
+    const auto *blaze_res = blaze_symbolize_elf_virt_offsets(
+        symbolizer_wrapper.symbolizer.get(), &src_elf, elf_addrs.data(),
+        elf_addrs.size());
+    if (!blaze_res && symbolizer_wrapper.use_debug) {
```
r1viollet, line 82: Do you want to change to a retry at every symbolization attempt ?

r1viollet (approved):
LGTM

src/symbolizer.cc
nsavoire: Hum, it should only retry on the first symbolization attempt:
* on first attempt `symbolizer_wrapper.use_debug` is `true`, so symbolizer is called with debug symbols
* if it fails, `symbolizer_wrapper.use_debug` is set to false and symbolizer is called again without debug symbols
* on next attempt, `symbolizer_wrapper.use_debug` is false, so symbolizer is called without debug symbols and no other attempt is made is it fails

r1viollet: That makes sense :+1:

PR #398, opened 2024-03-29
Remove code locations by default

r1viollet:
# What does this PR do?

There seems to be an excessive RSS usage with the usage of debug information. This is an experiment to see what is the cost of debug information with blazesym.

# Motivation

Understand the cost of debug information.

# Additional Notes

The Inlinie feature still allows to access code location feature.

# How to test the change?

The aim is to see what this yields in staging applications.


nsavoire (approved):
LGTM

PR #399, opened 2024-04-03
Consider urls starting with a / to be UDS

r1viollet:
# What does this PR do?

Consider urls starting with a / to be UDS

# Motivation

Feedback from @scottopell 

# Additional Notes

NA

# How to test the change?

I added a unit test case to test the logic.

### DIFF
--- a/src/exporter/ddprof_exporter.cc
+++ b/src/exporter/ddprof_exporter.cc
@@ -213,6 +213,9 @@ DDRes ddprof_exporter_init(const ExporterInput &exporter_input,
       // check if schema is already available
       if (strstr(exporter_input.url.c_str(), "://") != nullptr) {
         exporter->_url = alloc_url_agent("", exporter_input.url, port_str);
+      } else if (exporter_input.url[0] == '/') {
+        // Starts with a '/', assume unix domain socket
+        exporter->_url = alloc_url_agent("unix://", exporter_input.url, {});
       } else {
         // not available, assume http
         exporter->_url =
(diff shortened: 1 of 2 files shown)


src/exporter/ddprof_exporter.cc
```diff
@@ -273,7 +276,8 @@ DDRes ddprof_exporter_new(const UserTags *user_tags, DDProfExporter *exporter) {
     exporter->_exporter = res_exporter.ok;
   } else {
     defer { ddog_Error_drop(&res_exporter.err); };
-    DDRES_RETURN_ERROR_LOG(DD_WHAT_EXPORTER, "Failure creating exporter - %s",
+    DDRES_RETURN_ERROR_LOG(DD_WHAT_EXPORTER, "Failure creating exporter - %.*s",
```
r1viollet, line 279: there was also an issue with the strings not being null terminated here

nsavoire (dismissed):

src/exporter/ddprof_exporter.cc
```diff
@@ -273,7 +276,8 @@ DDRes ddprof_exporter_new(const UserTags *user_tags, DDProfExporter *exporter) {
...
+                           static_cast<long int>(res_exporter.err.message.len),
```
nsavoire, line 280:
```suggestion
                           static_cast<int>(res_exporter.err.message.len),
```

r1viollet: indeed. fixed

nsavoire (approved):

PR #411, opened 2024-05-29
Add crash tracking to profiler

nsavoire:
# What does this PR do?

Add crash tracker and reporter to ddprof.

Crash tracker installs signal handlers to catch crashes in the profiler.
Upon crash, it forks and execs another instance of ddprof passing it `--report-crash` argument as well as crashed thread tid as `--pid` and then waits for this other process to finish before re-raising the signal.

Other ddprof process does a ptrace attach on its parent, unwind crashed thread and reports the crash through libdatadog crash tracker.


r1viollet (approved):
LGTM!
Looking forward to trying this out

src/dwfl_thread_callbacks.cc
```diff
@@ -51,7 +51,11 @@ bool set_initial_registers(Dwfl_Thread *thread, void *arg) {
 bool memory_read_dwfl(Dwfl *dwfl, Dwarf_Addr addr, Dwarf_Word *result,
                       int regno, void *arg) {
   (void)dwfl;
-  return memory_read(addr, result, regno, arg);
+  auto *us = static_cast<UnwindState *>(arg);
```
r1viollet, line 54: why would we not have 2 separate unwinding configurations ?

r1viollet:
@nsavoire moving this to draft

PR #417, opened 2024-07-16
Timeline ON by default

r1viollet:
# What does this PR do?

- Adjust timeline as ON by default
- bump version to 0.18.0

# Motivation

Ensure users discover the timeline feature

# Additional Notes

We should monitor the increase in memory usage in relenv

# How to test the change?

Prof-correctness can have some timestamp tests

### DIFF
--- a/src/ddprof_cli.cc
+++ b/src/ddprof_cli.cc
@@ -184,14 +184,18 @@ int DDProfCLI::parse(int argc, const char *argv[]) {
                  inlined_functions,
                  "Report inlined functions in call stacks.\n"
                  "This is possible if debug sections are available.\n"
-                 "This can have performance impacts for the profiler.")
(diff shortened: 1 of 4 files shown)


src/ddprof_cli.cc
```diff
@@ -188,10 +188,12 @@ int DDProfCLI::parse(int argc, const char *argv[]) {
       ->group("Profiling settings")
       ->default_val(false)
       ->envname("DD_PROFILING_INLINED_FUNCTIONS");
-  app.add_flag("--timeline,-t", timeline,
+
+  app.add_flag("--timeline,!--no-timeline", timeline,
```
sanchda, line 192: I'm stunned.  This is an incredibly expressive library!

sanchda (dismissed):
LGTM!

nsavoire (approved):

PR #419, opened 2024-07-17
Shutdown worker server only when all opened connections are closed

nsavoire:
# What does this PR do?

Finish handling all opened connections before shutting down worker server.

# Motivation

This fixes an issue when profiling library opens a connection to the profiler just before a profiler worker restart:
* library opens connection
* worker accepts connection
* worker initiates shutdown
* worker sends termination signal to server thread
* server threads closes all connections and terminates
* library loses connection and fails to get profiler information


r1viollet (approved):
Well done figuring this one out. :bow: 
LGTM

PR #420, opened 2024-07-26
Ensure that logger context name is used instead of MYNAME

nsavoire:
# What does this PR do?

Ensure that `LOG_setname` effectively changes name on log lines.

# Motivation

Before this change, setting name with `LOG_setname` had no effect because all log macros were passing `MYNAME` as name to log function, that would override name set with `LOG_setname`.

### DIFF
--- a/include/logger.hpp
+++ b/include/logger.hpp
@@ -78,18 +78,15 @@ bool LOG_open(int mode, const char *opts);
 
 // Formatted print to the ddprof logging facility
 // Log-print-Formatted with Level, Facility, and Name
-printflike(4, 5) void lprintfln(int lvl, int fac, const char *name,
-                                const char *fmt, ...);
+printflike(3, 4) void lprintfln(int lvl, int fac, const char *fmt, ...);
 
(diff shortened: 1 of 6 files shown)


include/logger.hpp
```diff
@@ -118,7 +118,7 @@ void LOG_set_logs_allowed_function(LogsAllowedCallback logs_allowed_function);
 #define LG_IF_LVL_OK(level, ...)                                               \
   do {                                                                         \
     if (unlikely(LOG_is_logging_enabled_for_level(level))) {                   \
-      olprintfln(ABS(level), -1, MYNAME, __VA_ARGS__);                         \
+      olprintfln(ABS(level), -1, nullptr, __VA_ARGS__);                        \
```
r1viollet, line 121: Minor:
It is slightly strange to put a nullptr here.
I understand that we default to the context name

r1viollet (dismissed):
LGTM
Please update description before merging :bow:

r1viollet (approved):
LGTM

PR #427, opened 2024-10-04
Timeline thread names

r1viollet:
# What does this PR do?

Ensure we have thread names / process name in the timeline

# Motivation

This makes the timeline easier to read

# Additional Notes

This has some amount of overhead
I left the timeline trigger to allow the fallback
There is no new lookup if we have a thread name change. 

# How to test the change?

Using timeline you should know see thread names.

### DIFF
--- a/src/unwind.cc
+++ b/src/unwind.cc
@@ -41,7 +41,7 @@ void add_container_id(Process &process, UnwindState *us) {
   }
 }
 
-void add_exe_name(Process &process, UnwindState *us) {
+void add_exe_name(UnwindState *us) {
   us->output.exe_name =
       us->symbol_hdr._base_frame_symbol_lookup.get_exe_name(us->pid);
 }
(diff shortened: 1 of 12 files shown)


src/unwind.cc
```diff
@@ -36,11 +35,20 @@ void find_dso_add_error_frame(DDRes ddres, UnwindState *us) {
 }
 
 void add_container_id(Process &process, UnwindState *us) {
-  auto container_id = process.get_container_id();
+  const auto &container_id = process.get_container_id();
```
r1viollet, line 38: This is slightly controversial, though the lifetime of the container ID should be tied to the process, which will outlive the aggregation of the sample

nsavoire (dismissed):

nsavoire:
Build is failing

r1viollet:
> Build is failing

Thanks, I think we are fine now.

nsavoire (approved):

PR #442, opened 2025-07-31
Octo policy: enable automatic releasing

r1viollet:
# What does this PR do?

Define octo policies to ensure we can manage releasing automatically. The associated build logics can be seen in ddprof-build.

# Motivation

Fix the releasing process.
Remove the usage of Personal Access Tokens.

# Additional Notes

NA

# How to test the change?

This can be tested from ddprof-build


r1viollet:
There is a better way to do this, we can hide these details. I'll close this for now.

PR #443, opened 2025-08-01
Update build dependencies

r1viollet:
# What does this PR do?

Update build dependencies.

# Motivation

Ensure builds are done with latest toolchains and latest versions of elfutils.

# Additional Notes

This requires a PR in ddprof-build to update build caches.

# How to test the change?

Already tested

### DIFF
--- a/app/base-env-alpine/Dockerfile
+++ b/app/base-env-alpine/Dockerfile
@@ -1,4 +1,4 @@
-FROM alpine:3.19
+FROM alpine:3.22
 ENV CC=gcc
 ENV CXX=g++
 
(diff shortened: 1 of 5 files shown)


src/lib/elfutils.cc
```diff
@@ -233,13 +233,15 @@ class SymbolLookup {
 
   bool operator()(std::string_view object_name, const DynamicInfo &dyn_info) {
     const ElfW(Sym) *s = nullptr;
+    // NOLINTBEGIN(bugprone-suspicious-stringview-data-usage)
     if (dyn_info.gnu_hash) {
       s = gnu_hash_lookup(dyn_info.strtab.data(), dyn_info.symtab.data(),
```
r1viollet, line 238: I'm not sure if we could fix this in a cleaner way

nsavoire (approved):

PR #453, opened 2025-10-21
Allocation tracker: fix the live heap tracking

r1viollet:
# What does this PR do?

Replace the bitset data structure with something that will not suffer from collisions.
Please refer to the `allocation_tracker.md` file for up to date notes.

# Motivation

Ensure that allocation tracking is correct.

# Additional Notes

NA

# How to test the change?

There are benchmarks in this PR. They are mostly generated through AI. I do have some doubts on some of the results.
I did some manual testing to check that overhead seems reasonable.
Next step is end to end testing on real workloads.

### DIFF
--- a/docs/allocation_tracker.md
+++ b/docs/allocation_tracker.md
@@ -1,6 +1,7 @@
 # AddressBitset Evolution: Sharded Hash Tables
 
-This document chronicles the evolution of the `AddressBitset` data structure used for live allocation tracking in ddprof, from a collision-prone bitset to a sharded hash table. It includes production benchmark results comparing the final implementation against alternatives.
+The aim is to have bookkeeping on the in-use addresses for live heap profiling.
+This document talks about how we landed on a sharded hash map and alternatives that we could be exploring.
 
 ## The Problem: Bitset Collisions
 
@@ -16,18 +17,6 @@ _bitset[bit_index / 64] |= (1ULL << (bit_index % 64));
 - Two different addresses hash to same bit: collision
 - When freeing, we can't distinguish which address the bit represents
 - Result: Inaccurate leak detection, missed deallocations
-
-Tests showed significant impact:
- - 6% loss rate
-
-## Solution Requirements
-
-We need a data structure that:
-1. **No collisions** - track actual addresses, not just bits
-2. **Thread-safe** - signal-safe, atomic operations only
-3. **Fast lookups** - 100× more reads (free) than writes (malloc)
-4. **Bounded memory** - fixed size or minimal growth
-5. **Production-ready** - handle millions of allocations
 
 ## Explored Alternatives
 
@@ -122,11 +111,6 @@ Glibc malloc spaces thread arenas
 
 While not suitable as the primary solution due to sampling bias, stateless sampling could be offered as an **optional high-performance mode**:
 
-**Use cases for sampling mode**:
-- **Cheap leak detection** - quick scan for memory leaks with minimal overhead
-- **High-frequency profiling** - when sampling overhead must be minimal
-- **Complementary tool** - run alongside full tracking for different insights
-
 **Benchmark results** (stateless sampling):
 ```
 BM_AddressSampler_SingleThreaded                 7.11 ns    items_per_second=140.669M/s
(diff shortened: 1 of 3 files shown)


include/lib/address_bitset.hpp
```diff
@@ -5,67 +5,130 @@
...
+  unsigned table_mask;
+  unsigned
+      max_capacity; // max_capacity = table_size * kMaxLoadFactorPercent / 100
+  std::unique_ptr<std::atomic<uintptr_t>[]> slots;
+  std::atomic<int> count{0};
+
+  explicit AddressTable(unsigned size);
+  ~AddressTable() = default;
+
+  // Delete copy/move operations (non-copyable due to atomic members)
```
KowalskiThomas: This is saying non-copyable, should it be movable? I see those constructors are `= delete` as well

src/lib/address_bitset.cc
```diff
@@ -41,80 +49,212 @@ AddressBitset &AddressBitset::operator=(AddressBitset &&other) noexcept {
   return *this;
 }
 
+AddressBitset::~AddressBitset() {
+  // This should not run unless we are sure that we are no longer running the
+  // allocation tracking. This is not enough of a synchronization
```
KowalskiThomas: Does this need to be fixed in this PR?

src/lib/address_bitset.cc
```diff
@@ -30,6 +27,17 @@ unsigned round_up_to_power_of_two(unsigned num) {
 }
 } // namespace
 
+// AddressTable implementation
+AddressTable::AddressTable(unsigned size)
+    : table_size(round_up_to_power_of_two(size)), table_mask(table_size - 1),
+      max_capacity(table_size * kMaxLoadFactorPercent / kPercentDivisor),
+      slots(std::make_unique<std::atomic<uintptr_t>[]>(table_size)) {
+  // Initialize all slots to empty
+  for (unsigned i = 0; i < table_size; ++i) {
```
KowalskiThomas: **question** any reasons to use `unsigned` over  `size_t` here for `i`?

src/lib/address_bitset.cc
```diff
@@ -41,80 +49,212 @@ AddressBitset &AddressBitset::operator=(AddressBitset &&other) noexcept {
...
+      current = table->slots[slot].load(std::memory_order_acquire);
+    }
+
+    // Check if slot already contains our address
+    if (current == addr) {
+      return false; // Already tracked
+    }
+
+    // Slot occupied by different address - probe next slot
+    slot = (slot + 1) & table->table_mask;
```
KowalskiThomas: To confirm, this is valid because the size of the table is a power of two, correct?

docs/allocation_tracker.md
```diff
@@ -0,0 +1,254 @@
...
+
+### Benefits
+
+**No collisions** - stores actual addresses with linear probing  
+**Thread separation** - different threads use different tables (not always true...)
+**Lazy allocation** - only allocate tables for active chunks  
+**Bounded memory** - max 128 chunks × 32K slots × 8 bytes = 32 MB worst case (typically much less)
+**Fast lookups** - O(1) with minimal probing
+**Signal-safe** - atomic operations only after initialization
+**Hash-based sharding** - distributes addresses across 128 shards using hash function
```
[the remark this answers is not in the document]
r1viollet: I do want to keep some of this for students & to pick things up next time. I'll clean up the doc.

include/lib/address_bitset.hpp
r1viollet: We can add these if needed. I don't think we ever move stuff in current implementation.

src/lib/address_bitset.cc
```diff
@@ -41,80 +49,212 @@ AddressBitset &AddressBitset::operator=(AddressBitset &&other) noexcept {
...
+  // Check if table is at max capacity (60% load factor)
+  if (table->count.load(std::memory_order_relaxed) >=
+      static_cast<int>(table->max_capacity)) {
+    return false; // Table is full
+  }
+
+  uint32_t slot = hash_to_slot(hash, table->table_mask);
+
+  // Linear probing to find an empty/deleted slot or the address
+  for (unsigned probe = 0; probe < AddressTable::kMaxProbeDistance; ++probe) {
```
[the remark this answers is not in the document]
r1viollet: Yes, this was a pretty good suggestion. I will get back to testing this. Though I did want to secure a first fix.

src/lib/address_bitset.cc
r1viollet: Overall, I think we are OK with the static lifetime. Feel free to disagree.

PR #468, opened 2025-11-05
Revert usage of hash to select the shard

r1viollet:
# What does this PR do?

- In commit 6bfad54 I made a mistake where I broke the sharding logic
- Minor cleanup on unused variables
- Increase probing to 64 & load factor
The assumption is that users care more about invesitgating leaks. This is already not great in perf.

# Motivation

Go back to putting addresses in close addresses in the same shard as this could be causing issues in memory usage.

# Additional Notes

NA

# How to test the change?

Once this is merged in main, I expect relenv deployments will be healthy.

### DIFF
--- a/test/simple_malloc.cc
+++ b/test/simple_malloc.cc
@@ -11,10 +11,10 @@
 #include <cstdlib>
 #include <cstring>
 #include <ctime>
+#include <deque>
 #include <dlfcn.h>
 #include <functional>
 #include <iostream>
-#include <sstream>
 #include <sys/resource.h>
 #include <thread>
 #include <unistd.h>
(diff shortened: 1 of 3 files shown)


test/simple_malloc.cc
```diff
@@ -126,9 +128,19 @@ extern "C" DDPROF_NOINLINE void do_lot_of_allocations(const Options &options,
...
+    if (options.keep_live_allocations > 0) {
+      if (p2 != nullptr) {
+        live_allocations.push_back(p2);
+        while (live_allocations.size() > options.keep_live_allocations) {
+          free(live_allocations.front());
+          live_allocations.pop_front();
+        }
+      }
+    } else {
+      if (skip_free++ >= options.skip_free) {
```
r1viollet: maybe I remove the skip free ? I don't like skip free it ends up killing my laptop :smile:

nsavoire: skip free is useful to check that we are able to detect leaks

PR #475, opened 2025-12-03
Arm64 fixes

r1viollet:
# What does this PR do?

- Upgrade elfutils to 0.194
- Add some support to ensure we can work with virtiofs paths
```
(/run/host_virtiofs/Users/erwan.viollet/go/src/github.com/DataDog/ddprof/build_gcc_unknown-linux-2.39_Deb/test/simple_malloc)(T-Standard)(r-x)(ID#-1)
```
- Add a heuristic to get past vdso in arm64

# Motivation

Working from MacOS aarch64, I was not able to get proper frames.
With these changes, I no longer have failures on Simple Malloc tests.

# Additional Notes

This will require a branch in `ddprof-build` to cache elfutils:
https://github.com/DataDog/ddprof-build/pull/152

I would rather we first merge the fixes on mirroring before we consider these changes.

# How to test the change?

### DIFF
--- a/tools/launch_local_build.sh
+++ b/tools/launch_local_build.sh
@@ -88,6 +88,15 @@ if [ -e "${CURRENTDIR}/.env" ]; then
 fi
 
 MOUNT_CMD="-v ${DEFAULT_DEV_WORKSPACE:-${CURRENTDIR}}:/app"
+# avoid symlinks
+HOST_SHARE_PATH=$(python3 - <<'PY'
+import os, sys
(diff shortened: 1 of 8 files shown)


tools/launch_local_build.sh
```diff
@@ -88,6 +88,15 @@ if [ -e "${CURRENTDIR}/.env" ]; then
 fi
 
 MOUNT_CMD="-v ${DEFAULT_DEV_WORKSPACE:-${CURRENTDIR}}:/app"
+# avoid symlinks
```
r1viollet: slightly hacky, though not sure if this can be done in shell

PR #478, opened 2025-12-04
Unwinding fix - remove heuristic around infinite loop detection

r1viollet:
# What does this PR do?

Heuristic was triggering in the wrong cases (recursive functions).

# Motivation

Fix call stacks on some arm test cases.

# Additional Notes

NA

# How to test the change?

The bad-boggle-solver is a good test for this.

### DIFF
--- a/src/unwind_dwfl.cc
+++ b/src/unwind_dwfl.cc
@@ -42,23 +42,6 @@ void trace_unwinding_end(UnwindState *us) {
   }
 }
 
-bool is_infinite_loop(UnwindState *us) {
-  UnwindOutput &output = us->output;
-  uint64_t const nb_locs = output.locs.size();
-  unsigned const nb_frames_to_check = 3;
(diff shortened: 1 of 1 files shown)


src/unwind_dwfl.cc
```diff
@@ -184,16 +167,9 @@ int frame_cb(Dwfl_Frame *dwfl_frame, void *arg) {
...
-  if (dwfl_error_value) {
-    // Check if dwarf unwinding was a failure we can get stuck in infinite loops
-    if (is_infinite_loop(us)) {
-      LG_DBG("Break out of unwinding (possible infinite loop)");
-      return DWARF_CB_ABORT;
-    }
-  }
 #ifdef DEBUG
   // We often fallback to frame pointer unwinding (which logs an error)
+  int const dwfl_error_value = dwfl_errno();
```
r1viollet: I have a minor doubt on how useful it is to read the errno value.

PR #479, opened 2025-12-09
Avoid matching symbols at address 0

r1viollet:
# What does this PR do?

Avoid matching symbols at address 0.

# Motivation

This will help our backend trigger remote symbolication calls
Ticket number: SMPTNG-723

# Additional Notes

NA

# How to test the change?

Before we would have a single DD_AGENT_1 symbol and we would not trigger remote symbolication

image


r1viollet:
For reference, matching fix in the agent is here: https://gitlab.ddbuild.io/DataDog/datadog-agent/-/commit/6975db2ae280023390c684f5e3fa2d4bebb92057

PR #483, opened 2026-01-14
Crash fix in case of forks: remove usage of pthread_once

r1viollet:
# What does this PR do?

Disclaimer: I do not fully understand why we have a crash where the key was not initialized.
However we know that in the case of forks we can be in a state where we try to init in the hot path (during an allocation).
This could be the reason for the crash.

This PR addresses the symptoms.

- We remove the init from the hot path
- We add a fork handler to make sure init of the key is OK
We could go one step further here, to make sure we wrap the full init sequence (not just thread local state)
- We use the key as synchronization mechanism

# Motivation

Avoid a crash.

# Additional Notes

NA

# How to test the change?

I added a test for forking. Though it does not reproduce the crash.

### DIFF
--- a/src/lib/allocation_tracker.cc
+++ b/src/lib/allocation_tracker.cc
@@ -48,6 +48,12 @@ DDPROF_NOINLINE auto sleep_and_retry_reserve(MPSCRingBufferWriter &writer,
   }
   return Buffer{};
 }
+
+void child_after_fork_handler() {
+  // After fork in child, verify the pthread key is still valid.
+  // The key itself survives fork, but we want to ensure it's properly set.
+  AllocationTracker::ensure_key_initialized();
+}
 } // namespace
 
 TrackerThreadLocalState *AllocationTracker::get_tl_state() {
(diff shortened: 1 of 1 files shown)


src/lib/allocation_tracker.cc
```diff
@@ -56,9 +54,16 @@ TrackerThreadLocalState *AllocationTracker::get_tl_state() {
...
-  pthread_once(&_key_once, make_key);
-  auto *tl_state = static_cast<TrackerThreadLocalState *>(
-      pthread_getspecific(_tl_state_key));
+
+  // The pthread key is initialized during allocation_tracking_init() and
+  // maintained by the fork handler, so we can directly use it here (hot path)
+  pthread_key_t key = _tl_state_key.load(std::memory_order_relaxed);
+  
+  // In debug builds, verify our assumption
+  assert(key != kInvalidKey && "pthread key should be initialized before use");
```
r1viollet: this might trigger with a relaxed order, but at least I will know about it
I might change mem order

r1viollet:
Ok, I'm convinced! Thanks for the proposal, I'll close this again to do a new refactor.

PR #485, opened 2026-01-29
[PROF-12579] Libdatadog String interning

r1viollet:
# What does this PR do?

Switches the pprof pipeline to libdatadog's `ProfilesDictionary` so strings, functions, and mappings are interned once instead of copied per sample.

- `Symbol` stores a `FunctionId2` handle; `MapInfoTable` is now `vector<MappingId2>` — no owned strings in either.
- All symbol/mapping lookups intern into the dictionary on first insert and return opaque handles. Subsequent lookups are cache hits at the ddprof level.
- Labels are interned once (key IDs at profile creation, values as `CharSlice` at sample time).
- Aggregation goes through `Profile_add2` on a `Profile_with_dictionary`-backed profile — no re-interning at sample time.
- `ProfilesDictionary` held by `SymbolHdr` (declared first, destroyed last, outlives all handles).
- Mapping cache (`MapInfoLookup`) detects remaps at the same start address by reading identity back from the dictionary.

# Motivation

Copying stack symbol strings on every sample was the wrong design. Interned IDs let pprof aggregation be pointer-sized work.

# How to test the change?

- `ddprof_pprof-ut` includes a new aggregate → reset → aggregate cycle exercising ID stability across `pprof_reset`.
- `mapinfo_lookup-ut` covers cache hits, remap detection, build-id mismatch, pid isolation, and erase behaviour.
- New `pprof_aggregate-bench` micro-benchmark: on aarch64 Release, mean per-sample aggregation drops **877 ns → 333 ns** (~2.6×).
- Existing `ddprof_exporter-ut` already destroys `SymbolHdr` before serialization — covers dictionary-outliving-samples under ASan.


nsavoire:
@codex review

include/mapinfo_table.hpp
```diff
@@ -26,6 +28,7 @@ class MapInfo {
   Offset_t _offset{0};
   std::string _sopath;
   BuildIdStr _build_id;
+  ddog_prof_Mapping2 *_mapping_id{nullptr};
```
nsavoire:
```suggestion
  ddog_prof_MappingId2 _mapping_id{nullptr};
```

nsavoire: `MapInfo ` carries both its full data and its interned form (contrary to `Symbol` which only carries its interned form).
Why not do something similar to `Symbol` or completely replace `MapInfoIdx_t` with `ddog_prof_MappingId2` ?

include/symbol_hdr.hpp
```diff
@@ -15,21 +15,81 @@
...
+      reset();
+      _handle = std::exchange(o._handle, nullptr);
+    }
+    return *this;
+  }
+
+  explicit operator bool() const { return _handle != nullptr; }
+
+  // The C API expects a pointer-to-handle (so it can mutate the slot).
+  [[nodiscard]] const ddog_prof_ProfilesDictionaryHandle *get() const {
```
nsavoire, line 47: Not sure if this is necessary, the only used  C API that requires a pointer-to-handle is `ddog_prof_Profile_with_dictionary` and it should not mutate the slot.

src/ddog_profiling_utils.cc
```diff
@@ -25,51 +28,146 @@ std::string_view get_or_insert_demangled_sym(
...
-void write_mapping(const MapInfo &mapinfo, ddog_prof_Mapping *ffi_mapping) {
-  ffi_mapping->memory_start = mapinfo._low_addr;
-  ffi_mapping->memory_limit = mapinfo._high_addr;
-  ffi_mapping->file_offset = mapinfo._offset;
-  ffi_mapping->filename = to_CharSlice(mapinfo._sopath);
-  ffi_mapping->build_id = to_CharSlice(mapinfo._build_id);
+Symbol make_symbol(std::string symname, const std::string &demangled_name,
+                   uint32_t lineno, const std::string &srcpath,
+                   const ddog_prof_ProfilesDictionary *dict) {
+  [[maybe_unused]] const std::string ignored_symname = std::move(symname);
```
nsavoire: `symname` is never used

r1viollet:
@codex review

r1viollet:
@codex review

src/ddprof_worker.cc
```diff
@@ -710,8 +710,12 @@ DDRes ddprof_worker_init(DDProfContext &ctx,
     DDRES_CHECK_FWD(
         ddprof_exporter_new(ctx.worker_ctx.user_tags, ctx.worker_ctx.exp[1]));
 
-    DDRES_CHECK_FWD(pprof_create_profile(ctx.worker_ctx.pprof[0], ctx));
-    DDRES_CHECK_FWD(pprof_create_profile(ctx.worker_ctx.pprof[1], ctx));
+    DDRES_CHECK_FWD(pprof_create_profile(
+        ctx.worker_ctx.pprof[0], ctx,
+        ctx.worker_ctx.us->symbol_hdr._profiles_dictionary.get()));
```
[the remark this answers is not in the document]
r1viollet: The key libdatadog header says ProfilesDictionary contains data common to multiple profiles, “whether that’s multiple profiles simultaneously or multiple profiles through time”, and that the current implementation is thread-safe. That matches our double-buffer setup: two Profiles sharing one long-lived SymbolHdr dictionary.

I think we are fine, though codex does not have access to the libdatadog APIs.

include/symbol_hdr.hpp
r1viollet: I think this comment was about the get, I did not really adjust this, tell me if you feel strongly. I might have missed why this was important.

r1viollet:
To answer a question, mappings are never cleared afaik, so we still partially rely on the hard worker reset.

src/ddprof_worker.cc
nsavoire, line 750: `worker_library_free` should be moved after `pprof_free_profile` since profiles hold a pointer on dictionary

src/pprof/ddprof_pprof.cc
```diff
@@ -33,6 +33,13 @@ namespace ddprof {
 namespace {
 constexpr size_t k_max_pprof_labels{8};
 
+constexpr std::string_view k_container_id_label = "container_id"sv;
+constexpr std::string_view k_process_id_label = "process_id"sv;
+constexpr std::string_view k_process_name_label = "process_name"sv;
+constexpr std::string_view k_thread_id_label = "thread id"sv;
```
nsavoire, line 41: nit: re-add comment:
```
// This naming has an impact on backend side (hence the inconsistency with
// process_id)
```

include/symbol_helper.hpp
```diff
@@ -40,7 +43,19 @@ std::vector<std::string> collect_symbols(UnwindState &state,
...
+        demangled_name = "unknown";
+      } else {
+        ddog_CharSlice slice{nullptr, 0};
+        ddog_prof_Status status = ddog_prof_ProfilesDictionary_get_str(
+            &slice, dict, symbol._function_id->name);
+        if (status.err != nullptr) {
+          ddog_prof_Status_drop(&status);
+          demangled_name = "unknown";
+        } else {
+          demangled_name = std::string(slice.ptr, slice.len);
```
nsavoire: nit: check if ptr is null

src/mapinfo_lookup.cc
```diff
@@ -5,30 +5,53 @@
...
+
+std::string_view basename(std::string_view path) {
+  size_t const pos = path.rfind('/');
+  return (pos == std::string_view::npos) ? path : path.substr(pos + 1);
+}
+
+} // namespace
+
+MapInfoIdx_t
+MapInfoLookup::get_or_insert(pid_t pid, MapInfoTable &mapinfo_table,
```
nsavoire, line 30: Did the behaviour change ?

src/symbol_hdr.cc
```diff
@@ -0,0 +1,34 @@
...
+
+namespace ddprof {
+
+ProfilesDictionary::ProfilesDictionary() {
+  ::ddog_prof_Status status = ::ddog_prof_ProfilesDictionary_new(&_handle);
+  if (status.err != nullptr) {
+    LG_WRN("Failed to create ProfilesDictionary for string interning: %s",
+           status.err);
+    ::ddog_prof_Status_drop(&status);
+    _handle = nullptr;
```
nsavoire: Bubble up failure

r1viollet: I might have to throw as this is a constructor

src/mapinfo_lookup.cc
r1viollet: Yes, it has changed:
```
- Same mapping at same start: same behavior, returns existing index.
- Different mapping at new start: same behavior, appends new row.
- Different mapping at same start: changed behavior. Old code reused stale row; new code appends a fresh row.
- Storage: old table stored owned MapInfo fields; new table stores ddog_prof_MappingId2 handles into libdatadog’s dictionary.
- Growth: old table grew only for new (pid, start) entries after cache erase/new start. New table can also grow for same-start remaps with changed end/offset/filename/build-id. That is the correctness tradeoff.
```
We do not really have APIs to mutate libdatadog objects and it is unclear if that would be valid as samples reference these objects.

r1viollet:
@nsavoire I would be in favour of merging and letting this one bake in staging environments to see how it behaves. What do you think ?

src/mapinfo_lookup.cc
nsavoire: Thanks for the clarification !

nsavoire (approved):

PR #531, opened 2026-03-31
fix: allow lazy TLS init for threads started before profiling

nsavoire:
# What does this PR do?

When profiling starts on a process with already-running threads calling pthread_getattr_np, the allocation tracker's TLS state may not be initialized yet. This adds a pthread_getattr_np hook that guards against deadlock withthe following sequence:
(pthread_getattr_np -> malloc -> tracker -> init_tl_state -> save_context -> pthread_getattr_np deadlock) and enables lazy TLS init in get_tl_state() when safe.

# Motivation

Allow allocation profiling of existing threads when profiler is not started at process startup.

### DIFF
--- a/src/lib/allocation_tracker.cc
+++ b/src/lib/allocation_tracker.cc
@@ -64,11 +64,23 @@ DDPROF_NOINLINE auto sleep_and_retry_reserve(MPSCRingBufferWriter &writer,
 }
 } // namespace
 
-TrackerThreadLocalState *AllocationTracker::get_tl_state() {
+TrackerThreadLocalState &AllocationTracker::get_tl_state_no_init() {
+  return *reinterpret_cast<TrackerThreadLocalState *>(ddprof_lib_state);
+}
+
+TrackerThreadLocalState *
+AllocationTracker::get_tl_state(bool init_if_not_initialized) {
   // ddprof_lib_state is zero-initialized by libc for each new thread.
   // After placement new (init_tl_state), initialized is set to true.
-  auto *state = reinterpret_cast<TrackerThreadLocalState *>(ddprof_lib_state);
-  return state->initialized ? state : nullptr;
+  auto &state = get_tl_state_no_init();
+  if (state.initialized) {
+    return &state;
+  }
+  if (init_if_not_initialized && !state.reentry_guard) {
+    // We are not inside pthread_getattr_np, so we can initialize the state
+    return init_tl_state();
+  }
+  return nullptr;
 }
 
 TrackerThreadLocalState *AllocationTracker::init_tl_state() {
@@ -560,12 +572,8 @@ AllocationTracker::next_sample_interval(std::minstd_rand &gen) const {
 void AllocationTracker::notify_thread_start() {
   TrackerThreadLocalState *tl_state = get_tl_state();
   if (unlikely(!tl_state)) {
-    tl_state = init_tl_state();
-    if (!tl_state) {
-      LG_DBG("Unable to start allocation profiling on thread %d",
-             ddprof::gettid());
-      return;
-    }
+    LG_DBG("Unable to start allocation profiling on thread %d",
+           ddprof::gettid());
   }
 }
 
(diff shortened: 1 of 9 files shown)


src/lib/allocation_tracker.cc
```diff
@@ -64,11 +64,22 @@ DDPROF_NOINLINE auto sleep_and_retry_reserve(MPSCRingBufferWriter &writer,
...
   // ddprof_lib_state is zero-initialized by libc for each new thread.
   // After placement new (init_tl_state), initialized is set to true.
-  auto *state = reinterpret_cast<TrackerThreadLocalState *>(ddprof_lib_state);
-  return state->initialized ? state : nullptr;
+  auto &state = get_tl_state_no_init();
+  if (state.initialized) {
+    return &state;
+  }
+  if (!state.initialized && !state.reentry_guard) {
+    // We are not inside pthread_getattr_np, so we can initialize the state
```
r1viollet, line 80: I think the `state.reentry_guard` captures also when we are in an allocation hook ?

r1viollet: although this should not happen as we should never enter without a state, though this is a shared reentry mechanism

r1viollet:
Nicolas had a smart idea, the scenario we are worried about is doing an init while we already have a lock (allocation) held. This can happens when you are doing an mmap within a malloc (especially as we can not hook all mallocs).
His proposal was to only perform the init on APIs where it is impossible for us to hold a lock (like inside a malloc hook).
This is a great idea!

src/lib/allocation_tracker.cc
nsavoire: I reused `state.reentry_guard` since we don't use it when state is not yet initialized and the purpose is similar

src/lib/allocation_tracker.cc
```diff
@@ -64,11 +64,23 @@ DDPROF_NOINLINE auto sleep_and_retry_reserve(MPSCRingBufferWriter &writer,
...
+AllocationTracker::get_tl_state(bool init_if_not_initialized) {
   // ddprof_lib_state is zero-initialized by libc for each new thread.
   // After placement new (init_tl_state), initialized is set to true.
-  auto *state = reinterpret_cast<TrackerThreadLocalState *>(ddprof_lib_state);
-  return state->initialized ? state : nullptr;
+  auto &state = get_tl_state_no_init();
+  if (state.initialized) {
+    return &state;
+  }
+  if (init_if_not_initialized && !state.reentry_guard) {
```
r1viollet, line 79: So if we request init and we can not do it, we return `null` (which I think makes sense)

src/lib/symbol_overrides.cc
```diff
@@ -71,26 +71,26 @@ class MaybeReentryGuard {
...
+template <bool mmap_alloc> class AllocTrackerHelperImpl {
 public:
-  AllocTrackerHelper()
-      : _tl_state{ddprof::AllocationTracker::get_tl_state()},
+  // Do not try to initialize the TLS state when intercepting
+  // mmap/mmap64/munmap/munmap_ because those functions might be called from
+  // malloc implementation and cause a deadlock when pthread_getattr_np (which
+  // can call malloc) is called by tls state initialization.
+  AllocTrackerHelperImpl()
+      : _tl_state{ddprof::AllocationTracker::get_tl_state(mmap_alloc)},
```
r1viollet: is the point not to have `get_tl_state(!mmap_alloc)`
I guess this is what codex is saying ?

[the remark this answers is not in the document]
r1viollet: good bot 😄

nsavoire: oups,indeed I inverted the condition

r1viollet: the fact that tests are still OK with this inverted is strange ? Or just lucky ?

nsavoire: that's because there is no test for "mmap hook called inside malloc without malloc hook being called"
I added one

test/allocation_tracker_fork_test.cc
r1viollet: nit, this requires false ? to avoid running init in child

r1viollet (dismissed):
LGTM

test/allocation_tracker_fork_test.cc
nsavoire: I changed all `get_tl_state()` to `get_tl_state(false)` to ensure that the behaviour remains the same in the test.

r1viollet (approved):
Thanks for adjusting tests.
LGTM

PR #533, opened 2026-04-01
fix: out-of-bounds access in `process_symbolization`

KowalskiThomas:
# What does this PR do?

Fixes an out-of-bounds array access in the `process_symbolization` function inside `ddprof_pprof.cc`.

In the inner loop that collects consecutive locations for the same file, `index` is incremented via `++index` before checking `locs[index].symbol_idx`. When the last element of `locs` is reached inside the loop, `index` becomes equal to `locs.size()` and the subsequent access to `locs[index]` is out of bounds, result in undefined behavior.


r1viollet (approved):
LGTM

PR #534, opened 2026-04-02
loader: replace version check with SHA-256 hash verification

nsavoire:
# What does this PR do?

Verify the installed libdd_profiling-embedded.so by computing its SHA-256 on disk and comparing against the build-time hash, before calling dlopen.  This avoids loading a library that cannot be safely unloaded on mismatch.

The build system now hashes the .so directly (instead of the .o wrapper), so the same hash serves as both the temp-file cache key and the installed-lib verification digest.

The ddprof_profiling_version symbol and the version-check infrastructure are removed — the hash check is strictly stronger and does not require dlopen.

SHA-256 implementation lives in src/lib/sha256.c.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -448,8 +448,17 @@ if(BUILD_UNIVERSAL_DDPROF)
   endif()
 endif()
 
+set(_install_targets ddprof dd_profiling-static dd_profiling-shared)
+if(USE_LOADER)
+  # Install the embedded profiling library so the loader can dlopen it from a
+  # standard path instead of extracting the copy baked into the loader to /tmp.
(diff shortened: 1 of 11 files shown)


r1viollet (dismissed):
LGTM
The sha256 is a better strategy, thanks

xroche:
> The sha256 is a better strategy, thanks

~~There is an issue though: the extracted embed library is stripped after computing the hash!~~

Edit: The issue was on our side (you can not strip the libs), sorry for the false positive!

xroche:
@r1viollet @nsavoire The latest RC appears to work well, we're looking forward for the release!

image

Issue #535, opened 2026-04-02
Loader always extracts to /tmp when USE_LOADER=ON: libdd_profiling-embedded.so not installed


xroche:
## Description

When `USE_LOADER=ON` (the default), the loader (`libdd_profiling.so`) tries to `dlopen("libdd_profiling-embedded.so")` at construction time (`loader.c:312`). This always fails because `libdd_profiling-embedded.so` is not included in the install targets or the release tarball — it only exists as binary data embedded inside the loader.

Since the dlopen fails, the loader extracts both `libdd_profiling-embedded.so` and `ddprof` to `/tmp`. The `/tmp` copy of `ddprof` lacks file capabilities (`CAP_PERFMON`, `CAP_SYS_PTRACE`, `CAP_IPC_LOCK`), so `perf_event_open` fails with `EPERM`.

This makes the "skip /tmp extraction when libraries are installed" fix from #529 / commit 90b41db5 ineffective in practice.

## Reproduction

Install ddprof 0.25.0 from the release tarball on any system:

```
$ ls /usr/lib/libdd_profiling*
/usr/lib/libdd_profiling.so        # the loader — installed
                                   # libdd_profiling-embedded.so — NOT installed

$ ls /tmp/ddprof-* /tmp/libdd_profiling-embedded*
/tmp/ddprof-<hash>                 # extracted, no file caps
/tmp/libdd_profiling-embedded.so-<hash>  # extracted
```

## Proposed fix

Add `dd_profiling-embedded` to the CMake install targets when `USE_LOADER=ON`:

```cmake
set(_install_targets ddprof dd_profiling-static dd_profiling-shared)
if(USE_LOADER)
  list(APPEND _install_targets dd_profiling-embedded)
endif()

install(
  TARGETS ${_install_targets}
  RUNTIME DESTINATION bin
  LIBRARY DESTINATION lib
  ARCHIVE DESTINATION lib
  PUBLIC_HEADER DESTINATION include)
```

With this change, the loader finds `libdd_profiling-embedded.so` via standard dlopen search, skips `/tmp` extraction entirely, and `execvp("ddprof")` finds `/usr/bin/ddprof` with its file capabilities intact.

xroche:
Fixed in https://github.com/DataDog/ddprof/pull/534

PR #536, opened 2026-04-02
Install libdd_profiling-embedded.so to fix /tmp extraction and CAP_PERFMON loss

xroche:
## Summary

When `USE_LOADER=ON`, the loader tries `dlopen("libdd_profiling-embedded.so")` before falling back to extracting the embedded copy. This dlopen always fails because `dd_profiling-embedded` is built but not included in the install targets.

As a result, the loader extracts both `libdd_profiling-embedded.so` and `ddprof` to `/tmp`. The extracted `ddprof` copy lacks file capabilities (`CAP_PERFMON`), so `perf_event_open` fails with `EPERM`.

## Fix

Add `dd_profiling-embedded` to the CMake install targets when `USE_LOADER=ON`. The loader finds `libdd_profiling-embedded.so` via standard dlopen, skips extraction, and `execvp("ddprof")` uses the installed binary with its file capabilities.

## Refs

- #535
- #529

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -441,8 +441,17 @@ if(BUILD_UNIVERSAL_DDPROF)
   endif()
 endif()
 
+set(_install_targets ddprof dd_profiling-static dd_profiling-shared)
+if(USE_LOADER)
+  # Install the embedded profiling library so the loader can dlopen it from a
+  # standard path instead of extracting the copy baked into the loader to /tmp.
+  # The /tmp fallback loses file capabilities (CAP_PERFMON) on the ddprof
+  # binary, which breaks perf_event_open.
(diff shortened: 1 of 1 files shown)


xroche:
cc @r1viollet — this fixes the /tmp extraction issue from #529. The embedded library was never installed, so the loader's bare-name dlopen always fell back to extracting to /tmp.

r1viollet:
Thanks for the fix, I'll handover the integration to @nsavoire as I'm off for a few days.

nsavoire:
> Thanks for the fix, I'll handover the integration to @nsavoire as I'm off for a few days.

I cherry-picked the commit into https://github.com/DataDog/ddprof/pull/534

xroche:
Already cherry-picked

PR #537, opened 2026-04-03
fix: strip embedded lib before hashing to prevent install-time mismatch

### LINKED ISSUE #538: Installed libdd_profiling-embedded.so hash mismatch after cmake --install --strip
The SHA-256 hash baked into the loader at build time doesn't match the installed `libdd_profiling-embedded.so` when the release tarball is produced with `cmake --install --strip`.

The install-time strip removes sections (`.symtab`, `.strtab`) that the link-time `--strip-debug` leaves intact, producing a different binary than what was hashed. The loader rejects the installed copy and falls back to `/tmp` extraction, which loses file capabilities (`CAP_PERFMON`) on the ddprof binary. This is what 8c26b1cf was supposed to prevent.

Production symptom:
```
ddprof: hash mismatch for installed libdd_profiling-embedded.so (got 1e3c9ca388e60d38..., expected af5015f127e9683d...); falling back to embedded library.
```

Fix in #537.

xroche:
## Summary

The SHA-256 hash baked into the loader at build time doesn't match the installed `libdd_profiling-embedded.so` when the release tarball is produced with `cmake --install --strip`. The install-time strip removes sections (`.symtab`, `.strtab`) that the link-time `--strip-debug` leaves intact, producing a different binary than what was hashed.

The loader rejects the installed copy and falls back to `/tmp` extraction, which loses file capabilities (`CAP_PERFMON`) on the ddprof binary. This is what 8c26b1cf was supposed to prevent.

Fixes #538

## How

- Add a `POST_BUILD` command that fully strips `libdd_profiling-embedded.so` before hashing and embedding. `cmake --install --strip` becomes a no-op.
- Add `verify_embedded_hash-ut.sh` (CTest) to catch mismatches during builds.
- Add `tools/verify_install_hashes.sh` for CI to run between install and tarball creation.
- Remove orphaned `test/version_mismatch-ut.sh` and `test/fake_profiling_version.c` (dead since ff0371f5).

Generated with [Claude Code](https://claude.ai/claude-code)


nsavoire:
Thanks for your feedback !
On our side we never `install/strip`, so we did not see the issue.
I guess the same issue impacts the ddprof binary.
Your change unconditionally strip all symbol information, which makes debugging / profiling a lot harder.
We prefer to keep symbol information for our releases (we already remove debug information).
I need some context to make a fix that can work for the different use cases:
- Do you really need to strip symbols (size gain is not big since debug information is already removed) ?
- Would a cmake option (something like `STRIP_ALL`  similar to the existing `STRIP_DEBUG_INFO`) work for you ?

xroche:
Closing — the issue is on our packaging side. We were running `cmake --install --strip` which re-strips the binary after the hash is computed. Dropping the `--strip` flag from our install step fixes it without any upstream changes. Thanks for the quick feedback.

PR #540, opened 2026-04-13
fix: improve `statsd` write error handling

KowalskiThomas:
# What does this PR do?

This makes an error handling code path non dead. Previously, we would check the `errno` before getting into the block, making the second check on `errno` useless/redundant/dead.

### DIFF
--- a/src/statsd.cc
+++ b/src/statsd.cc
@@ -137,12 +137,11 @@ DDRes statsd_send(int fd_sock, const char *key, const void *val, int type) {
   }
 
   // Nothing to do if the write fails
-  while (static_cast<ssize_t>(sz) != write(fd_sock, buf, sz) &&
-         errno == EINTR) {
+  while (static_cast<ssize_t>(sz) != write(fd_sock, buf, sz)) {
     // Don't consider this as fatal.
     if (errno == EWOULDBLOCK || errno == EAGAIN) {
       DDRES_RETURN_WARN_LOG(DD_WHAT_STATSD, "Write failed (sys buffer full)");
-    } else {
+    } else if (errno == EINTR) {
       DDRES_RETURN_WARN_LOG(DD_WHAT_STATSD, "Write failed");
     }
   }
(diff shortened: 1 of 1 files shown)


src/statsd.cc
```diff
@@ -137,12 +137,11 @@ DDRes statsd_send(int fd_sock, const char *key, const void *val, int type) {
   }
 
   // Nothing to do if the write fails
-  while (static_cast<ssize_t>(sz) != write(fd_sock, buf, sz) &&
```
nsavoire, line 140: Lots of suspicious things in this loop!
* it looks like it attempts to retry write in case of EINTR (interruption by a signal), but in fact it doesn't
* errno should only be checked when `write` returns -1 
* does not handle partial writes

r1viollet: We should check for partial write vs -1
- if partial write --> continue
- if error check what type of error we are getting
EINTR means we should continue
otherwise we should return a warning log and exit

KowalskiThomas: I'll take a look at all that!

r1viollet:
Thanks for looking into this, how did you find this issue ?

KowalskiThomas:
> how did you find this issue ?

Messing around with coding agents on plane wifi 😛

r1viollet:
> > how did you find this issue ?
> 
> Messing around with coding agents on plane wifi 😛

Perhaps we will want to generalize this messing around process ^^, we can talk about this.

src/statsd.cc
KowalskiThomas: Okay so I updated the change:
- If we do a partial write we continue until completion
- If we get `EINTR` we retry (I put a limit so that we don't accidentally end up infinite looping here -- I guess it shouldn't happen in real life but you never know)
- Any other error we early return after logging 

LMK if that makes sense!

nsavoire (dismissed):

KowalskiThomas:
@nsavoire re-requested review since fixing CI and pushing the branch dismissed it 😭

nsavoire (approved):
Thanks for the fixes !

PR #544, opened 2026-04-17
chore: add CLAUDE.md symlink and compile skill

r1viollet:
# What does this PR do?

- Symlink CLAUDE.md -> AGENTS.md so Claude Code auto-loads the agent guide (formatting rules, build workflow) in every session without duplicating content.
- Add a project-level `compile` skill that reuses a running Ubuntu 24 dev container when possible, otherwise starts one detached, then runs the DebTidy (clang + clang-tidy) build via `docker exec`.

# Motivation

Ensure you can compile from a claude agent.

# Additional Notes

NA

# How to test the change?

Start claude, do a /compile

### DIFF
--- a/.claude/skills/compile/SKILL.md
+++ b/.claude/skills/compile/SKILL.md
@@ -1,0 +1,127 @@
+---
+name: compile
+description: Build ddprof inside the Ubuntu 24 dev Docker container. Defaults to Debug (gcc, no clang-tidy) for speed. Use when the user asks to compile, build, or rebuild the project. Reuses a running container when possible, otherwise launches one in the background. Accepts an optional mode arg (Deb/Rel/DebTidy/San/TSan/Cov).
+---
+
+# Compile ddprof (Ubuntu 24 dev container)
+
+Builds **all targets** inside the Ubuntu 24 dev container. Default mode is
+**Debug** (gcc, no clang-tidy) — fastest iteration. Override via `args` when
+the user wants something else.
+
+## Args
+
+`args` is a single optional token: the build mode.
+
+| `args`     | Mode      | CC/CXX    | CMake helper      | Notes                                    |
+|------------|-----------|-----------|-------------------|------------------------------------------|
+| (empty)    | `Deb`     | gcc       | `DebCMake`        | Default. Fastest. No clang-tidy.         |
+| `rel`      | `Rel`     | gcc       | `RelCMake`        | Optimised + LTO. Slow link.              |
+| `debtidy`  | `DebTidy` | clang     | `DebTidyCMake`    | Enables clang-tidy. Slowest. Lint gate.  |
+| `san`      | `San`     | gcc       | `SanCMake`        | ASan + UBSan.                            |
+| `tsan`     | `TSan`    | gcc       | `TSanCMake`       | ThreadSanitizer.                         |
+| `cov`      | `Cov`     | gcc       | `CovCMake`        | Coverage instrumentation.                |
+
+Match `args` case-insensitively. If the user's natural-language request
+implies a mode (e.g. "do a release build", "run with sanitizers"), prefer
+that over the default. Only `DebTidy` requires `CC=clang CXX=clang++`; for
+the others leave `CC`/`CXX` unset so the container's gcc is used.
+
+## Step 1 — Find a running container
+
+```bash
+CID=$(docker ps \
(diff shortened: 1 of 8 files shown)


.claude/skills/compile/SKILL.md
```diff
@@ -0,0 +1,127 @@
...
+Example for the default (`Deb`):
+
+```bash
+docker exec "$CID" bash -lc '
+  set -euo pipefail
+  cd /app
+  source ./setup_env.sh
+  MkBuildDir Deb
+  DebCMake ../
+  make -j10
```
nsavoire: nit: I like to use `ninja` instead of `make`  (and it has the benefits of not having to specify a `-jXX` flag)

.claude/skills/compile/SKILL.md
```diff
@@ -0,0 +1,127 @@
...
+```
+
+Example for `DebTidy`:
+
+```bash
+docker exec "$CID" bash -lc '
+  set -euo pipefail
+  cd /app
+  export CC=clang CXX=clang++
```
nsavoire: nit: would be nice to have `DebTidyCMake` setting correctly `CC` and `CXX`

nsavoire (dismissed):

setup_env.sh
nsavoire: Ensure that `MkBuildDir DebTidy` uses a correct suffix:
```suggestion
  # $1: mode suffix (Rel, Debug, DebTidy, SanitizedDebug, ...)
  # DebTidy is clang-only by design — force the clang segment regardless of CC.
  local CC_SUFFIX=${DDPROF_EXTENSION_CC}
  if [[ "$1" == "DebTidy" ]]; then
    CC_SUFFIX=clang
  fi
  echo "_${CC_SUFFIX}_${DDPROF_EXTENSION_OS}_${1}"
```

setup_env.sh
```diff
@@ -71,7 +71,7 @@ CmakeWithOptions() {
   shift
   local VENDOR_EXTENSION=$(GetDirectoryExtention ${BUILD_TYPE})
   # shellcheck disable=SC2086
-  cmake_cmd="cmake ${COMMON_OPT} -DCMAKE_BUILD_TYPE=${BUILD_TYPE} -DVENDOR_EXTENSION=${VENDOR_EXTENSION} $@"
+  cmake_cmd="cmake -GNinja ${COMMON_OPT} -DCMAKE_BUILD_TYPE=${BUILD_TYPE} -DVENDOR_EXTENSION=${VENDOR_EXTENSION} $@"
```
nsavoire, line 81: Examples in `Agents.MD` need to be adjusted to use ninja instead of make

.claude/skills/compile/SKILL.md
```diff
@@ -0,0 +1,128 @@
...
+isn't blocked. **Do not pipe to `tee`** when backgrounding — `tee` swallows
+the failing exit code from `ninja` and the build looks like it succeeded. If
+you need the log on disk, redirect with `> /tmp/ddprof_compile.log 2>&1`
+instead, or set `set -o pipefail` and check `${PIPESTATUS[0]}` explicitly.
+
+After completion, check the exit code reported by the task notification.
+On failure, extract the first compiler/clang-tidy errors:
+
+```bash
+grep -E "error:" /tmp/ddprof_compile.log | head -20
```
nsavoire: This seems fragile and probably misses some errors

.claude/skills/compile/SKILL.md
```diff
@@ -0,0 +1,128 @@
...
+'
+```
+
+### Backgrounding & exit codes
+
+Long builds should run via `Bash` with `run_in_background: true` so the user
+isn't blocked. **Do not pipe to `tee`** when backgrounding — `tee` swallows
+the failing exit code from `ninja` and the build looks like it succeeded. If
+you need the log on disk, redirect with `> /tmp/ddprof_compile.log 2>&1`
+instead, or set `set -o pipefail` and check `${PIPESTATUS[0]}` explicitly.
```
nsavoire: This is confusing, it says don't pipe to tee, then says set `pipefail` AND check `${PIPESTATUS[0]}`.
With `pipefail` (already set in the heredoc one block above) `tee` is fine and `${PIPESTATUS[0]}` is redundant (it's an alternative to setting `pipefail`).

nsavoire (approved):

PR #546, opened 2026-04-24
fix: symbolizer cache never rotates on cache hits

r1viollet:
# What does this PR do?

Fixes a bug in the per-file symbolizer cache: reused entries were evicted after each cycle.

`remove_unvisited()` drops entries with `visited=false`, and `reset_unvisited_flag()` clears `visited` on every surviving entry at the end of each cycle. But `get_symbolizer()` only sets `visited=true` on the insert path — never on the cache-hit path. So an entry reused in the next cycle still ended that cycle with `visited=false` and was immediately evicted, throwing away the demangle / blazesym state every period.

One-line fix: set `visited=true` on cache hits too.

# Motivation

The cache was effectively never caching reused entries — the intended cross-cycle reuse of blazesym symbolizers and demangle maps was silently thrown away.

### DIFF
--- a/src/symbolizer.cc
+++ b/src/symbolizer.cc
@@ -42,6 +42,7 @@ void Symbolizer::reset_unvisited_flag() {
 Symbolizer::BlazeSymbolizerWrapper &
 Symbolizer::get_symbolizer(FileInfoId_t file_id, const std::string &elf_src) {
   if (auto it = _symbolizer_map.find(file_id); it != _symbolizer_map.end()) {
+    it->second.visited = true;
     return it->second;
   }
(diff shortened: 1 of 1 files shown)


r1viollet:
I noticed this while revisiting the interning 🤦

src/symbolizer.cc
nsavoire: nit: `BlazeSymbolizerWrapper` already initializes `visited{true}`
```suggestion
```

nsavoire (dismissed):

src/symbolizer.cc
r1viollet: True, thanks, I'll add a comment as it is slightly hidden

nsavoire (approved):

PR #547, opened 2026-04-29
fix: flip inverted overflow check

KowalskiThomas:
# What does this PR do?

This flips a comparison operator. Previously, we would early-return / fail if `raw_sz + raw_offset` was less than `sample->size_raw` and continue otherwise. In other words, if we were able to read past the bounds of `sample`, we would continue, and fail when within bounds. I think this should be the other way around.


r1viollet (approved):
LGTM
this would need to be tested with tracepoints, but this is not on you to fix the coverage there.

PR #548, opened 2026-05-05
Small fixes

nsavoire:


setup_env.sh
```diff
@@ -89,7 +96,14 @@ DebCMake() {
 
 DebTidyCMake() {
     local BUILD_TYPE=Debug
+    # VENDOR_EXTENSION uses BUILD_TYPE (Debug), so the DebTidy special-case in
+    # GetDirectoryExtention does not fire here — override locally so vendored
+    # deps land in _clang_*_Debug and are shared with CC=clang DebCMake builds
+    # rather than rebuilt for clang-tidy specifically.
     local DDPROF_EXTENSION_CC=clang
+    if [[ ( -n "${CC:-}" && "${CC%-*}" != "clang" ) || ( -n "${CXX:-}" && "${CXX%-*}" != "clang++" ) ]]; then
```
r1viollet, line 104: makes sense to fail early on this

.claude/skills/compile/SKILL.md
```diff
@@ -102,20 +100,34 @@ docker exec "$CID" bash -lc '
...
+instead in that case.)
+
+After completion, **check the exit code reported by the task notification
+first.** Only inspect the log if it's non-zero. Use the pattern below to find
+errors, and fall back to `tail` if nothing matches so a real failure can never look
+like success:
 
 ```bash
-grep -E "error:" /tmp/ddprof_compile.log | head -20
+errs=$(grep -nE "FAILED:|ninja: build stopped|error:|fatal error:|undefined reference|CMake Error|No space left|Killed" \
```
r1viollet, line 118: makes sense, I like the else branch

app/base-env/Dockerfile
```diff
@@ -31,11 +30,7 @@ FROM base-${UBUNTU_VERSION} AS base-gcc
 ENV CC=gcc-${GCC_VERSION}
 ENV CXX=g++-${GCC_VERSION}
 
-FROM base-${UBUNTU_VERSION} AS base-clang
```
r1viollet, line 34: So we no longer need this ? I thought that CI relies on this part. No breaking changes on CI ?

r1viollet: So this does impact CI, however we do not really use the default CC/CXX in the build. So the cleanup is trivial
```
--build-arg COMPILER=${COMPILER} 
```

nsavoire: Both gcc and clang are always installed in the image, `COMPILER` was only impacting the default one used, and in CI we always passed gcc as COMPILER and then controlled the compiler used with cmake arguments.

setup_env.sh
nsavoire: This does not fail, but warns that CC/CXX are ignored.

PR #550, opened 2026-05-26
Preserve live-allocation tracking across worker resets

r1viollet:
## What

Snapshot `LiveAllocation` to a parent-held memfd before a worker restart and
restore it in the newly-forked worker, so live-heap tracking survives the
periodic worker reset instead of starting from zero each time.

## Why

Workers are reset by forking a fresh process from the parent (every `worker_period` exports).
Everything in `DDProfWorkerContext` — including the heap-tracking aggregator
in `LiveAllocation` — is discarded. The target process keeps allocating
addresses the profiler no longer has stacks for, so live-heap is undercounted
until natural alloc/free traffic refills the map. The library has no way to
replay since it only tracks addresses, not stacks.

## How

- `main_loop` creates a memfd that the parent keeps open and every worker
  child inherits.
- Outgoing child captures a self-owned snapshot: `UnwindOutput` handles are
  resolved back to portable strings via libdatadog `Function2`/`Mapping2`
  read-back; all string_views into `Process`/base-frame caches are copied.
- The blob is written to the memfd just before the child exits (after the
  final synchronous export).
- Incoming child reads the blob in `worker_library_init` after the new
  `SymbolHdr` / `Symbolizer` / `ProfilesDictionary` are constructed but
  before the poll loop drains events. Mappings and functions are re-interned
  into the fresh dictionary; `LiveAllocation` owns a string deque backing
  the `string_view`s of restored `UnwindOutput`s.
- The memfd is truncated on read so a non-restarting exit doesn't leak stale
  state into the next worker.

## Budget enforcement (value-preserving)

- Default target 4 MB, hard ceiling 20 MB.
- Over budget: rank stacks by aggregate value, drop the lowest. Their
  addresses get remapped to a synthetic `[live-alloc cleared]` common frame
  so per-PID heap totals remain correct even when detail is shed.
- Still over: drop entire PIDs from lowest aggregate value upwards.
- New stats: `live_alloc.snapshot.bytes`, `live_alloc.snapshot.cleared_stacks`,
  `live_alloc.snapshot.dropped_pids`.

## Known gap

Events arriving between the old child's exit and the new child's first poll
are still lost. A library-side pause hook can be added separately.

## Tests

- New `live_allocation_snapshot-ut` (binary round-trip, bad-magic and
  truncation rejection, empty snapshot).
- Existing `live_allocation-ut` still green.

### DIFF
--- a/include/common_symbol_errors.hpp
+++ b/include/common_symbol_errors.hpp
@@ -14,10 +14,11 @@ using namespace std::string_view_literals;
 
 namespace ddprof {
 
-inline constexpr std::array<std::string_view, 6> k_common_frame_names = {
-    "[truncated]"sv,      "[unknown mapping]"sv,
-    "[unwind failure]"sv, "[incomplete]"sv,
-    "[lost]"sv,           "[maximum pids]"sv};
+inline constexpr std::array<std::string_view, 7> k_common_frame_names = {
+    "[truncated]"sv,           "[unknown mapping]"sv,
+    "[unwind failure]"sv,      "[incomplete]"sv,
+    "[lost]"sv,                "[maximum pids]"sv,
+    "[live-alloc cleared]"sv};
 
 enum SymbolErrors : std::uint8_t {
   truncated_stack = 0,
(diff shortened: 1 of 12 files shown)


r1viollet:
@codex review

r1viollet:
@codex review

r1viollet:
@codex review

PR #551, opened 2026-05-27
fix(jit): close JITDump startup race for Julia / .NET

r1viollet:
## Problem

JIT runtimes (Julia, .NET) publish a JITDump file via mmap that ddprof uses to resolve symbols in JIT'd anon-exec regions. At process startup, samples can land in JIT'd code before `pid_mapping._jitdump_addr` is set, leaving frames unsymbolized.

Two races contribute:

1. **Lost JITDump MMAP2 event.** `maybe_insert_erase_overlap` drops perf MMAP2 events whose timestamp predates the last backpopulate. If a backpopulate runs while /proc/maps hasn't yet caught up to the JITDump mmap, the subsequent perf event for the JITDump (with an older timestamp) is silently dropped.

2. **Sample before MMAP2 is consumed.** Even with events in order, the first samples can run `add_symbol` before the JITDump MMAP2 has been drained from the ring buffer. `_jitdump_addr == 0` sends us down the perf-map fallback (which Julia doesn't publish) and frames stay unsymbolized.

## Changes

* **`maybe_insert_erase_overlap`**: bypass the backpopulate-timestamp guard for `kJITDump` events. They are idempotent (`dso_find_adjust_same` dedupes) and the marker is required for JIT symbolization.

* **`DsoHdr::try_jitdump_discovery`**: new helper that scans `/proc/<pid>/maps` to discover a JITDump file. Gated to fire **at most once per PID lifetime**, only when no full /proc scan has ever run for this PID (`last_backpopulate_time == 0`). After any backpopulate, late JITDump events are already covered by the bypass above, so further rescans would be redundant.

* **`add_symbol` (unwind_dwfl.cc)**: when a sample lands in a runtime-symbols DSO with no known JITDump, call `try_jitdump_discovery` before falling back to perf-map. The rescan may mutate `pid_mapping._map` so the DSO reference is re-fetched afterwards; an `unknown_mapping` error frame is emitted if the mapping no longer qualifies.

## Cost

Negligible. The discovery path is bounded to one extra `/proc/<pid>/maps` parse per PID lifetime, and only when samples have arrived purely via perf events without ever triggering a backpopulate. Cleared naturally on exec/exit via `worker_pid_free`.

## Tests

* New `DSOTest.jitdump_bypasses_backpopulate_timestamp` covers the timestamp-guard bypass.
* Full `dso-ut` and `runtime_symbol_lookup-ut` suites pass.

## Follow-ups (not in this PR)

* End-to-end correctness scenario for the Julia race.

### DIFF
--- a/include/dso_hdr.hpp
+++ b/include/dso_hdr.hpp
@@ -99,6 +99,9 @@ public:
     BackpopulateState _backpopulate_state;
     // save the start addr of the jit dump info if available
     ProcessAddress_t _jitdump_addr = {};
+    // True if we already tried to discover a JITDump file for this pid
+    // during the current cycle (cleared by reset_backpopulate_state).
+    bool _jitdump_discovery_attempted = false;
   };
   using DsoPidMap = std::unordered_map<pid_t, PidMapping>;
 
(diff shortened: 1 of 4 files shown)


r1viollet:
@codex review

r1viollet:
I spent some time trying to repro and was not able to. I'll leave this PR as an attempt which might fix flaky prof-correctness runs. Alternative is to mark the run in prof-correctness as flaky.

nsavoire (approved):

PR #552, opened 2026-05-29
feat(pprof): emit per-sample process_language label

r1viollet:
## What

Emit a per-sample pprof label `process_language` carrying a heuristic native-language family for the profiled process' main executable (`go` / `rust` / `cpp`). When detection is inconclusive, no label is set and the existing `language="native"` tag continues to be the sole signal.

## Why

Today every native sample is reported as `language="native"` — Rust, C++, plain C and (mostly) Go are indistinguishable on the backend side. With this label, the backend can refine the per-process language for the dominant case where one language clearly drives a binary, without splitting profiles.

Mirrors the approach we will likely use in the OTel eBPF profiler / Full-Host Profiler (also no DWARF involvement, also per-sample label rather than a global tag), so the two producers will be consistent.

## How

* New `NativeLanguage` enum + cheap detection (`include/native_language.hpp` / `src/native_language.cc`):
  * **Go**   → `.go.buildinfo` / `.gopclntab` ELF section
  * **Rust** → `.note.rustc` section, or rustc-mangled symbols in `.symtab`/`.dynsym` (`_R…` v0, or legacy `…17h<16 hex>E` tail)
  * **Cpp**  → any `_Z…` Itanium-mangled symbol
  * else     → `kUnknown` (label omitted)
* Symbol-table scan bounded to 4096 entries.
* Detection runs **once per PID**, lazily on first sample, against the `Elf*` already cached by libdwfl (`dwfl_module_getelf`) — **no extra file open**.
* Result cached on `Process`, cleared via the existing `ProcessHdr::clear(pid)` on process exit.
* A `/proc/<pid>/exe` fallback path is kept for callers without an `Elf*` in hand (tests / early bring-up).
* Emitted as a pprof per-sample label `process_language` alongside `process_id`, `process_name`, etc. (`k_max_pprof_labels` raised 8 → 9).

## Verification

| Check | Result |
|---|---|
| `tools/style-check.sh` | clean |
| Full `ninja` (853 targets) | passes |
| `ctest -j4` | 52/55 pass; the 3 failures (`ddprof_stats-ut`, `ipc-ut`, `simple_malloc-with-event-reordering`) reproduce on `main` without this patch — pre-existing, unrelated |

Smoke-tested end-to-end with `ddprof -p <pid>`:
* Busy Go binary → `[NATIVE-LANG] -> go` (single detection log across all samples; libdwfl-cached Elf* reused)
* C++ binary → `cpp`
* `/bin/cat`, `sleep` (C) → `kUnknown` → no label, falls back to `native` tag

## Notes for review

* **Heuristic by design.** Mixed-language processes (e.g. C++ with linked Rust crates) will resolve to whichever signal hits first (Rust wins over C++). Acceptable for the dominant-language reporting goal.
* **No DWARF reads.** Symbol tables are scanned via libelf only; cost is one-shot per FileID and bounded.
* **No new dependency.** Reuses already-linked elfutils.
* **Per-sample label vs global tag.** Picked per-sample so multi-PID profiling stays correct; the global `language="native"` tag is unchanged.
* **Symbol-table availability.** For fully stripped binaries with no `.dynsym` C++ symbols, Cpp detection may miss — expected, label stays unset → `native`.

### DIFF
--- a/include/ddprof_process.hpp
+++ b/include/ddprof_process.hpp
@@ -10,6 +10,12 @@
 #include "ddres_def.hpp"
 #include "dwfl_wrapper.hpp"
 #include "logger.hpp"
+#include "native_language.hpp"
+
+// libelf forward declaration
+extern "C" {
+struct Elf;
+}
 
 #include <limits>
 #include <memory>
@@ -40,6 +46,19 @@ public:
 
   [[nodiscard]] std::string_view get_or_insert_thread_name(pid_t tid);
 
+  // Cached native language of the process' main executable.
+  // Returns kUnknown until populated by set_language().
+  NativeLanguage get_language() const { return _language; }
+
+  // Detect (only once) the language using an already-opened Elf* (typically
+  // libdwfl's main-module Elf*). No-op on subsequent calls.
+  // Returns true if detection was attempted on this call.
+  bool detect_language_once(::Elf *main_exe_elf);
+
+  // Fallback path: detect by opening /proc/<pid>/exe ourselves. Use only when
+  // no Elf* is available yet.
+  bool detect_language_once_from_proc(std::string_view path_to_proc);
+
   [[nodiscard]] DwflWrapper *get_or_insert_dwfl();
   [[nodiscard]] DwflWrapper *get_dwfl();
   [[nodiscard]] const DwflWrapper *get_dwfl() const;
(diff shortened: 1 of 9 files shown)


r1viollet:
@codex review

PR #553, opened 2026-06-01
fix(jit): retry jitdump read when file grows

r1viollet:
## Why

`RuntimeSymbolLookup::get_or_insert_jitdump` freezes a `(pid, path)` pair for the rest of the export cycle as soon as one sampled PC misses the symbol map after a successful jitdump read. The intent was to avoid reparsing the file for every subsequent miss in the same cycle.

The side effect: a single unlucky first sample — fired while the runtime is still flushing `jit-<pid>.dump` (Julia, .NET) — produces **zero symbolised JIT frames for the entire run** on scenarios that only export once. This matches the flaky Julia failure on prof-correctness, e.g. `ddprof_julia-20260601-074729`:

```
datadog.profiling.native.symbols.jit.reads: 1
datadog.profiling.native.symbols.jit.failed_lookups: 2177
datadog.profiling.native.symbols.jit.symbol_count: 0
```

## What

Add a soft-miss retry path. Every `k_jitdump_retry_check_every = 64` soft misses on a frozen jitdump, stat the file via the existing `get_file_inode` helper. If size or inode changed since the last read, unfreeze and reparse; otherwise keep the freeze cheaply. Hard read errors keep the existing behaviour (no churn on unreachable files such as `//anon`).

The new logic is in two small helpers — `check_jitdump_changed` and `read_jitdump` — so `get_or_insert_jitdump` remains a short orchestrator.

## Cost

At 99 Hz CPU sampling on up to ~20 cores, the worst case (every sample misses) is ~`99 * 20 / 64 ≈ 30` `stat()` calls per second profiler-wide. Reparses only happen when the file actually grew. In practice JIT map hits dominate and `stat` is invoked far less.

## Tests

Three new unit tests (`JitdumpRetry` fixture):
- `no_growth_does_not_reread` — burning the retry budget on an unchanged file does not trigger extra reads.
- `growth_triggers_reread_and_resolves_pc` — after `k_jitdump_retry_check_every` misses on a partial file, replacing it with a fuller one resolves the PC on the next lookup.
- `unreachable_path_does_not_churn` — `stat()` failing on `//anon` does not loop on reads.

## Notes / follow-ups

- `//anon` showing up as the jitdump path is a separate (pre-existing) issue: the kJITDump DSO gets shadowed by an anon mapping over the same range, likely during a backpopulate. Worth investigating, but orthogonal to this fix.
- Incremental parse (track `last_offset`, `pread` from there) is a natural follow-up if reparse cost ever becomes visible.

### DIFF
--- a/include/runtime_symbol_lookup.hpp
+++ b/include/runtime_symbol_lookup.hpp
@@ -58,9 +58,20 @@ public:
 private:
   using FailedCycle = HeterogeneousLookupStringMap<uint32_t>;
 
+  // Per-(pid, jitdump path) state used to decide when to retry a jitdump
+  // read after a soft miss (file readable, but the sampled PC was not in
+  // the records we already ingested). See get_or_insert_jitdump.
+  struct JitdumpRetryState {
+    int64_t size_bytes = 0;
+    inode_t inode = 0;
+    uint32_t misses_since_check = 0;
+  };
+  using JitdumpRetryMap = HeterogeneousLookupStringMap<JitdumpRetryState>;
+
   struct SymbolInfo {
     SymbolMap _map;
     FailedCycle _failed_cycle;
+    JitdumpRetryMap _jitdump_retry;
   };
   using PidUnorderedMap = std::unordered_map<pid_t, SymbolInfo>;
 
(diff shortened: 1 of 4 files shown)


nsavoire (approved):

KowalskiThomas (commented):
I have 0 knowledge of all this but removing the JIT dump part and just looking at the logic, it makes sense me / don't see anything wrong!

PR #555, opened 2026-06-23
Add mremap and shmat/shmdt instrumentation

r1viollet:
## What

Adds allocation tracking hooks for `mremap()` and System V shared memory (`shmat()`/`shmdt()`).

## Why

These APIs can allocate or deallocate memory without going through malloc/mmap, leading to under-reporting in allocation profiles. Notable examples:
- `mremap()` is used to grow large allocations (e.g., df-executor mmaps)
- `shmat()`/`shmdt()` are used by databases (PostgreSQL, Oracle) for System V shared memory

## Changes

**Instrumentation added** (`src/lib/symbol_overrides.cc`):
- `mremap()` - tracks old region as dealloc + new region as alloc
- `shmat()` - queries segment size via `shmctl(IPC_STAT)` and tracks as alloc
- `shmdt()` - tracks as dealloc

**Not instrumented** (with comment explaining why):
- `brk()`/`sbrk()` - manipulate the program break, not individual allocations. Would double-count if malloc uses them internally, and have wrong semantics (heap boundary vs allocated objects).

**Tests added** (`test/allocation_tracker-ut.cc`):
- `mremap` test via `test_realloc` pattern
- `shmat`/`shmdt` test with IPC_PRIVATE segment

All tests gated by weak symbol checks.

## Verification

Unit tests pass for the new hooks. Full CI validation pending.

### DIFF
--- a/include/common_symbol_errors.hpp
+++ b/include/common_symbol_errors.hpp
@@ -14,10 +14,11 @@ using namespace std::string_view_literals;
 
 namespace ddprof {
 
-inline constexpr std::array<std::string_view, 6> k_common_frame_names = {
-    "[truncated]"sv,      "[unknown mapping]"sv,
-    "[unwind failure]"sv, "[incomplete]"sv,
-    "[lost]"sv,           "[maximum pids]"sv};
+inline constexpr std::array<std::string_view, 7> k_common_frame_names = {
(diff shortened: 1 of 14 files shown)


r1viollet:
Replaced by #556 (clean single commit off main)

